PR #20, opened 2026-01-14
feat: allow selective module hooking

### LINKED ISSUE #19: Allow hooking only selected modules to reduce overhead
Currently, TraceML attaches hooks broadly to collect memory and timing information. While this is useful for debugging, it can add overhead when many leaf modules are hooked.

As an improvement, it may be useful to let users specify which modules or blocks to hook, instead of always descending to child layers.

Examples:
1. hook only modules whose names match "attention" or "transformer"
2.hook block-level modules and skip leaf layers
3. keep current behavior as a debug / detailed mode

This would allow:
1. lower overhead on large or fast GPUs (overhead is very important)
2. simpler, higher-level attribution (block instead of individual Linear layers)

This is not meant to replace detailed (leaf-level) tracing, only to provide a lighter alternative when users care about coarse attribution and lower overhead.

ppraneth:
**Hi @abhinavsriva and TraceML team I am new to the project and this is my first PR.**

This pull request addresses #19 by introducing flexible module filtering for instrumentation. Currently, TraceML attaches hooks broadly across all leaf modules, which can introduce significant overhead for large models or on memory-constrained hardware.

#### **What I have done:**

1. **Centralized Filtering**: Added a `get_hookable_modules` utility in `shared_utils.py` that allows filtering modules by name (`include_names`, `exclude_names`) and toggle between "leaf-only" and "coarse" (block-level) attribution.
2. **API Enhancements**: Updated the `trace_model_instance` decorator to accept these new filtering parameters.
3. **Refactored Hooks**: Updated all memory and timing hook attachment logic to utilize the new filtering system, ensuring consistency across forward/backward passes and execution tracking.
4. **Sampler Consistency**: Updated the `LayerMemorySampler` to respect user-defined filters so that the dashboard reports only the modules the user intended to trace.

#### **Issue Addressed:**

* **Closes #19**: This change allows users to significantly reduce tracing overhead by selecting specific high-level blocks (e.g., "attention") instead of hooking every individual layer.

#### **Verification:**

I successfully verified these changes using a BERT training script (`tracing_on_bert.py`). By targeting only "embeddings" and "attention" blocks with `leaf_only=False`, the tracker provided attribution for those blocks with a much lighter hook footprint.
Screenshot 2026-01-14 135359

Looking forward to your feedback!

### DIFF
--- a/src/traceml/utils/entry_hook.py
+++ b/src/traceml/utils/entry_hook.py
@@ -2,6 +2,7 @@ import torch.nn as nn
 import torch
 from traceml.utils.shared_utils import EXECUTION_LAYER
 
+
 _execution_entry_hook_registry = {}
 
 


ppraneth:
@abhinavsriva I have also added a `.gitignore` file

src/traceml/utils/entry_hook.py
```diff
@@ -33,14 +33,17 @@ def attach_backward_entry_hook(output, layer_name: str):
...
+    leaf_only=True
+):
     model_id = id(model)
     if _execution_entry_hook_registry.get(model_id):
         return
 
-    for name, module in model.named_modules():
-        if any(module.children()):
-            continue
+    for name, module in get_hookable_modules(model, include_names, exclude_names, leaf_only):
```
abhinavsriva: In entry hook this should be done. We want to attach hooks on all layer and not only at certain layers. The role is to find approximate layer where OOM came (not exact but closest). 

Once reverted, the PR is ready to merge.

abhinavsriva (changes requested):

abhinavsriva:
Hey @ppraneth  quick heads-up: I am relicensing TraceML to Apache-2.0 to make adoption easier.
Your contribution in this PR is currently in main. Can you confirm you are OK with relicensing your contribution under Apache-2.0 as well?
A simple “Yes, I agree” reply here is enough. Thanks again!

PR #31, opened 2026-02-11
Hf support

### LINKED ISSUE #30: Hugging Face integration: add TraceMLTrainer (subclass of transformers.Trainer) that wraps training_step with trace_step
Add an official Hugging Face integration by introducing a `TraceMLTrainer` that subclasses `transformers.Trainer` and automatically wraps each train step with context manager `trace_step(...)`.

Goal: **minimal/no code changes** for HF users to get TraceML’s always-on step timing + rank-aware aggregation (DDP single-node to start).

A huge fraction of PyTorch training happens via `transformers.Trainer`. Today, users need to edit the training loop to add traceML. 

That breaks the “one-line instrumentation” story and makes adoption harder.

Proposed API

```
from traceml.hf_decorators import TraceMLTrainer

trainer = TraceMLTrainer(
    model=model,
    args=training_args,
    train_dataset=train_ds,
    eval_dataset=eval_ds,
    traceml_enabled=True,          # optional (default True)
    traceml_kwargs=dict(...),      # optional
)

trainer.train()

```

Acceptance criteria

- Minimal example in docs/README: “Replace Trainer with TraceMLTrainer”
- A small integration test (CPU ok) that runs a few steps and asserts: training completes
- TraceML step counter increments (or events are emitted)
-  No additional dependencies introduced (HF remains optional)

Out of scope for v1 (can be follow-ups)
- Seq2SeqTrainer / TRL SFTTrainer support
- multi-node specific wiring
- wrapping eval loop

ppraneth:
image

Add support for HF
Closes #30

### DIFF
--- a/src/traceml/hf_decorators.py
+++ b/src/traceml/hf_decorators.py
@@ -50,20 +50,13 @@ class TraceMLTrainer(Trainer if HAS_TRANSFORMERS else object):
         if self.traceml_enabled:
             # Lazily attach hooks on the first step to ensure we catch the
             # final wrapped/moved model (e.g. DDP, Accelerator)
-            if (
+            if self.traceml_kwargs is not None and (
                 not self._traceml_hooks_attached
-                and self.traceml_kwargs is not None
+                or id(model) != getattr(self, "_attached_model_id", None)
             ):
                 try:
-                    # Unwrap model to ensure clean layer names and avoid DDP/FSDP wrapping issues
-                    if HAS_TRANSFORMERS:
-                        from transformers.modeling_utils import unwrap_model
-
-                        _model = unwrap_model(model)
-                    else:
-                        _model = model
-
-                    trace_model_instance(_model, **self.traceml_kwargs)
+                    trace_model_instance(model, **self.traceml_kwargs)
+                    self._attached_model_id = id(model)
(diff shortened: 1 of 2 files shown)


src/traceml/hf_decorators.py
```diff
@@ -0,0 +1,79 @@
...
+            ):
+                try:
+                    # Unwrap model to ensure clean layer names and avoid DDP/FSDP wrapping issues
+                    if HAS_TRANSFORMERS:
+                        from transformers.modeling_utils import unwrap_model
+
+                        _model = unwrap_model(model)
+                    else:
+                        _model = model
+
```
abhinavsriva: @ppraneth we don't need to worry about DDP/accelarate wrapping from hugging face. We should just use model instead and check if it wrapped or not. 

On GPU it was working as you showed in the photo. However on only CPU it is not working. I guess, internally accelerate/ddp wrapping get activated when it see GPU/CUDA and doesn't work with CPU only. 

I created a branch with the solution: https://github.com/traceopt-ai/traceml/tree/pr-hf

Can you please change the code accordingly so I can merge this PR.

ppraneth: @abhinavsriva pushed the changes

PR #75, opened 2026-04-14
perf: event-driven aggregator loop + inlined drain hot path

ppraneth:
Two performance fixes in the aggregator's message processing pipeline.

### 1. Event-driven drain (tcp_transport.py + trace_aggregator.py)

Previously the aggregator loop called `stop_event.wait(render_interval_sec)` on every iteration, meaning it would sleep for the full interval (default 1 s) even if messages were already sitting in the TCP queue.

Added a `threading.Event` (`_data_ready`) to `TCPServer` that is set by the client handler immediately after each decoded message is placed in the queue. A new `wait_for_data(timeout)` method wraps the event wait and auto-clears after returning.

The aggregator `_loop` now calls `wait_for_data(timeout=interval_sec)` instead of sleeping blindly. It wakes up as soon as data arrives and drains immediately.

The UI tick is still rate-limited via a `time.monotonic()` guard so the display refresh cadence is unchanged.

**Result:** end-to-end ingestion latency drops from up to `render_interval_sec` (1 s by default) down to near-zero.

### 2. Inlined drain sinks (trace_aggregator.py)

`_drain_tcp` previously called `_safe()` twice per message, each time constructing a `lambda m=msg: ...` closure. At 1000 messages/s that is 2000 closure allocations per second in the hot path.

Replaced both calls with direct `try/except` blocks. Behavior is identical: errors in one sink are still isolated from the other.

## Testing

Wrote targeted tests covering:
- `TCPServer._data_ready` exists and is a `threading.Event`
- `wait_for_data()` is callable and returns `False` on timeout, `True` on signal
- The event auto-clears so repeated waits behave correctly
- A burst of messages produces the correct single wakeup
- Store errors do not prevent the writer from receiving the same message
- Writer errors do not propagate out of `_drain_tcp`
- The fixed drain contains no `MAKE_FUNCTION` bytecode (no closures)
- A mock loop driven by the same event pattern drains within 200 ms of message arrival, not after the full interval
- The UI tick is not triggered faster than `interval_sec` under continuous data

All 13 tests passed.

## Files changed

- `src/traceml/transport/tcp_transport.py`
- `src/traceml/aggregator/trace_aggregator.py`


abhinavsriva:
@ppraneth Would it be possible to resolve this conflict? I introduced filtering so in remote db so we save only vertain samplers into remote db inmemory cache and mostly goes into sqlite db.

ppraneth:
> @ppraneth Would it be possible to resolve this conflict? I introduced filtering so in remote db so we save only vertain samplers into remote db inmemory cache and mostly goes into sqlite db.

Done

abhinavsriva (approved):

PR #87, opened 2026-04-24
feat: add step-scoped H2D .to() timing 

### LINKED ISSUE #82: Add step-scoped H2D .to() timing
### Description:

Add instrumentation for host-to-device transfers so it can measure tensor.to(cuda_device, non_blocking=True) during training. The automatic path should patch .to() globally, but only record timing while inside trace_step(...) so model initialization and other setup transfers are not included.

For manual, selective, and custom modes, expose an explicit wrapper such as traceml.wrap_h2d(...) for the object whose transfer we want to measure. The wrapper can return a proxy around the tensor or batch object so calls like x.to(...) are timed without requiring a broader API change. Keep this simple for now and avoid a context-manager design in the first version.

```
x = traceml.wrap_h2d(x)
x.to_device(..) 
```

### Acceptance Criteria:

.to() timing is recorded only during an active TraceML step.
Automatic mode patches .to() safely and does not measure model-moving/setup transfers.
Manual/selective/custom modes provide an explicit traceml.wrap_h2d(...) path.
The metric is captured as a TraceML sample and stored in the DB like other training signals.
No rendering/UI work is needed in this issue.

ppraneth:
## What this PR does

Closes #82.

Adds instrumentation for host-to-device transfers so TraceML can measure how long `tensor.to(cuda_device)` takes during training steps. This is a meaningful signal in data-heavy pipelines where the GPU can sit idle waiting for DMA to finish.

## How it works

**Automatic mode** patches `torch.Tensor.to()` once at `traceml.init()` time. The patch is gated by a thread-local flag that is only raised inside `trace_step()`, so model initialization, checkpoint loading, and any other setup transfers are completely ignored. Only CUDA-targeted calls are timed -- dtype-only casts and CPU-to-CPU copies pass through with zero overhead.

**Manual and selective modes** get an explicit `traceml.wrap_h2d(x)` wrapper. It returns a thin proxy around the tensor or batch object so the next `.to(device)` call is timed. No context manager, no broader API change -- just wrap and go:

```python
x = traceml.wrap_h2d(x)
x = x.to(device)
```

Both paths emit CUDA events (same approach as forward/backward timing) and feed into the existing `StepTimeSampler` pipeline, so the measurement lands in the `step_time_samples` SQLite table automatically alongside all other training signals.

## Files changed

- `src/traceml/utils/patches/h2d_auto_timer_patch.py` -- new patch module with TLS gating, CUDA target detection, and the `h2d_auto_timer` context manager
- `src/traceml/instrumentation.py` -- `h2d_auto_timer` nested inside `trace_step` alongside `forward_auto_timer` and `backward_auto_timer`
- `src/traceml/initialization.py` -- `patch_h2d` field added to `TraceMLInitConfig`, wired into `init()` and `start()`
- `src/traceml/wrappers.py` -- `_WrappedH2D` proxy class and `wrap_h2d()` function with duplicate instrumentation guard
- `src/traceml/api.py` -- `wrap_h2d` exposed at the public API layer; `init()` and `start()` accept `patch_h2d`
- `src/traceml/__init__.py` -- `wrap_h2d` added to top-level exports
- `tests/test_h2d_timing.py` -- 30 tests covering TLS gating, CUDA target detection, step scoping, manual wrapper, duplicate guards, and init config

## What is not in this PR

No rendering or UI changes. The `_event_bucket` summary function does not yet map `h2d_time` to a display bucket -- that is a follow-up consistent with what the issue specifies.

## Testing

All 30 tests pass without a GPU. Timing falls back to CPU wall-clock when CUDA is unavailable, so the full test suite runs in any environment.

```
python -m pytest tests/test_h2d_timing.py -v
# 30 passed in 1.81s
```


Pendu (commented):
test

Pendu (changes requested):
PR #87 addresses issue #82 with a competent implementation that matches the existing patch family (TLS gate, idempotent install, `timed_region` body, sentinel). The wrapper API and the `examples/h2d_timing_demo.py` example are good and should be kept.

Review converged on **4 must-fix items** before merge — all metric-correctness or test-isolation concerns:

- **C1 + C2** — `_traceml_tensor_to` filter only checks destination. Add source-device check (`if self.is_cuda`) and Parameter-receiver filter (`if isinstance(self, torch.nn.Parameter)`) to prevent D2D from being timed as h2d_time and `model.to(device)` from inflating `n_calls` by N parameters.
- **C3** — `wrap_h2d()` race window. Move sentinel check from constructor into `_WrappedH2D.to()`; defer gracefully when auto patch is active. Prevents wrap-then-init double-counting.
- **C4** — Docstring inaccuracy. `non_blocking=True` timing is *not* a "lower bound" — `try_resolve()` reads elapsed_time after the GPU completes, so the metric is accurate for both blocking and async transfers.

Plus **3 should-fix items**:
- **I1** — `_WrappedH2D` missing dunder forwarding (`__len__`, `__iter__`, `__getitem__`, `__contains__`); breaks the advertised dict-batch support.
- **I2** — `_is_cuda_target` substring match falsely matches `"cudacheckpoint"`-style strings; use `torch.device()` parsing instead.
- **I3** — `TraceMLInitConfig.patch_h2d: bool = False` default is asymmetric with the other three patch fields (which are required).

All must-fix items are localised (≤30 LOC each), each has a concrete verification recipe in the inline comments. C1 / C2 / C3 were verified end-to-end on a Tesla T4 — bugs reproduce on the current PR head, fixes produce the expected counts (C1: 1→0 events, C2: 20→0 events, C3: 2→1 events).

For reference, all proposed fixes are applied as atomic commits at https://github.com/Pendu/traceml/tree/pr-87-review-suggestions — feel free to cherry-pick or replicate independently. Each commit message is C1/C2/C3/C4/I1/I2/I3/M1-prefixed for clarity.

**Recommend: request changes** on the 4 must-fix items; approve once those are addressed.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,151 @@
...
+   timing is active.  The flag is raised only while inside ``trace_step()``.
+3. The ``h2d_auto_timer`` context manager toggles the flag.  It is entered by
+   ``trace_step`` the same way ``forward_auto_timer`` and
+   ``backward_auto_timer`` are used.
+
+GPU timing
+----------
+CUDA events are recorded around the ``.to()`` call via ``timed_region``.
+``try_resolve()`` calls ``elapsed_time()`` which blocks until both CUDA events
+have completed, so the measurement reflects actual GPU DMA time for both
```
Pendu: **C4 (must-fix): Docstring claims `non_blocking=True` is a "lower bound" — actually accurate**

Lines 23-25 currently say:

> "For `non_blocking=True` the events bound the time until the DMA is queued on the current stream, which is a useful **lower bound**."

This is inaccurate. `try_resolve()` in [`utils/timing.py:65-90`](https://github.com/traceopt-ai/traceml/blob/main/src/traceml/utils/timing.py#L65-L90) reads `start.elapsed_time(end)` *after* `end_evt.query()` returns True — meaning the elapsed value spans both record markers on the stream **including the asynchronous DMA between them**. CUDA events on the same stream are FIFO with the queue, so the timing reflects full GPU-side wall-clock, not just the enqueue cost.

A user reading the current docstring would expect the metric to undercount async transfers and add fudge factors. The metric is actually accurate.

Suggested rewrite for the "GPU timing" section (lines 20-25):

```rst
GPU timing
----------
CUDA events are recorded on the current stream around the ``.to()`` call.
``start.record()`` enqueues a timestamp marker before the DMA op;
``end.record()`` enqueues one after.  Once ``end`` fires (resolved later via
``event.query()`` in the sampler — see ``utils/timing.py::TimeEvent.try_resolve``),
``start.elapsed_time(end)`` returns the GPU-side wall-clock duration between
the two markers — including the asynchronous DMA itself.  Accuracy is the
same for ``non_blocking=True`` and ``non_blocking=False``.
```

No code changes, no test changes — verify by reading the new text against `try_resolve()` semantics.

Reference: commit [`72b43c9`](https://github.com/Pendu/traceml/commit/72b43c9) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Handles the common calling conventions:
+      tensor.to("cuda:0")
+      tensor.to(torch.device("cuda"))
+      tensor.to(device="cuda:0")
+      tensor.to(other_cuda_tensor)   # copies device from other_tensor
+    """
+    first = args[0] if args else None
+
+    if isinstance(first, str):
```
Pendu: **I2 (should-fix): `_is_cuda_target` uses fragile substring match**

`if isinstance(first, str) and "cuda" in first:` falsely matches strings containing "cuda" anywhere:

```python
"cuda" in "cudacheckpoint"     # True
"cuda" in "/path/to/cuda_dir"  # True
"cuda" in "acuda"              # True
```

A user passing `'/path/to/cudacheckpoint'` (or any directory name containing "cuda") to `.to(...)` would be classified as a CUDA target. Also case-sensitive (`"CUDA:0"` returns False — minor, since `torch.device("CUDA:0")` raises anyway).

Suggested fix — use `torch.device(...)` parsing in `try/except`:

```python
from typing import Optional

def _device_type(value: Any) -> Optional[str]:
    if isinstance(value, torch.device):
        return value.type
    if isinstance(value, torch.Tensor):
        return value.device.type
    if isinstance(value, str):
        try:
            return torch.device(value).type
        except (RuntimeError, TypeError):
            return None
    return None


def _is_cuda_target(args: tuple, kwargs: dict) -> bool:
    first = args[0] if args else None
    if _device_type(first) == "cuda":
        return True
    if _device_type(kwargs.get("device")) == "cuda":
        return True
    return False
```

All 9 existing `TestIsCudaTarget` assertions continue to pass under this version (verified by tracing each input). Add 2 new tests for the substring-bug case (`"/path/to/cudacheckpoint"`) and invalid-device strings (`"not_a_device"`).

Reference: commit [`0710c4b`](https://github.com/Pendu/traceml/commit/0710c4b) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,151 @@
...
+    device = kwargs.get("device")
+    if isinstance(device, str):
+        try:
+            if torch.device(device).type == "cuda":
+                return True
+        except RuntimeError:
+            pass
+    if isinstance(device, torch.device) and device.type == "cuda":
+        return True
+
```
Pendu: **C1 + C2 (must-fix): Filter only checks destination, not source or receiver type**

Two related metric-correctness bugs in the same function:

**C1 — D2D timed as h2d_time.** `_is_cuda_target()` checks destination only, so `cuda_tensor.to('cuda:1')` (a D2D peer copy via `cudaMemcpyPeer` over NVLink/PCIe) is recorded as h2d_time. This conflates two physically distinct DMA paths with different mitigations — a user with high `h2d_time` on a multi-GPU job might apply `pin_memory=True` (an H2D fix) when the real bottleneck is sharding overhead.

**C2 — `model.to(device)` overcounts.** `nn.Module.to(device)` calls `self._apply(fn)`, which calls `tensor.to(device)` once per parameter. With the auto patch, a 70B-param model fires ~12,000 h2d events per step. The aggregator collapses to one row but `n_calls` and `sum_ms` are inflated 100×–10000×. Depth tracking won't fix this — the calls are sequential under `_apply`, not nested.

Suggested fix — three short-circuits added in fastest-first order:

```python
def _traceml_tensor_to(self: torch.Tensor, *args: Any, **kwargs: Any) -> Any:
    if not _enabled():
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if self.is_cuda:                                # C1 — D2D / no-op
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if isinstance(self, torch.nn.Parameter):        # C2 — _apply traversal
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if not _is_cuda_target(args, kwargs):
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    with timed_region("_traceml_internal:h2d_time", scope="step", use_gpu=True):
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
```

Trade-offs: (a) Buffers go through `_apply` as plain `Tensor` (not `Parameter`), so they remain timed — minor residual overcount, acceptable for v1. (b) Direct `param.to('cuda')` outside model setup also short-circuits — rare pattern.

Verification on Tesla T4:
- C1 gate (`cuda_tensor.to('cuda:0')` inside `trace_step`): pre-fix = 1 event, post-fix = 0 ✓
- C2 gate (`model.to('cuda')` for 20-param model): pre-fix = 20 events, post-fix = 0 ✓

Reference: commits [`b224dcf`](https://github.com/Pendu/traceml/commit/b224dcf) (C1) and [`5f3c311`](https://github.com/Pendu/traceml/commit/5f3c311) (C2) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/sdk/wrappers.py
```diff
@@ -286,9 +294,97 @@ def _wrapped_step(*args: Any, **kwargs: Any) -> Any:
...
+
+    Unknown attributes are forwarded to the wrapped object so code that reads
+    tensor attributes (shape, dtype, ...) before calling ``.to(...)`` continues
+    to work naturally.
+    """
+
+    def __init__(self, obj: Any) -> None:
+        self._obj = obj
+
+    def to(self, *args: Any, **kwargs: Any) -> Any:
```
Pendu, line 289: **C3 (must-fix): `wrap_h2d()` race window — wrap-then-init produces double-counted events**

`_ensure_h2d_wrapper_allowed()` is called inside `wrap_h2d()` at construction time only. This handles the *static* case (auto already active when wrap is called) but NOT the temporal-ordering race:

```python
wrapped = traceml.wrap_h2d(batch)   # sentinel False → succeeds
traceml.init(mode="auto")           # sentinel becomes True
y = wrapped.to('cuda')              # both proxy AND patched Tensor.to fire → 2 events
```

The proxy entered `timed_region` (sentinel was False at construction time), and the underlying `tensor.to(...)` is now patched and also fires. One logical call → two events.

Suggested fix — move the sentinel check from constructor into `_WrappedH2D.to()`, and degrade gracefully (don't raise; defer to the auto patch):

```python
class _WrappedH2D:
    def __init__(self, obj: Any) -> None:
        self._obj = obj

    def to(self, *args: Any, **kwargs: Any) -> Any:
        # Re-check at use time — if auto patch is now active, defer to it
        if getattr(torch.Tensor, "_traceml_h2d_patched", False):
            return self._obj.to(*args, **kwargs)
        with timed_region(
            name="_traceml_internal:h2d_time",
            scope=TimeScope.STEP,
            use_gpu=True,
        ):
            return self._obj.to(*args, **kwargs)
    # ... __getattr__, __repr__ unchanged ...
```

Then remove `_ensure_h2d_wrapper_allowed()` from `wrap_h2d()` (the function is no longer called from anywhere; safe to delete).

Replace `test_wrap_h2d_raises_when_auto_patch_active` with `test_wrap_h2d_defers_when_auto_patch_active` that asserts no extra event when sentinel is True.

Verification on Tesla T4: wrap-then-init-then-`.to('cuda')` → pre-fix = 2 events (double-count), post-fix = 1 event ✓.

Reference: commit [`f471776`](https://github.com/Pendu/traceml/commit/f471776) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/sdk/wrappers.py
```diff
@@ -286,9 +294,97 @@ def _wrapped_step(*args: Any, **kwargs: Any) -> Any:
...
+        # If the auto-patch was installed after this wrapper was created
+        # (wrap-then-init race), defer to avoid double-counting: the patch
+        # will time this call.
+        if getattr(torch.Tensor, "_traceml_h2d_patched", False):
+            return self._obj.to(*args, **kwargs)
+
+        with timed_region(
+            name="_traceml_internal:h2d_time",
```
Pendu, line 297: **I1 (should-fix): `_WrappedH2D` missing dunder forwarding for dict-like batches**

The class advertises support for "tensor or batch object" (line 299), but Python bypasses `__getattr__` for special methods on the class. So:

```python
batch = traceml.wrap_h2d({"x": tensor, "y": tensor})
batch["x"]   # → TypeError: '_WrappedH2D' object is not subscriptable
len(batch)   # → TypeError
"x" in batch # → TypeError
```

`_WrappedDataLoaderFetch` in this same file (line 143 — `def __len__(self) -> int: return len(self._loader)`) defines `__len__` for exactly this reason; the precedent exists in the family.

Suggested fix — add four special-method forwards alongside `__getattr__`:

```python
def __getattr__(self, name: str) -> Any:
    return getattr(self._obj, name)

# Special methods bypass __getattr__ on the class — forward explicitly.
def __len__(self) -> int:
    return len(self._obj)

def __iter__(self) -> Any:
    return iter(self._obj)

def __getitem__(self, key: Any) -> Any:
    return self._obj[key]

def __contains__(self, item: Any) -> bool:
    return item in self._obj
```

If `self._obj` doesn't support an operation, the underlying `TypeError` propagates cleanly with the right type name.

Test: a fake dict-like batch with `.to()` and `__getitem__/__len__/__iter__/__contains__`, wrapped, then assert `wrapped["x"]`, `len(wrapped)`, `"x" in wrapped`, `set(wrapped)` all work.

Reference: commit [`a3272bc`](https://github.com/Pendu/traceml/commit/a3272bc) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

ppraneth:
Ran test on demo example:
```sql
root@5eaccde5a19d:/workspace/traceml/examples#  sqlite3 /workspace/traceml/examples/./logs/session_20260507_101854_2b8d32/aggregator/telemetry "SELECT step, events_json FROM step_time_samples ORDER BY step;"    
1|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":18.137311935424805,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":7.367610931396484,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":118.37523651123047,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.12815999682061374,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":42.59270477294922,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":179.65936660766602,"is_gpu":false,"n_calls":1}}}
2|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.4994879961013794,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":4.642724990844727,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.27427200973033905,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.11654399917460978,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.2598080039024353,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":1.438140869140625,"is_gpu":false,"n_calls":1}}}
3|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.26015999913215637,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.598285675048828,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15891199931502342,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.10048000304959714,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.18028800189495087,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8478164672851562,"is_gpu":false,"n_calls":1}}}
4|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.25017601251602173,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.1352767944335938,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15635200217366219,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.10048000235110521,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17849600315093994,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8370876312255859,"is_gpu":false,"n_calls":1}}}
5|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2553600072860718,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.124309539794922,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15689599886536598,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.09974399791099131,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.18512000143527985,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8335113525390625,"is_gpu":false,"n_calls":1}}}
6|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2641279995441437,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.1238327026367188,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15795200318098068,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.0999679989181459,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17814399302005768,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8370876312255859,"is_gpu":false,"n_calls":1}}}
7|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2544640004634857,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.1042823791503906,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.1547519974410534,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.09990399796515703,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17455999553203583,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8249282836914062,"is_gpu":false,"n_calls":1}}}
8|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.5720639824867249,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":67.69990921020508,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.24569600820541382,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.11942399758845568,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.2595199942588806,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":1.436471939086914,"is_gpu":false,"n_calls":1}}}
9|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.26342400908470154,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.182483673095703,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15856000036001205,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.09932800219394267,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.18300800025463104,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8547306060791016,"is_gpu":false,"n_calls":1}}}
10|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2863999903202057,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.100229263305664,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15731199830770493,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.09999999962747097,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17606399953365326,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8575916290283203,"is_gpu":false,"n_calls":1}}}
11|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2605440020561218,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.107858657836914,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.1576320007443428,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.10089599876664579,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17750400304794312,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8385181427001953,"is_gpu":false,"n_calls":1}}}
12|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2518720030784607,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.1152496337890625,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.16527999192476273,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.10067199706099927,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17776000499725342,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8368492126464844,"is_gpu":false,"n_calls":1}}}
13|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2507840096950531,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.2199153900146484,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15545599907636642,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.09990399843081832,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17449599504470825,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8161067962646484,"is_gpu":false,"n_calls":1}}}
14|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.6035839915275574,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":82.54051208496094,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.2785920053720474,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.12255999981425703,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.2832320034503937,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":1.5437602996826172,"is_gpu":false,"n_calls":1}}}
15|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.25836798548698425,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":1.8281936645507812,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15625599771738052,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.10076799825765193,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.18614399433135986,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8404254913330078,"is_gpu":false,"n_calls":1}}}
16|{"_traceml_internal:backward_time":{"cuda:0":{"duration_ms":0.2791360020637512,"is_gpu":true,"n_calls":1}},"_traceml_internal:dataloader_next":{"cpu":{"duration_ms":2.0956993103027344,"is_gpu":false,"n_calls":1}},"_traceml_internal:forward_time":{"cuda:0":{"duration_ms":0.15519999712705612,"is_gpu":true,"n_calls":2}},"_traceml_internal:h2d_time":{"cuda:0":{"duration_ms":0.1000320001039654,"is_gpu":true,"n_calls":2}},"_traceml_internal:optimizer_step":{"cuda:0":{"duration_ms":0.17350399494171143,"is_gpu":true,"n_calls":1}},"_traceml_internal:step_time":{"cpu":{"duration_ms":0.8404254913330078,"is_gpu":false,"n_calls":1}}}
```

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,150 @@
...
+   timing is active.  The flag is raised only while inside ``trace_step()``.
+3. The ``h2d_auto_timer`` context manager toggles the flag.  It is entered by
+   ``trace_step`` the same way ``forward_auto_timer`` and
+   ``backward_auto_timer`` are used.
+
+GPU timing
+----------
+CUDA events are recorded on the current stream around the ``.to()`` call.
+``start.record()`` enqueues a timestamp marker before the DMA op;
+``end.record()`` enqueues one after.  Once ``end`` fires (resolved later via
```
Pendu, line 24: **C4 (must-fix): Docstring claims `non_blocking=True` is a "lower bound" — actually accurate**

Lines 23-25 currently say:

> "For `non_blocking=True` the events bound the time until the DMA is queued on the current stream, which is a useful **lower bound**."

This is inaccurate. `try_resolve()` in [`utils/timing.py:65-90`](https://github.com/traceopt-ai/traceml/blob/main/src/traceml/utils/timing.py#L65-L90) reads `start.elapsed_time(end)` *after* `end_evt.query()` returns True — meaning the elapsed value spans both record markers on the stream **including the asynchronous DMA between them**. CUDA events on the same stream are FIFO with the queue, so the timing reflects full GPU-side wall-clock, not just the enqueue cost.

A user reading the current docstring would expect the metric to undercount async transfers and add fudge factors. The metric is actually accurate.

Suggested rewrite for the "GPU timing" section (lines 20-25):

```rst
GPU timing
----------
CUDA events are recorded on the current stream around the ``.to()`` call.
``start.record()`` enqueues a timestamp marker before the DMA op;
``end.record()`` enqueues one after.  Once ``end`` fires (resolved later via
``event.query()`` in the sampler — see ``utils/timing.py::TimeEvent.try_resolve``),
``start.elapsed_time(end)`` returns the GPU-side wall-clock duration between
the two markers — including the asynchronous DMA itself.  Accuracy is the
same for ``non_blocking=True`` and ``non_blocking=False``.
```

No code changes, no test changes — verify by reading the new text against `try_resolve()` semantics.

Reference: commit [`72b43c9`](https://github.com/Pendu/traceml/commit/72b43c9) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,150 @@
...
+def _device_type(value: Any) -> Optional[str]:
+    """Return the device type string for a value, or None if not recognisable."""
+    if isinstance(value, torch.device):
+        return value.type
+    if isinstance(value, torch.Tensor):
+        return value.device.type
+    if isinstance(value, str):
+        try:
+            return torch.device(value).type
+        except (RuntimeError, TypeError):
```
Pendu, line 73: **I2 (should-fix): `_is_cuda_target` uses fragile substring match**

`if isinstance(first, str) and "cuda" in first:` falsely matches strings containing "cuda" anywhere:

```python
"cuda" in "cudacheckpoint"     # True
"cuda" in "/path/to/cuda_dir"  # True
"cuda" in "acuda"              # True
```

A user passing `'/path/to/cudacheckpoint'` (or any directory name containing "cuda") to `.to(...)` would be classified as a CUDA target. Also case-sensitive (`"CUDA:0"` returns False — minor, since `torch.device("CUDA:0")` raises anyway).

Suggested fix — use `torch.device(...)` parsing in `try/except`:

```python
from typing import Optional

def _device_type(value: Any) -> Optional[str]:
    if isinstance(value, torch.device):
        return value.type
    if isinstance(value, torch.Tensor):
        return value.device.type
    if isinstance(value, str):
        try:
            return torch.device(value).type
        except (RuntimeError, TypeError):
            return None
    return None


def _is_cuda_target(args: tuple, kwargs: dict) -> bool:
    first = args[0] if args else None
    if _device_type(first) == "cuda":
        return True
    if _device_type(kwargs.get("device")) == "cuda":
        return True
    return False
```

All 9 existing `TestIsCudaTarget` assertions continue to pass under this version (verified by tracing each input). Add 2 new tests for the substring-bug case (`"/path/to/cudacheckpoint"`) and invalid-device strings (`"not_a_device"`).

Reference: commit [`0710c4b`](https://github.com/Pendu/traceml/commit/0710c4b) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
```diff
@@ -0,0 +1,150 @@
...
+      tensor.to(torch.device("cuda"))
+      tensor.to(device="cuda:0")
+      tensor.to(other_cuda_tensor)   # copies device from other_tensor
+    """
+    first = args[0] if args else None
+    if _device_type(first) == "cuda":
+        return True
+    if _device_type(kwargs.get("device")) == "cuda":
+        return True
+    return False
```
Pendu, line 93: **C1 + C2 (must-fix): Filter only checks destination, not source or receiver type**

Two related metric-correctness bugs in the same function:

**C1 — D2D timed as h2d_time.** `_is_cuda_target()` checks destination only, so `cuda_tensor.to('cuda:1')` (a D2D peer copy via `cudaMemcpyPeer` over NVLink/PCIe) is recorded as h2d_time. This conflates two physically distinct DMA paths with different mitigations — a user with high `h2d_time` on a multi-GPU job might apply `pin_memory=True` (an H2D fix) when the real bottleneck is sharding overhead.

**C2 — `model.to(device)` overcounts.** `nn.Module.to(device)` calls `self._apply(fn)`, which calls `tensor.to(device)` once per parameter. With the auto patch, a 70B-param model fires ~12,000 h2d events per step. The aggregator collapses to one row but `n_calls` and `sum_ms` are inflated 100×–10000×. Depth tracking won't fix this — the calls are sequential under `_apply`, not nested.

Suggested fix — three short-circuits added in fastest-first order:

```python
def _traceml_tensor_to(self: torch.Tensor, *args: Any, **kwargs: Any) -> Any:
    if not _enabled():
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if self.is_cuda:                                # C1 — D2D / no-op
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if isinstance(self, torch.nn.Parameter):        # C2 — _apply traversal
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    if not _is_cuda_target(args, kwargs):
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
    with timed_region("_traceml_internal:h2d_time", scope="step", use_gpu=True):
        return _ORIG_TENSOR_TO(self, *args, **kwargs)
```

Trade-offs: (a) Buffers go through `_apply` as plain `Tensor` (not `Parameter`), so they remain timed — minor residual overcount, acceptable for v1. (b) Direct `param.to('cuda')` outside model setup also short-circuits — rare pattern.

Verification on Tesla T4:
- C1 gate (`cuda_tensor.to('cuda:0')` inside `trace_step`): pre-fix = 1 event, post-fix = 0 ✓
- C2 gate (`model.to('cuda')` for 20-param model): pre-fix = 20 events, post-fix = 0 ✓

Reference: commits [`b224dcf`](https://github.com/Pendu/traceml/commit/b224dcf) (C1) and [`5f3c311`](https://github.com/Pendu/traceml/commit/5f3c311) (C2) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/sdk/initial.py
```diff
@@ -56,6 +58,7 @@ def _build_config(
     patch_dataloader: Optional[bool],
     patch_forward: Optional[bool],
     patch_backward: Optional[bool],
+    patch_h2d: Optional[bool],
     source: str,
```
Pendu, line 62: **I3 (should-fix): `patch_h2d` field default is asymmetric with the family**

```suggestion
    patch_h2d: bool
```

The other three patch fields (`patch_dataloader`, `patch_forward`, `patch_backward`) are required (no default). `_build_config` always passes `patch_h2d` explicitly in all three mode branches (auto/manual/selective), so the default is dead code in normal flow — and asymmetric with the family invariant.

No test changes needed; existing tests pass.

Reference: commit [`222a936`](https://github.com/Pendu/traceml/commit/222a936) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/sdk/wrappers.py
```diff
@@ -266,9 +266,96 @@ def _wrapped_step(*args: Any, **kwargs: Any) -> Any:
...
+
+    def __iter__(self) -> Any:
+        return iter(self._obj)
+
+    def __getitem__(self, key: Any) -> Any:
+        return self._obj[key]
+
+    def __contains__(self, item: Any) -> bool:
+        return item in self._obj
+
```
Pendu, line 317: **C3 (must-fix): `wrap_h2d()` race window — wrap-then-init produces double-counted events**

`_ensure_h2d_wrapper_allowed()` is called inside `wrap_h2d()` at construction time only. This handles the *static* case (auto already active when wrap is called) but NOT the temporal-ordering race:

```python
wrapped = traceml.wrap_h2d(batch)   # sentinel False → succeeds
traceml.init(mode="auto")           # sentinel becomes True
y = wrapped.to('cuda')              # both proxy AND patched Tensor.to fire → 2 events
```

The proxy entered `timed_region` (sentinel was False at construction time), and the underlying `tensor.to(...)` is now patched and also fires. One logical call → two events.

Suggested fix — move the sentinel check from constructor into `_WrappedH2D.to()`, and degrade gracefully (don't raise; defer to the auto patch):

```python
class _WrappedH2D:
    def __init__(self, obj: Any) -> None:
        self._obj = obj

    def to(self, *args: Any, **kwargs: Any) -> Any:
        # Re-check at use time — if auto patch is now active, defer to it
        if getattr(torch.Tensor, "_traceml_h2d_patched", False):
            return self._obj.to(*args, **kwargs)
        with timed_region(
            name="_traceml_internal:h2d_time",
            scope=TimeScope.STEP,
            use_gpu=True,
        ):
            return self._obj.to(*args, **kwargs)
    # ... __getattr__, __repr__ unchanged ...
```

Then remove `_ensure_h2d_wrapper_allowed()` from `wrap_h2d()` (the function is no longer called from anywhere; safe to delete).

Replace `test_wrap_h2d_raises_when_auto_patch_active` with `test_wrap_h2d_defers_when_auto_patch_active` that asserts no extra event when sentinel is True.

Verification on Tesla T4: wrap-then-init-then-`.to('cuda')` → pre-fix = 2 events (double-count), post-fix = 1 event ✓.

Reference: commit [`f471776`](https://github.com/Pendu/traceml/commit/f471776) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

src/traceml/sdk/wrappers.py
```diff
@@ -266,9 +266,96 @@ def _wrapped_step(*args: Any, **kwargs: Any) -> Any:
...
+    def __repr__(self) -> str:
+        return f"_WrappedH2D({self._obj!r})"
+
+
+def wrap_h2d(obj: Any) -> "_WrappedH2D":
+    """
+    Wrap a tensor or batch object so the next ``.to(...)`` call is timed.
+
```
Pendu, line 325: **I1 (should-fix): `_WrappedH2D` missing dunder forwarding for dict-like batches**

The class advertises support for "tensor or batch object" (line 299), but Python bypasses `__getattr__` for special methods on the class. So:

```python
batch = traceml.wrap_h2d({"x": tensor, "y": tensor})
batch["x"]   # → TypeError: '_WrappedH2D' object is not subscriptable
len(batch)   # → TypeError
"x" in batch # → TypeError
```

`_WrappedDataLoaderFetch` in this same file (line 143 — `def __len__(self) -> int: return len(self._loader)`) defines `__len__` for exactly this reason; the precedent exists in the family.

Suggested fix — add four special-method forwards alongside `__getattr__`:

```python
def __getattr__(self, name: str) -> Any:
    return getattr(self._obj, name)

# Special methods bypass __getattr__ on the class — forward explicitly.
def __len__(self) -> int:
    return len(self._obj)

def __iter__(self) -> Any:
    return iter(self._obj)

def __getitem__(self, key: Any) -> Any:
    return self._obj[key]

def __contains__(self, item: Any) -> bool:
    return item in self._obj
```

If `self._obj` doesn't support an operation, the underlying `TypeError` propagates cleanly with the right type name.

Test: a fake dict-like batch with `.to()` and `__getitem__/__len__/__iter__/__contains__`, wrapped, then assert `wrapped["x"]`, `len(wrapped)`, `"x" in wrapped`, `set(wrapped)` all work.

Reference: commit [`a3272bc`](https://github.com/Pendu/traceml/commit/a3272bc) on https://github.com/Pendu/traceml/tree/pr-87-review-suggestions.

Pendu:
Thanks for running the demo and posting the SQLite dump, @ppraneth — that's confirmation that the data plane works end-to-end. A few observations, then a note on what the review is actually asking for.

## What your run validates ✅

- `_traceml_internal:h2d_time` fires every step (`n_calls=2`, matching the two
  `batch_x.to()` / `batch_y.to()` calls in the demo).
- Event lands on `cuda:0` with `is_gpu: true`, ~0.1ms per step — sane on this
  hardware.
- TLS gate works: `model.to(device)` outside `trace_step` did not pollute
  the per-step counts.

I got the same shape on Tesla T4 with the loop extended to 2000 steps:
median 0.16ms, p95 0.25ms. So the happy-path patch behavior is solid.

## What this run doesn't cover

The h2d_timing_demo deliberately exercises only the textbook usage: CPU
source, CUDA destination, tensor batches (not dicts), no `wrap_h2d` mixed
with `mode="auto"`. The review findings target paths the demo never touches:

| Review item | Scenario that triggers it (not in the demo) |
|---|---|
| **C1** | `cuda_tensor.to('cuda:1')` — a D2D peer copy gets counted as h2d_time |
| **C2** | `model.to(device)` issued *inside* `trace_step` — inflates `n_calls` by parameter count |
| **C3** | `wrap_h2d(x)` *before* `traceml.init(mode="auto")` — the same transfer is timed twice |
| **C4** | docstring inaccuracy — `non_blocking=True` events are accurate, not a "lower bound" |
| **I1** | `wrap_h2d({"input": x, "labels": y})` — `batch["input"]` fails because `__getitem__` isn't forwarded |
| **I2** | destination string is `"cudacheckpoint"` or `/some/cuda_path/` — passes substring check, fails as a real device |

Each of those is wired up as a 5-15 line verification recipe inside the
inline review comments (look for the **Verification** section). Running
those recipes against the current PR head reproduces the bug; running them
against a fixed branch shows the expected behavior.

Concretely on a Tesla T4 against the current PR head:
- **C1** recipe yields 1 h2d_time event for a D2D copy (expected: 0)
- **C2** recipe yields 20 h2d_time events for a 20-param `model.to(device)` (expected: 0)
- **C3** recipe yields 2 events for a single wrapped transfer (expected: 1)

For reference, a branch with all 8 fixes applied as atomic commits is at
https://github.com/Pendu/traceml/tree/pr-87-review-suggestions — each
commit is prefixed `C1`, `C2`, `C3`, etc. to map back to the review.
Feel free to cherry-pick, replicate, or take a different approach.

Happy to walk through any specific finding if the verification recipe
isn't clear — feel free to push back if you think any of these aren't
worth fixing.

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
ppraneth: This is already fixed @Pendu

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
ppraneth: Did something similar

src/traceml/instrumentation/patches/h2d_auto_timer_patch.py
ppraneth: **C1 + C2  Filter only checks destination, not source or receiver type:**

> Already fixed. Both short-circuits are in `_traceml_tensor_to`, evaluated in fastest-first order before the destination check:
>
> - **C1 (`self.is_cuda` gate):** If the source tensor is already on CUDA the call is a D2D peer copy, not an H2D transfer — returned immediately without entering `timed_region`. Identical to the suggested fix.
> - **C2 (`isinstance(self, torch.nn.Parameter)` gate):** `nn.Module.to(device)` traverses all parameters via `_apply`; each fires `.to()` once. Skipping `Parameter` instances eliminates the per-parameter overcount. Trade-offs acknowledged: non-`Parameter` buffers under `_apply` are still timed (minor residual, acceptable for v1); direct `param.to('cuda')` outside model setup also short-circuits (rare pattern).
>
> Tests already in place:
> - `test_c1_d2d_skips_timing_when_source_is_cuda` — mocks `.is_cuda = True`, calls `.to("cuda:0")` inside a step, asserts zero `h2d_time` events recorded
> - `test_c2_parameter_skips_timing_to_prevent_model_to_device_overcount` — wraps an `nn.Parameter`, calls `.to("cuda:0")` inside a step, asserts zero events recorded

src/traceml/sdk/initial.py
ppraneth: **I3 `patch_h2d` field default asymmetric with the family:**

> Already fixed. `TraceMLInitConfig` now declares `patch_h2d: bool` with no default, symmetric with `patch_dataloader`, `patch_forward`, and `patch_backward`. All three branches in `_build_config` (auto / manual / selective) set it explicitly, so the field is always provided at construction time the previous default was effectively dead code.
>
> The `Optional[bool]` on the `_build_config` parameter signature is intentional and symmetric. All four patch parameters are typed as `Optional[bool]` there because `init()` passes `None` for user-omitted overrides.

src/traceml/sdk/wrappers.py
ppraneth: **C3 `wrap_h2d()` race window — wrap-then-init double-count:**

> Already fixed. The sentinel check was moved from `wrap_h2d()` construction time into `_WrappedH2D.to()`, so it is re-evaluated on every call. If `init(mode="auto")` runs after `wrap_h2d()`, the next `.to()` call sees `_traceml_h2d_patched = True` and delegates directly to `self._obj.to(...)` — the proxy becomes a pass-through and the auto-patch remains the sole timing source.
>
> `_ensure_h2d_wrapper_allowed()` was removed entirely since it is no longer referenced anywhere.
>
> `test_wrap_h2d_raises_when_auto_patch_active` was replaced with `test_wrap_h2d_defers_when_auto_patch_active`, which creates the wrapper first, sets `_traceml_h2d_patched = True` to simulate the race, then calls `.to("cuda:0")` and asserts `timed_region` is never entered (zero extra events recorded).

src/traceml/sdk/wrappers.py
ppraneth: **I1 — `_WrappedH2D` missing dunder forwarding for dict-like batches:**

> Already fixed. `_WrappedH2D` now explicitly defines `__len__`, `__iter__`, `__getitem__`, and `__contains__` alongside `__getattr__`, following the same pattern as `_WrappedDataLoaderFetch` in the same file. Python's special-method lookup bypasses `__getattr__` entirely for dunder calls, so these must be declared on the class.
>
> Test `test_wrap_h2d_dict_batch_dunder_forwarding` covers all four: a `DictBatch` stub with `.to()` is wrapped, then `wrapped["x"]`, `len(wrapped)`, `"x" in wrapped`, and `set(wrapped)` (iterator) are all asserted to work without `TypeError`.

ppraneth:
> Thanks for running the demo and posting the SQLite dump, @ppraneth — that's confirmation that the data plane works end-to-end. A few observations, then a note on what the review is actually asking for.
> 
> ## What your run validates ✅
> * `_traceml_internal:h2d_time` fires every step (`n_calls=2`, matching the two
>   `batch_x.to()` / `batch_y.to()` calls in the demo).
> * Event lands on `cuda:0` with `is_gpu: true`, ~0.1ms per step — sane on this
>   hardware.
> * TLS gate works: `model.to(device)` outside `trace_step` did not pollute
>   the per-step counts.
> 
> I got the same shape on Tesla T4 with the loop extended to 2000 steps: median 0.16ms, p95 0.25ms. So the happy-path patch behavior is solid.
> 
> ## What this run doesn't cover
> The h2d_timing_demo deliberately exercises only the textbook usage: CPU source, CUDA destination, tensor batches (not dicts), no `wrap_h2d` mixed with `mode="auto"`. The review findings target paths the demo never touches:
> 
> Review item	Scenario that triggers it (not in the demo)
> **C1**	`cuda_tensor.to('cuda:1')` — a D2D peer copy gets counted as h2d_time
> **C2**	`model.to(device)` issued _inside_ `trace_step` — inflates `n_calls` by parameter count
> **C3**	`wrap_h2d(x)` _before_ `traceml.init(mode="auto")` — the same transfer is timed twice
> **C4**	docstring inaccuracy — `non_blocking=True` events are accurate, not a "lower bound"
> **I1**	`wrap_h2d({"input": x, "labels": y})` — `batch["input"]` fails because `__getitem__` isn't forwarded
> **I2**	destination string is `"cudacheckpoint"` or `/some/cuda_path/` — passes substring check, fails as a real device
> Each of those is wired up as a 5-15 line verification recipe inside the inline review comments (look for the **Verification** section). Running those recipes against the current PR head reproduces the bug; running them against a fixed branch shows the expected behavior.
> 
> Concretely on a Tesla T4 against the current PR head:
> 
> * **C1** recipe yields 1 h2d_time event for a D2D copy (expected: 0)
> * **C2** recipe yields 20 h2d_time events for a 20-param `model.to(device)` (expected: 0)
> * **C3** recipe yields 2 events for a single wrapped transfer (expected: 1)
> 
> For reference, a branch with all 8 fixes applied as atomic commits is at https://github.com/Pendu/traceml/tree/pr-87-review-suggestions — each commit is prefixed `C1`, `C2`, `C3`, etc. to map back to the review. Feel free to cherry-pick, replicate, or take a different approach.
> 
> Happy to walk through any specific finding if the verification recipe isn't clear — feel free to push back if you think any of these aren't worth fixing.

I believe all of the reported issues are already addressed, and the requested fixes plus regression tests are already in place. I’ve also tried to cover the obvious edge cases discussed in review, and I’m not currently seeing any remaining gaps in the implementation.

That said, if you spot any additional edge cases or scenarios that should be handled differently, I’m happy to follow up with further fixes or tests.

Pendu:
> > Thanks for running the demo and posting the SQLite dump, @ppraneth — that's confirmation that the data plane works end-to-end. A few observations, then a note on what the review is actually asking for.
> > ## What your run validates ✅
> > 
> > * `_traceml_internal:h2d_time` fires every step (`n_calls=2`, matching the two
> >   `batch_x.to()` / `batch_y.to()` calls in the demo).
> > * Event lands on `cuda:0` with `is_gpu: true`, ~0.1ms per step — sane on this
> >   hardware.
> > * TLS gate works: `model.to(device)` outside `trace_step` did not pollute
> >   the per-step counts.
> > 
> > I got the same shape on Tesla T4 with the loop extended to 2000 steps: median 0.16ms, p95 0.25ms. So the happy-path patch behavior is solid.
> > ## What this run doesn't cover
> > The h2d_timing_demo deliberately exercises only the textbook usage: CPU source, CUDA destination, tensor batches (not dicts), no `wrap_h2d` mixed with `mode="auto"`. The review findings target paths the demo never touches:
> > Review item	Scenario that triggers it (not in the demo)
> > **C1**	`cuda_tensor.to('cuda:1')` — a D2D peer copy gets counted as h2d_time
> > **C2**	`model.to(device)` issued _inside_ `trace_step` — inflates `n_calls` by parameter count
> > **C3**	`wrap_h2d(x)` _before_ `traceml.init(mode="auto")` — the same transfer is timed twice
> > **C4**	docstring inaccuracy — `non_blocking=True` events are accurate, not a "lower bound"
> > **I1**	`wrap_h2d({"input": x, "labels": y})` — `batch["input"]` fails because `__getitem__` isn't forwarded
> > **I2**	destination string is `"cudacheckpoint"` or `/some/cuda_path/` — passes substring check, fails as a real device
> > Each of those is wired up as a 5-15 line verification recipe inside the inline review comments (look for the **Verification** section). Running those recipes against the current PR head reproduces the bug; running them against a fixed branch shows the expected behavior.
> > Concretely on a Tesla T4 against the current PR head:
> > 
> > * **C1** recipe yields 1 h2d_time event for a D2D copy (expected: 0)
> > * **C2** recipe yields 20 h2d_time events for a 20-param `model.to(device)` (expected: 0)
> > * **C3** recipe yields 2 events for a single wrapped transfer (expected: 1)
> > 
> > For reference, a branch with all 8 fixes applied as atomic commits is at https://github.com/Pendu/traceml/tree/pr-87-review-suggestions — each commit is prefixed `C1`, `C2`, `C3`, etc. to map back to the review. Feel free to cherry-pick, replicate, or take a different approach.
> > Happy to walk through any specific finding if the verification recipe isn't clear — feel free to push back if you think any of these aren't worth fixing.
> 
> I believe all of the reported issues are already addressed, and the requested fixes plus regression tests are already in place. I’ve also tried to cover the obvious edge cases discussed in review, and I’m not currently seeing any remaining gaps in the implementation.
> 
> That said, if you spot any additional edge cases or scenarios that should be handled differently, I’m happy to follow up with further fixes or tests.

You're absolutely right, @ppraneth — apologies for the confusion. I re-pulled the PR
head (c7ddefc) just now and re-ran each verification recipe. Every fix is accounted. 

I had a stale snapshot of your branch when I wrote the previous reply. 

I'll switch my review verdict to Approve.

Pendu (approved):

PR #96, opened 2026-04-28
diagnostics cleanup

abhinavsriva:


abhinavsriva:
Made diagnostics/common.py the single home for DiagnosticIssue, DiagnosticResult, DiagnosticRule and sort_issues

Added src/traceml/core/ (protocols)

PR #109, opened 2026-05-08
Add .yaml support

### LINKED ISSUE #85: Add traceml.yaml for user-facing TraceML defaults
Add support for a single user-authored traceml.yaml at the project root so TraceML can load default runtime and aggregator settings without requiring long CLI flags or environment variable setup. This should map onto the existing TRACEML_* settings already used by the launcher, runtime executor, and aggregator entrypoint, while preserving current behavior when no config file is present.

The config loader should follow a simple precedence model: 

#### Configuration precedence:

- Command-line flags have the highest priority.
- Environment variables override values from traceml.yaml.
- traceml.yaml overrides built-in defaults.
- Built-in defaults are used only when no higher-priority source provides a value.

### Acceptance Criteria:

- TraceML reads traceml.yaml from the project root
- Existing CLI and env-var behavior continues to work
- CLI flags override env vars, env vars override YAML, YAML overrides defaults
- No runtime behavior changes when no config file exists
- No new runtime connection logic in this issue

ppraneth:
Closes #85


abhinavsriva (approved):
Follow-up: validate resolved `mode: dashboard` after config resolution so YAML/env dashboard mode gets the same dependency preflight as CLI `--mode dashboard`. Also add user-facing docs and a commented `traceml.yaml` example.

PR #130, opened 2026-05-25
Store batch size samples in DB for all ranks 

### LINKED ISSUE #81: Store batch size samples in DB for all ranks
### Description:

Add instrumentation to record input batch size in bytes. The sampler should aggregate at the step/rank level using sum so gradient accumulation is handled naturally, then the aggregator should receive the sample and persist it to the database for all ranks, similar to how other metrics are stored today.

This is a storage-only change for now. No rendering or dashboard UI is needed yet.

### Acceptance Criteria:

Batch size in bytes is captured during training.
The sampler aggregates batch size per step, rank using sum (hint: step timing samples)
The aggregated sample is transported downstream like existing TraceML samples.
The aggregator writes the batch-size data into the database for all ranks.
No renderer or visualization work is added in this issue.

ppraneth:
Closes #81

Adds input batch size in bytes as a per-step, per-rank metric. Storage only, no rendering work.

What it does:
- H2D auto-patch records `element_size * numel` bytes whenever a tensor moves from CPU to CUDA inside `trace_step`.
- `wrap_h2d` does the same for manual mode (tensor + dict/list/tuple containers).
- New `BatchSizeSampler` sums bytes within a step so gradient accumulation is handled naturally, then emits one record per (step, rank).
- New SQLite projection writer persists rows to `batch_size_samples` for every rank, same retention rules as other per-rank tables.

Files:
- new: `utils/batch_size.py`, `samplers/batch_size_sampler.py`, `samplers/schema/batch_size_schema.py`, `aggregator/sqlite_writers/batch_size.py`
- edited: `h2d_auto_timer_patch.py`, `sdk/wrappers.py`, `utils/flush_buffers.py`, `runtime/sampler_registry.py`, `aggregator/sqlite_writer.py`
- tests: `tests/test_batch_size_sampler.py` (24 tests), `tests/runtime/test_sampler_registry.py` (updated tuples)
- demo: `tests/batch_size_demo.py`

Verified on RunPod:
- Single GPU, batch 512, float32 input + int64 label, grad accum = 2.
- Got 30 rows, `bytes_total = 2,105,344`, `n_transfers = 4` per step. Matches expected `2 * (512*512*4 + 512*8)`.


Pendu:
Hi @ppraneth, thanks for this contribution, and sorry it sat unreviewed for so long. The sampler, schema, and transport layers were solid, and we want the feature in the 0.3.6 release, so rather than send you a change list after two months of base drift we finished the branch directly. Your commits are untouched; our changes are separate commits on top.

What changed on top of your work:

1. Merged `version_0.3.6` into the branch (no history rewrite). The base had moved nearly 80 commits since June, including a registry cleanup that removed the layer samplers, so the registry and its test were reconciled with that.
2. Fixed the SQLite projection writer contract. The writer loop calls `build_rows(envelope=...)` with a `TelemetryEnvelope`, but `batch_size.py` still took a flat `payload_dict`. The resulting `TypeError` was swallowed by the writer's best-effort `except`, so `batch_size_samples` was created but never populated. It is now ported to the envelope contract like `step_time.py`, with an end-to-end projection test that fails on the old contract.
3. Moved the measurement from the H2D `.to()` patch to the dataloader fetch (the automatic `DataLoader` patch and `wrap_dataloader_fetch`). Same numbers on GPU runs, but it also works for CPU-only training, recording gates on the same state as `timed_region`, and it cannot double-count container batches whose `.to()` routes through the patched `Tensor.to`. This placement came out of maintainer discussion about where the metric should live.
4. Renamed `n_transfers` to `n_fetches` to match the new semantics, and added the new test files to the CI integration job.
5. Hardened the capture path after a review pass. Batch sizing never raises into the user's training loop, and sizing is skipped entirely outside the recording window. A wrapped iterator now defers to the auto patch instead of double-counting the same fetch, the fetch buffer is bounded like the queue, rows carry the step-flush timestamp the schema documents, and the sampler reports pending batches during the final drain, like `StepTimeSampler`.

One documented limitation, shared with the existing `dataloader_next` timing events: fetches from any dataloader inside the recording window are attributed to the next flushed step, so an eval loop between training steps inflates that step's totals. That behavior is inherited from the timing design and is noted in the schema docstring; a follow-up will look at it for both metrics together.

Validation: `traceml run tests/batch_size_demo.py` on a CPU-only machine persists one row per step per rank with `bytes_total=2,105,344` and `n_fetches=2`, matching the expected `2*(512*512*4 + 512*8)` from your original RunPod run. Full test suite: 923 passed vs 890 on the base branch, no new failures.

abhinavsriva (dismissed):

abhinavsriva:
@Pendu  we do this later so I changed again the base

PR #135, opened 2026-05-28
feat: add callback-based Hugging Face integration

### LINKED ISSUE #88: Evaluate and add callback-based Hugging Face integration
TraceML’s Hugging Face integration currently relies on a Trainer subclass (TraceMLTrainer). This works, but it requires users to replace their trainer class. We should evaluate whether a transformers. TrainerCallback can provide reliable-enough step boundaries for TraceML diagnostics and, if viable, introduce it as the preferred Hugging Face integration path.

The existing TraceMLTrainer subclass can be supported for backward compatibility, if needed. 

**Acceptance Criteria**

- Add TraceMLTrainerCallback for standard transformers.Trainer.
- Users can enable TraceML with Trainer(..., callbacks=[TraceMLTrainerCallback()]).
- The callback works on a tiny local HF model/dataset test without downloading external models.
- Step timing and memory samples are emitted and appear in TraceML’s final summary.
- Compare callback behavior against current TraceMLTrainer on the same tiny test setup.
- Document any limitations around callback timing, optimizer timing, or gradient accumulation.
- Keep TraceMLTrainer supported, either as:
       - a thin wrapper that installs TraceMLTrainerCallback, or
       - a fallback subclass if callbacks are not precise enough.
- Update Hugging Face docs/examples to prefer the callback path only if validation passes.

ssrhaso:
Introduces TraceMLTrainerCallback as the preferred TraceML entry point for transformers.Trainer. The callback is a pure bracket around trace_step (opened in on_step_begin, closed in on_step_end), so step memory tracking, counter advancement, auto-timers, and per-step flush are all owned by trace_step rather than duplicated in the callback.

TraceMLTrainer is preserved as a thin wrapper that auto-installs the callback with a dedup guard, so existing users keep working without code changes.

Closes #88

## What changed

- New `TraceMLTrainerCallback`, a `transformers.TrainerCallback` users register with `Trainer(..., callbacks=[TraceMLTrainerCallback()])`. This is now the preferred Hugging Face integration path.
- `TraceMLTrainer` is collapsed to a thin wrapper that auto-installs the callback on construction, with a dedup guard for users who pass the callback explicitly. Its public API (`traceml_enabled`, `traceml_kwargs`) is unchanged.
- Example, integration docs, and public-API docs updated to lead with the callback. Limitations of the callback vs. the subclass are documented explicitly: step granularity, optimizer-timing mode dependency, exception safety, FSDP wrapped-model attribution, and callback registration timing.

## Why

The Trainer-subclass approach forces users to swap their trainer class, which conflicts with Hugging Face's recommended extension surface and blocks users who already subclass `Trainer` themselves. A `TrainerCallback` is the idiomatic integration point and composes with everything else. The callback also reuses TraceML's existing `trace_step` machinery rather than re-implementing it, which is what made the original subclass fragile.

## How I tested

- `pytest tests/integrations/test_hf_trainer.py -v` passes 5/5, including:
  - strict per-optimizer-step memory-event count (gate against double-recording)
  - gradient-accumulation forward-event folding (gate against losing sub-phase semantics)
  - wrapper-vs-direct-callback equivalence (regression gate for the thin wrapper)
  - dedup-guard regression (user passing the callback explicitly to TraceMLTrainer)
- `python examples/huggingface_trainer_minimal.py` runs 200 steps end-to-end on CPU with no errors.
- `black`, `ruff`, `isort` clean on touched files.
- `mkdocs build --strict` clean.
- All pre-commit hooks pass.

- [x] Unit tests
- [x] Integration or smoke test
- [x] Example training script
- [ ] Docs-only change
- [ ] Not run; reason:

## Runtime impact

- [x] No training-path runtime impact
- [ ] May affect training-path overhead
- [ ] May affect distributed launch, telemetry, or aggregator behavior
- [ ] Not applicable

Notes:

The callback brackets the same `trace_step` context manager the previous `TraceMLTrainer.training_step` override used, so per-step overhead is identical. One behavioral change worth flagging: with `gradient_accumulation_steps > 1`, one TraceML step now equals one optimizer step (the previous subclass counted each micro-batch as its own TraceML step). Per-step memory, step timing, and step-counter advancement therefore fire `1/grad_accum` as often. Forward and backward sub-phase events fold across the accumulated micro-batches within that step. This is documented as a limitation in the Hugging Face integration docs.

## Documentation

- [ ] README updated
- [x] User docs updated
- [x] Examples updated
- [x] API docs updated
- [ ] Not needed

## Risk checklist

- [x] Does not add unnecessary CUDA synchronizations
- [x] Does not add blocking I/O on the training path
- [x] Fails safely without crashing user training
- [x] Keeps new dependencies optional unless discussed
- [x] Redacts or avoids sensitive training-environment data in examples

## Screenshots or output
image


Pendu:
Thanks for this, @ssrhaso — clean work, and the right architecture. Making the callback a pure bracket around `trace_step` (open in `on_step_begin`, close in `on_step_end`) keeps it a thin lifecycle adapter and lets `trace_step` retain ownership of step timing, memory, counter advance, and flush — so there's nothing to drift out of sync. I verified locally that global-step alignment holds: with `gradient_accumulation_steps=4` the TraceML step counter advances exactly once per optimizer step, folding the micro-batch forward/backward events into one step as documented. 👍

### Things I particularly liked
- **Self-heal at `on_step_begin`** (`:112`): closing a stale `trace_step` before opening a new one is a thoughtful guard against the auto-timer thread-local flags staying armed when a step raises — HF's inner loop has no `try/finally`, so `on_step_end` doesn't fire on a crash. The inline comment is accurate.
- **Back-compat fidelity**: preserving the original `traceml_kwargs`-gated hook attachment and the `_attached_model_id` re-attach logic means existing `TraceMLTrainer` users see no behavior change.
- **Tests** are strong — the strict one-`StepMemoryEvent`-per-step count, the grad-accum forward-fold, the wrapper-vs-callback equivalence, and the dedup-guard regression are all well-chosen gates.

### Suggestions

1. **(worth addressing) Also self-heal at `on_train_begin`.** The `on_step_begin` close covers most cases, but one window is left open: with `eval_on_start=True`, HF runs evaluation *between* `on_train_begin` and the first `on_step_begin`. If a callback instance is reused and a previous run crashed mid-step, the leaked auto-timer flag is still armed during that eval, so the `eval_on_start` forward passes get timed into the orphaned step. Adding `self._close_step_cm_safely()` at the top of `on_train_begin` (alongside the existing `on_step_begin` close) closes that window — cheap, no downside, robust at both boundaries.

2. **(minor) Read `TRACEML_DISABLED` dynamically.** It's captured as a module-level constant at import (`:10`), so toggling the env var after import (notebooks/tests) isn't honored. `trace_step`/`trace_model_instance` already read it per call, so a tiny helper that re-reads `os.environ` would make the integration layer consistent. (Functionally mitigated today since `trace_step` no-ops internally — so robustness, not a bug.)

3. **(minor) Two test additions** would tighten the gates: a `TRACEML_DISABLED=1` no-op test (assert the counter doesn't advance and no events are emitted), and extending the grad-accum test to also assert exactly one optimizer event per step plus `grad_accum` backward events (it currently checks forward counts only, so a dummy/missing optimizer event or dropped backward timing would slip through).

4. **(minor) Guard `model is None` in `on_step_begin`** before `trace_step(model)` — a `None` model currently raises inside `StepMemoryTracker` (caught, but it costs a lost step + log noise). Effectively unreachable in standard `Trainer` flow, so purely defensive.

### Not blocking / broader
- The crash-recovery close uses `__exit__(None, None, None)`, which resumes `trace_step` past its `yield` and marks the *failed* step as completed (advancing the counter + flushing partial data). That's really an SDK-level property of `trace_step` (advance-after-yield), so probably best addressed there rather than in the callback — flagging for awareness.
- Per-layer hooks attach to `callback_handler.model` (the raw model), not the DDP/Accelerate-wrapped model the original subclass saw via `training_step`. Usually transparent, but worth a doc note for users relying on wrapped-model attribution.

Overall this is a really nice contribution. The only item beyond polish is #1 (the `on_train_begin` self-heal). Thanks again! 🙏

ssrhaso:
> Thanks for this, @ssrhaso — clean work, and the right architecture. Making the callback a pure bracket around `trace_step` (open in `on_step_begin`, close in `on_step_end`) keeps it a thin lifecycle adapter and lets `trace_step` retain ownership of step timing, memory, counter advance, and flush — so there's nothing to drift out of sync. I verified locally that global-step alignment holds: with `gradient_accumulation_steps=4` the TraceML step counter advances exactly once per optimizer step, folding the micro-batch forward/backward events into one step as documented. 👍
> 
> ### Things I particularly liked
> * **Self-heal at `on_step_begin`** (`:112`): closing a stale `trace_step` before opening a new one is a thoughtful guard against the auto-timer thread-local flags staying armed when a step raises — HF's inner loop has no `try/finally`, so `on_step_end` doesn't fire on a crash. The inline comment is accurate.
> * **Back-compat fidelity**: preserving the original `traceml_kwargs`-gated hook attachment and the `_attached_model_id` re-attach logic means existing `TraceMLTrainer` users see no behavior change.
> * **Tests** are strong — the strict one-`StepMemoryEvent`-per-step count, the grad-accum forward-fold, the wrapper-vs-callback equivalence, and the dedup-guard regression are all well-chosen gates.
> 
> ### Suggestions
> 1. **(worth addressing) Also self-heal at `on_train_begin`.** The `on_step_begin` close covers most cases, but one window is left open: with `eval_on_start=True`, HF runs evaluation _between_ `on_train_begin` and the first `on_step_begin`. If a callback instance is reused and a previous run crashed mid-step, the leaked auto-timer flag is still armed during that eval, so the `eval_on_start` forward passes get timed into the orphaned step. Adding `self._close_step_cm_safely()` at the top of `on_train_begin` (alongside the existing `on_step_begin` close) closes that window — cheap, no downside, robust at both boundaries.
> 2. **(minor) Read `TRACEML_DISABLED` dynamically.** It's captured as a module-level constant at import (`:10`), so toggling the env var after import (notebooks/tests) isn't honored. `trace_step`/`trace_model_instance` already read it per call, so a tiny helper that re-reads `os.environ` would make the integration layer consistent. (Functionally mitigated today since `trace_step` no-ops internally — so robustness, not a bug.)
> 3. **(minor) Two test additions** would tighten the gates: a `TRACEML_DISABLED=1` no-op test (assert the counter doesn't advance and no events are emitted), and extending the grad-accum test to also assert exactly one optimizer event per step plus `grad_accum` backward events (it currently checks forward counts only, so a dummy/missing optimizer event or dropped backward timing would slip through).
> 4. **(minor) Guard `model is None` in `on_step_begin`** before `trace_step(model)` — a `None` model currently raises inside `StepMemoryTracker` (caught, but it costs a lost step + log noise). Effectively unreachable in standard `Trainer` flow, so purely defensive.
> 
> ### Not blocking / broader
> * The crash-recovery close uses `__exit__(None, None, None)`, which resumes `trace_step` past its `yield` and marks the _failed_ step as completed (advancing the counter + flushing partial data). That's really an SDK-level property of `trace_step` (advance-after-yield), so probably best addressed there rather than in the callback — flagging for awareness.
> * Per-layer hooks attach to `callback_handler.model` (the raw model), not the DDP/Accelerate-wrapped model the original subclass saw via `training_step`. Usually transparent, but worth a doc note for users relying on wrapped-model attribution.
> 
> Overall this is a really nice contribution. The only item beyond polish is #1 (the `on_train_begin` self-heal). Thanks again! 🙏

Thanks for the review, @Pendu . Follow-up commit pushed. 
Let me know if any other issues

abhinavsriva:
@ssrhaso @Pendu I think dataloader is not wrapped. Ideally traceml.init is required to patch dataloader and then it is enabled and wraps __iter__ . It might be better to introduce init separetly for the hugging face and the use call back something like in lightining so it is cleaner for user.

ssrhaso:
> @ssrhaso @Pendu I think dataloader is not wrapped. Ideally traceml.init is required to patch dataloader and then it is enabled and wraps **iter** . It might be better to introduce init separetly for the hugging face and the use call back something like in lightining so it is cleaner for user.

fixed in the latest commit. i agree that the callback can't wrap the dataloader on its own, it's just a per-step bracket around trace_step so the fetch patch was never guaranteed.

added huggingface.init() like lightning's, and moved the docs/example to the init() + callback flow. one difference: hf's init() uses mode="auto" instead of lightning's selective set, since trace_step drives forward/backward through the patch-gated auto-timers (no-ops without the global patches), whereas lightning's callback times those itself. added a test that init() turns on the dataloader + h2d patches. lmk if youd want it shaped differently.

@abhinavsriva

ssrhaso:
@Pendu @abhinavsriva just bumping this one, let me know if any issues still etc...

Pendu:
Thanks @ssrhaso for the round-trip. I've gone through the latest commits and from my side this is a green light. ✅
All the earlier review points are addressed.  The new huggingface.init() cleanly solves the DataLoader concern @abhinavsriva raised: it makes the process-wide patches ( DataLoader fetch, H2D, and the forward/backward/optimizer auto-timers trace_step arms) explicit and deterministic instead of relying on import order, and it mirrors the Lightning integration's init() + callback pattern. The init() test asserts patch_dataloader/patch_h2d, the full suite passes, and docs/example/public-API all lead with the init() + callback flow now Idiomatic, well-tested work.

One optional, non-blocking nit (take it or leave it): test_hf_init_enables_dataloader_and_h2d_patches asserts patch_dataloader and patch_h2d; since init()'s docstring also promises the forward/backward auto-timers, you could add assert config.patch_forward / assert config.patch_backward there for completeness. The grad-accum test already exercises forward/backward indirectly, so this is purely belt-and-suspenders.

Forward pointer: the DataLoader thread here surfaced something broader on our end — we're kicking off an instrumentation-completeness effort so each integration's telemetry streams are explicitly guaranteed and self-verified (rather than depending on import order). Your PR plus @abhinavsriva's catch are exactly what led to it; I'll drop the issue link here once it's filed.

thanks again. 🙏

Pendu:
Follow-up filed → #141: the instrumentation-completeness hardening this thread seeded — guaranteeing the DataLoader-fetch stream + a self-verifying telemetry conformance gate. Stage 1 is implemented; PR incoming.

PR #157, opened 2026-06-15
Dashboard robustness: fix headless/port-conflict hang, error surfacing, leak prune, staleness (TRA-68, TRA-75)

Pendu:
## Problem

`traceml run --mode dashboard` runs the NiceGUI web UI inside the aggregator
process. Today `start()` blocks until a **browser renders a page**, so:

- On a headless or remote box (no browser to auto-open), the aggregator blocks
  forever and **all telemetry is dropped**.
- On a port conflict, uvicorn calls `sys.exit(1)` (raising `SystemExit`, which
  the existing `except Exception` does not catch), so `start()` also hangs
  forever.

Port and browser-auto-open were also hardcoded (8765 / always open), which is
wrong for remote use.

## What changes

**Robustness (TRA-68)**
- `start()` now waits for the **server socket to be listening** (not for a
  browser), with a bounded timeout, and **always returns** so training never
  blocks. Startup is classified READY / FAILED / TIMEOUT and logged
  accordingly (new `server_readiness.py`, unit-tested). `_ui_ready` stays a
  pure compute gate.
- Catches `BaseException` around `ui.run()` so a port-conflict `SystemExit` is
  logged instead of wedging the daemon thread; a deterministic pre-check
  reports an in-use port cleanly.
- Subscriber/timer leak fixed: `app.on_disconnect` prunes a client's
  subscribers + timer (previously they accumulated across reconnects).
- Section compute errors are surfaced in-card (typed `LayoutError`) instead of
  being swallowed as a generic "could not update".
- Staleness indicator: a wedged display loop shows "stale Ns" instead of
  looking frozen-fresh.

**Config** — `dashboard_port` and `dashboard_auto_open` are configurable via the
existing CLI / `TRACEML_*` / `traceml.yaml` precedence
(`--dashboard-port`, `--no-dashboard-auto-open`).

**Cleanup (TRA-75)** — removes 3 orphaned dashboard modules and unused layout
constants.

Commits are split for review: (1) config, (2) robustness, (3) dead-code.

## Testing

```
pytest -q                  -> 433 passed, 1 skipped, 1 pre-existing failure*
new dashboard/config tests -> 37 passed (6 new test files)
ruff / black / isort       -> clean (line length 79)
```
\* `test_hf_trainer_callback_grad_accum_folds_microbatches` fails on
transformers 5.9.0 independently of this PR (verified against the base commit);
it touches none of these files.

Live behavior:
- headless / free port: `start()` returns READY in ~0.1s (was: hang forever)
- port conflict: `start()` returns FAILED in ~0.0s with a clear message
  (was: hang forever)

## Known limitations (follow-ups)
- Staleness catches a wedged display loop, not "telemetry stopped while the
  aggregator is alive" (needs a per-payload data timestamp).
- The staleness chip styling is intentionally minimal (a visual redesign is a
  separate PR).
- No in-browser render test (NiceGUI's async `User` fixture needs
  pytest-asyncio, not added here).

Refs: TRA-68, TRA-75


Pendu:
**Design notes.** A detailed write-up of this change (requirements, runtime scenarios, the key decisions and their rationale, quality/testing, and risks) is maintained in our internal tracker (refs `TRA-68` / `TRA-75`).

In short: the root cause was that the dashboard's `start()` waited for a *browser* to render a page rather than for the server socket to come up. So on a headless or remote box (no browser to auto-open), or on a port conflict (uvicorn raises `SystemExit`, which `except Exception` did not catch), the aggregator blocked forever and dropped all telemetry. The fix waits on socket-readiness with a bounded timeout that always returns (classified READY/FAILED/TIMEOUT), and catches `BaseException` around `ui.run()`. The remaining commits add configurable port/auto-open, real in-card error surfacing, an `app.on_disconnect` subscriber/timer-leak prune, and a UI-side staleness indicator. `_ui_ready` stays a pure compute gate, and the renderer/threading/registry plumbing is untouched.

Pendu:
**Verification update.** Beyond the test suite, I exercised this live:

- **Real browser (Playwright/Chrome):** the dashboard renders correctly against a real telemetry DB (full page, all sections) with **0 console errors / 0 warnings**. This covers the page-render path the unit tests can't reach: the `pages.py` change (`build_top_tabs`) and the dead-code removal not breaking the live page.
- **Staleness indicator** shows `stale Ns` when data stops flowing, and is empty when fresh.
- **End-to-end via the real launcher:** `traceml run --mode dashboard --dashboard-port 8796 --no-dashboard-auto-open` brings the dashboard up on the chosen port (8796, not the 8765 default) and shuts down cleanly, exercising the whole CLI -> env -> aggregator -> driver config chain.

Remaining: an *automated* in-browser render test is still future work (NiceGUI's async `User` fixture needs pytest-asyncio).

Pendu:
Design write-up for this PR — goals, constraints, building-block + runtime view, ADRs, and the verification log — is tracked in our internal knowledge base. Happy to walk through it in review.

abhinavsriva (approved):

PR #163, opened 2026-06-17
feat: move runtime startup into traceml.init()

### LINKED ISSUE #84: Move TraceML runtime startup into user code via traceml.init(...)
### Description:
Update TraceML so the runtime is started from the user’s Python code instead of being launched by traceml run. Today, traceml run starts the runtime and then executes the user script via runpy, but the new model should let users launch their training script directly with python or torchrun, and then call traceml.init(...) inside their code.

In this design, traceml.init(...) becomes responsible for starting the TraceML runtime threads and connecting to the aggregator over TCP using a configurable host and port. If the aggregator is not available, initialization should retry for a short bounded period and then fail with a clear message telling the user to start traceml serve or disable tracing.

This issue is about moving runtime ownership into the training process itself. It should preserve the existing no-op disabled path and avoid hanging forever when the aggregator is missing. traceml run should still remain available for compatibility, but the preferred path becomes direct execution of user code with explicit traceml.init(...).

### Acceptance Criteria:

- traceml.init(...) starts the TraceML runtime from inside user code
- traceml.init(...) connects to a running aggregator over TCP
- TCP host/port are configurable
- If the aggregator is missing, initialization fails clearly after bounded retries
- Disabled tracing remains a no-op
- traceml run remains available for backward compatibility
- No UI/rendering changes in this issue

ppraneth:
Makes `traceml.init(...)` start the TraceML runtime from inside user code. Now a training script can be launched directly with `python` or `torchrun` and just call `traceml.init(...)` to begin tracing. Before this, the runtime was only started by
the `traceml run` executor.

`init()` now:

- starts the per process runtime (samplers + telemetry) via the existing `lifecycle.start_runtime(...)`
- connects to a running aggregator over TCP, with a bounded retry, and fails with a clear message if it is not reachable
- takes a configurable `aggregator_host` and `aggregator_port` (falls back to `TRACEML_*` env vars, then defaults)
- takes a `disabled` flag (or `TRACEML_DISABLED=1`) that keeps tracing a full no-op
- skips runtime startup when a runtime is already running in the process, so `traceml run` is unchanged

`traceml run` still works exactly as before. No UI or rendering changes.

Closes #84

## Changes made:

- `src/traceml_ai/runtime/lifecycle.py`: add `wait_for_aggregator(...)` (bounded TCP preflight) and a per process active runtime registry (`get_active_runtime_handle()`); `start_runtime()` registers its handle
- `src/traceml_ai/sdk/initial.py`: `init()` and `start()` gain `disabled`, `aggregator_host`, `aggregator_port`, `connect_timeout_sec`, `connect_retry_interval_sec`; add the runtime bootstrap (preflight, start, atexit shutdown) and the disabled no-op path
- `src/traceml_ai/api.py`: forward the new params
- tests: new `tests/sdk/test_init_runtime.py`, lifecycle tests for the preflight and handle registration, and three existing patch policy tests updated to stub the runtime bootstrap

## Acceptance criteria(acc to issue)

- [x] `traceml.init(...)` starts the runtime from user code
- [x] `traceml.init(...)` connects to a running aggregator over TCP
- [x] TCP host and port are configurable
- [x] missing aggregator fails clearly after bounded retries
- [x] disabled tracing stays a no-op
- [x] `traceml run` still works for backward compatibility
- [x] no UI or rendering changes

## Testing

pytest tests/sdk tests/runtime/test_lifecycle.py tests/test_h2d_timing.py -q

Also tested end to end on a Linux GPU box: started a standalone aggregator, ran `src/dev/scenarios/cnn_mnist_deep.py` directly with `python` so `traceml.init()` started the runtime and connected, and got a full summary over 501 steps via `traceml view`. Confirmed the bounded failure error when no aggregator is running, and that `init(disabled=True)` returns a no-op config.


abhinavsriva:
@ppraneth  awesome. This will be helpful.

abhinavsriva:
The `traceml.init()` runtime looks right, but the direct-launch workflow is incomplete.

Blocking items:

1. `traceml.init()` tells users to run `traceml serve`, but `traceml serve` does not exist.

   Please add a real `traceml serve` command, or change the design/error/docs. I think adding `traceml serve` is the correct fix.

2. `traceml serve` should be a clean aggregator owner, not extra wiring around `traceml run`.

   Please reuse the existing aggregator lifecycle instead of adding another parallel path. There is already useful code in:

   - `src/traceml_ai/runtime/lifecycle.py::start_aggregator`
   - `src/traceml_ai/aggregator/aggregator_main.py`
   - existing launcher config/env handling

   Expected behavior for `serve`:

   - starts only the aggregator
   - binds host/port from CLI/config
   - prints the reachable endpoint clearly
   - blocks until SIGINT/SIGTERM
   - shuts down cleanly
   - writes/logs fatal aggregator errors the same way the existing aggregator process does
   - preserves final-summary behavior on shutdown
   - does not launch or wrap the user training script

3. Please document the direct launch contract.

   This PR changes the preferred user model, but no README/docs files are changed. README still shows:

   ```bash
   traceml run train.py
   ```

   The docs need to show the new flow, for example:

   ```bash
   # terminal 1
   traceml serve --aggregator-host 127.0.0.1 --aggregator-port 29765

   # terminal 2
   python train.py
   ```

   and for torchrun:

   ```bash
   traceml serve --aggregator-bind-host 0.0.0.0 --aggregator-host <node0-ip> --aggregator-port 29765
   torchrun ... train.py
   ```

   Also please make the configuration story explicit. In the direct `python` path, users cannot pass `traceml run` flags. Current `init()` supports explicit `traceml.init(...)` args and `TRACEML_*` env vars. If YAML config is intended to work here, please route through the existing config resolver instead of adding one-off parsing. If YAML is not intended for this issue, document that direct launch uses `traceml.init(...)`, env vars, and `traceml serve` flags. I think both should be done, meaning traceml.init should have all runtime related flags passed and if default values then use from yaml. For runtime aggregator, user can pass varaible with `traceml serve`

4. Add focused tests.

   I think two tests are enough:

   - A launcher/CLI test proving `traceml serve` is a real public command and maps host/bind-host/port/logs/run-name/mode into the aggregator settings.
   - A regression test for `start_runtime(..., fail_open=True)` registering a no-op active handle so a later `traceml.init()` does not try to start runtime again.

tests/sdk/test_init_runtime.py
abhinavsriva, line 1: Need change please see main page for comment

abhinavsriva (changes requested):

Pendu:
Thanks for the rework here, @ppraneth. serve and the `run_aggregator` extraction look solid, and the earlier review items are addressed. I did a follow-up pass on the new multi-node and `init()` surfaces and pushed four small commits straight to this branch (it has maintainer edits enabled) rather than leaving a long list. Here is what changed and why.

1. `init()` no longer crashes training when the aggregator is missing.
On this branch a bare `python train.py` (or a torchrun worker) that calls `traceml.init()` with no aggregator up raised a `RuntimeError` and stopped the run. TraceML's failure rule is that instrumentation should degrade and detach loudly, never take down the training process. So the default is now fail-open: one `[TraceML]` stderr warning, then continue as a no-op. Callers who want the old hard-fail behavior (for example CI that should fail on misconfigured telemetry) pass `traceml.init(on_missing_aggregator="raise")` or set `TRACEML_ON_MISSING_AGGREGATOR=raise`. `traceml run` keeps its existing fail-closed behavior, since there the user explicitly asked TraceML to launch the job.

This is a public-API behavior call, so flagging it explicitly: it reinterprets issue #84's "fails clearly" as "warns clearly" for the user-code path. Happy to discuss if you or @abhinavsriva  prefer fail-closed as the default.

2. Ray no longer goes silently dark. `start_runtime(fail_open=True)` (the Ray worker path) registered a no-op runtime with no output when the runtime failed to start, so a Ray run whose aggregator was down would train to completion and write an empty summary with no signal. It now emits the same one-line warning, so the fail-open path is loud everywhere.

3. `traceml serve` now waits for all ranks. `_resolve_serve_settings` did not set `expected_world_size`, so it defaulted to 1 and the aggregator finalized after the first rank reported, dropping later ranks' telemetry and skipping the missing-ranks warning. `serve` gains `--nnodes` / `--nproc-per-node` (product = rank count, matching `traceml run`), with a `TRACEML_EXPECTED_WORLD_SIZE` fallback.

4. Multi-node worker docs. The documented flow started `traceml serve --aggregator-host <node0-ip>` but then ran a bare `torchrun ... train.py` on workers. Workers resolve the host from `TRACEML_AGGREGATOR_HOST` (default `127.0.0.1`), so non-node0 workers could not reach node 0. The docs now set `TRACEML_AGGREGATOR_HOST` (and port) on the worker command and show the new `serve` world-size flags.

One item I did not change: the `init()` preflight is a bare TCP accept, so a foreign listener on the port passes the reachability check. A real session handshake is a wire-format change with backward-compat implications, so I documented it as a known limitation and left it as a follow-up rather than bundling it here. Tracked as #211.

Testing: the existing suite passes with these changes. The one failing test (`test_hf_trainer_callback_grad_accum_folds_microbatches`) fails the same way on the base commit, so it looks environmental and pre-existing rather than related to this PR. I also ran a two-GPU DDP job before and after to confirm the core diagnosis path is unchanged and the offline `init()` and `serve` paths behave as described.

Last thing: this branch is a couple of commits behind `main` (the `version_0.3.4` merge) and will need a rebase, including the `DASHBOARD_EXTRA_INSTALL_HINT` rename. Let me know if you would rather take these as suggestions instead of the pushed commits.

abhinavsriva (approved):
Approved based on @Pendu 's remark

PR #170, opened 2026-06-24
Improve storage

### LINKED ISSUE #164: Aggregator silently drops the end-of-run summary (final_summary / card) on heavy multi-rank runs
### Summary
On a heavy or long **multi-rank** `traceml run`, training completes cleanly (`rc=0`, the workload prints its `Done.` line) but the aggregator produces **no** `final_summary.json`, **no** `final_summary.txt`, and **no** `aggregator/telemetry_summary_card.json`. Only the raw telemetry SQLite remains, and it is left **uncleanly closed** (`-wal` / `-shm` sidecars present → the WAL was never checkpointed).

### Environment
TraceML on current `main`; 2× NVIDIA T4; PyTorch 2.11.0+cu130; 2-node DDP over TCP; `traceml run --nnodes 2 --mode summary`.

### Steps to reproduce
Run a 2-rank workload long enough to accumulate tens of MB of telemetry — e.g. a ~110M-parameter transformer for 15,000 steps under `traceml run --nnodes 2`. The end-of-run summary artifacts are absent; the node-0 log shows `[TraceML] SQLiteWriter thread did not terminate cleanly`.

### Evidence — the failure scales with telemetry volume
| run | steps × ranks | telemetry DB | summary written? |
|---|---|---|---|
| short 2-node | 500 × 2 | 2.3 MB | ✅ yes |
| long single-GPU | 15000 × 1 | 33 MB | ✅ yes |
| **long 2-node** | **15000 × 2** | **77 MB** | ❌ **no** |

Only the largest (77 MB) run loses its summary and logs the "did not terminate cleanly" warning.

### Root cause
At shutdown the aggregator stops the SQLite writer thread with a fixed **2.0 s** join — `SQLiteWriter.stop()` in `src/traceml_ai/aggregator/sqlite_writer.py`, invoked with its default from the aggregator shutdown in `src/traceml_ai/aggregator/trace_aggregator.py`. The writer is a **daemon** thread that drains its queue in bounded batches; with a large backlog (~77 MB, 2 ranks × 15k steps) it cannot drain within 2.0 s, so:
1. the thread is left alive (`did not terminate cleanly`),
2. the WAL is never checkpointed (the checkpoint runs only on a clean close), and
3. the subsequent `generate_summary(...)` opens a new reader on the WAL-mode DB, cannot see the still-open writer's uncommitted rows (and may hit a lock), so it fails — **silently**, because the finalize step is best-effort/swallowed.

Smaller backlogs (single-rank long, or short multi-rank) drain within 2.0 s, close cleanly, and finalize normally.

### Impact
The runs that lose their summary are exactly the high-value ones — **long, multi-node** training. Telemetry is recoverable from the raw SQLite/WAL, but the JSON/TXT/card that users and downstream tooling depend on are gone, and the DB is left uncleanly closed.

abhinavsriva:
## What changed

Made TraceML history storage projection-only and made summary-mode shutdown
deterministic for large/multi-rank runs.

- Fresh SQLite history DBs no longer create or append `raw_messages` /
  `payload_mp` raw telemetry blobs.
- SQLite history now stores only structured projection tables used by live CLI,
  dashboard, final summary, HTML reports, and compare.
- Aggregator shutdown now settles late rank telemetry, finalizes SQLite,
  checkpoints/closes WAL, then generates final summary artifacts.
- Added `rank_finished` control messages so the aggregator can distinguish
  late telemetry from completed ranks without storing control payloads.
- Added `--finalize-timeout-sec`, `TRACEML_FINALIZE_TIMEOUT_SEC`, and
  `traceml.yaml: finalize_timeout_sec`.
- Summary mode now fails clearly if TraceML cannot finalize SQLite or produce
  `final_summary.json`.
- Final retention pruning is best-effort; checkpoint/close failure remains
  fatal for summary mode.

## Why

Long multi-rank runs could finish training successfully but lose
`final_summary.json`, `final_summary.txt`, and summary-card artifacts because
the SQLite writer had only a fixed short shutdown join. Large telemetry
backlogs could leave the writer thread alive and WAL sidecars uncheckpointed.

Raw telemetry blobs were also being stored in SQLite even though normal product
paths read projection tables, increasing shutdown pressure without helping
live UI, dashboard, final summary, HTML, or compare.

## How I tested

- [x] Tests added or updated
- [x] Existing tests run
- [x] Manual smoke test
- [ ] Not run; reason:

Targeted suites:

```bash
  tests/aggregator tests/telemetry tests/runtime tests/config -q
```

Result:

```text
151 passed
```

Full suite:

```bash
PYTHONDONTWRITEBYTECODE=1 pytest -p no:cacheprovider tests -q
```

Result:

```text
507 passed, 8 skipped
```

Manual summary smoke:

```bash
PYTHONPATH=src python -c 'from traceml_ai.launcher.cli import main; main()' \
  run examples/summary_logging_minimal.py \
  --mode=summary \
  --logs-dir /tmp/traceml-finalize-smoke-prod \
  --run-name finalize_smoke_prod \
  --finalize-timeout-sec 30 \
  --args --steps 1
```

Smoke checks confirmed:

```text
final_summary.json: present
final_summary.txt: present
aggregator/telemetry_summary_card.json: present
SQLite -wal sidecar: absent
SQLite -shm sidecar: absent
raw_messages table: absent
residual_ms: present
wait_ms: absent
```

Also ran:

```bash
git diff --check
```

Result: clean.

## Runtime impact

- [x] No training-path impact
- [x] May affect tracing, runtime, telemetry, or distributed behavior
- [ ] Not sure

Notes:

This changes TraceML storage and aggregator finalization behavior, not the user
training loop. Runtime now sends one best-effort `rank_finished` control message
on shutdown after its final telemetry publish. Aggregator shutdown may wait up
to `finalize_timeout_sec` to settle late telemetry and close SQLite cleanly.

`history_enabled=false` still disables SQLite history and summary generation.
`enable_logging=false` still only controls the separate local msgpack logging
path.

Existing DBs are not migrated or dropped. Old `raw_messages` tables can remain
in reused DBs, but new runs do not append raw blob rows.

## Docs

- [x] Updated
- [ ] Not needed

Updated docs mention `--finalize-timeout-sec` / `TRACEML_FINALIZE_TIMEOUT_SEC`
and explain that larger multi-node runs may need a larger finalization budget
on slow filesystems or congested networks.

Touched docs include:

- `README.md`
- `docs/user_guide/quickstart.md`
- `docs/user_guide/distributed-training.md`
- `docs/user_guide/slurm.md`

## Extra context

The shutdown contract now follows a flush/finish-style sequence:

```text
settle late TCP telemetry
consume rank_finished markers
stop TCP
drain SQLite writer queue
run final retention prune best-effort
checkpoint WAL
close SQLite
generate final summary artifacts
```

Finalization diagnostics are written under `aggregator/` as
`finalization_warning.json` or `finalization_error.json` when needed. The
SQLite finalize result records queue counts, dropped rows, checkpoint status,
and separate prune/checkpoint errors for debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ traceml run train.py --html-report
 traceml view logs/<run_name>/final_summary.json --html   # writes <...>.html
 ```
 
+For very large or slow multi-node jobs, TraceML waits at shutdown for late
+telemetry, SQLite checkpointing, and final summary writing. Tune that single
+end-of-run budget with `--finalize-timeout-sec` or
+`TRACEML_FINALIZE_TIMEOUT_SEC` when running on slow filesystems or congested
+networks.
+
 Instead of guessing where step time and memory went, you get a compact
 diagnosis at the end of every run.
 
--- a/docs/user_guide/distributed-training.md
+++ b/docs/user_guide/distributed-training.md
@@ -55,6 +55,12 @@ Override that only when needed with `--aggregator-bind-host=<bind-host>`.
 `--session-id` remains accepted as a backward-compatible alias for
 `--run-name`.
 
+At the end of a summary run, node 0 waits for rank-finished markers, drains
+late telemetry, checkpoints SQLite, and then writes `final_summary.*`. The
+default finalization budget is 300 seconds. On slow shared filesystems or
+congested networks, increase it with `--finalize-timeout-sec <seconds>` on each
+node, or set `TRACEML_FINALIZE_TIMEOUT_SEC`.
+
 ## Running on Slurm
 
 On a Slurm-managed cluster, derive these flags from the job environment instead
--- a/docs/user_guide/quickstart.md
+++ b/docs/user_guide/quickstart.md
@@ -58,6 +58,10 @@ Add `--html-report` (`traceml run train.py --html-report`) to also write a
 shareable `final_summary.html`. See
 [Reading the output](reading-output.md#shareable-html-report).
 
+TraceML finalizes summaries after training by settling late telemetry, closing
+SQLite cleanly, and checkpointing WAL. Large distributed jobs can raise that
+end-of-run budget with `--finalize-timeout-sec <seconds>`.
+
 For DDP, FSDP, and multi-node launches, see
 [Distributed Training](distributed-training.md).
 
--- a/docs/user_guide/quickstart.md
+++ b/docs/user_guide/quickstart.md
@@ -62,6 +62,12 @@ TraceML finalizes summaries after training by settling late telemetry, closing
 SQLite cleanly, and checkpointing WAL. Large distributed jobs can raise that
 end-of-run budget with `--finalize-timeout-sec <seconds>`.
 
+In `--mode=summary`, if training finishes but TraceML cannot produce
+`final_summary.json`, `traceml run` exits non-zero, so a silently missing
+summary fails loudly instead of passing. (Reused history databases keep any
+rows written by older releases; new runs only write the structured projection
+tables.)
+
 For DDP, FSDP, and multi-node launches, see
 [Distributed Training](distributed-training.md).
 
--- a/docs/user_guide/slurm.md
+++ b/docs/user_guide/slurm.md
@@ -207,6 +207,11 @@ passing the flag on every node in a symmetric launch command is harmless.
 reachable after the job ends. You can compare two runs later with
 [`traceml compare`](compare.md).
 
+Large jobs can spend time flushing late telemetry and checkpointing SQLite at
+shutdown. If your cluster has slow scratch storage or delayed cross-node TCP
+delivery, pass `--finalize-timeout-sec <seconds>` in the shared launch command
+or set `TRACEML_FINALIZE_TIMEOUT_SEC`.
+
 ## Cluster-specific notes
 
 !!! tip "Adapt these for your cluster"
--- a/src/traceml_ai/aggregator/aggregator_main.py
+++ b/src/traceml_ai/aggregator/aggregator_main.py
@@ -38,6 +38,7 @@ from traceml_ai.loggers.error_log import get_error_logger, setup_error_logger
 from traceml_ai.reporting.config import DEFAULT_SUMMARY_WINDOW_ROWS
 from traceml_ai.runtime.lifecycle import start_aggregator
 from traceml_ai.runtime.settings import (
+    DEFAULT_FINALIZE_TIMEOUT_SEC,
     AggregatorTransportSettings,
     TraceMLSettings,
 )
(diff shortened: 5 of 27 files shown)


Pendu:
Reviewed the finalization fix and the projection-only storage change, reproduced #164 on a 2-node g4dn pair (before vs after), and went ahead and implemented the follow-ups directly on this branch. The fix resolves the bug and the code reads correctly.

**Reproduction** (same 2-node pair; `before` = this PR's merge-base, `after` = this PR):

| run | steps x ranks | telemetry DB (before -> after) | summary written? (before -> after) |
|---|---|---|---|
| short 2-node | 500 x 2 | 2.39 MB -> 1.29 MB | ✅ -> ✅ |
| long single-GPU | 15000 x 1 | 35.11 MB -> 17.59 MB | ✅ -> ✅ |
| long 2-node | 15000 x 2 | 78.83 MB -> 36.63 MB | ❌ -> ✅ |

On the pre-fix code the 2-node x 15000 run reproduced #164 exactly: training finished (rc 0), the telemetry DB reached ~79 MB, the log showed `SQLiteWriter thread did not terminate cleanly`, the WAL was left uncheckpointed (`-wal`/`-shm` present), and no `final_summary.json` or card were written. On this branch the same run closes cleanly (no sidecars, no `finalization_error.json`), the DB is about half the size (projection-only), and the summary is produced.

**The core fix is correct:** `finalize()` drains, checkpoints, and closes SQLite before `generate_summary` opens its reader, so the original WAL race is gone; the reserved finalize budget and the loud-failure path are nice.

**Follow-ups I pushed to this branch** (commits e0fe8f6 + 3fd6427 -- happy to adjust or drop any):
- Post-write summary errors are no longer fatal: `generate_summary` is wrapped so a render/print error that fires after `final_summary.json` is already written downgrades to `finalization_warning.json`. The missing-file and finalize-failure paths stay fatal, so #164 itself still fails loudly.
- Lock-safe `_finished_ranks` snapshot before `sorted()` in the finalize/settle paths, to avoid a "dictionary changed size" race if a late drain runs.
- Tests: a real-writer-through-`stop()` e2e that asserts `final_summary.json` appears with no `-wal`/`-shm` (the literal inverse of #164); the clean all-ranks-finished settle path plus duplicate `rank_finished` idempotency; a multi-rank backlog drain; and a `_sqlite_finalize_budget` clamp table.
- Docs: a quickstart note that summary mode exits non-zero if it cannot produce `final_summary.json`, plus the reused-DB raw-rows note; and a back-compat code comment for older workers that do not send `rank_finished`.

Full local test suite is green with these changes (repo pre-push hooks -- isort/ruff/black/codespell/mkdocs --strict -- pass). One unrelated test, `test_hf_trainer_callback_grad_accum_folds_microbatches`, also fails on the base commit in my local env (looks like a transformers-version thing, not related to this PR).

Pendu (dismissed):

PR #182, opened 2026-07-07
Add data-loading bottleneck Colab notebook

Pendu:
## What

Adds `notebooks/`, a home for Colab-runnable demos, with the first one:
`data_loading_bottleneck.ipynb`. It walks through diagnosing and fixing a
data-loading (input-bound) bottleneck on a real ResNet-18 + Imagenette run:
train twice, change only the DataLoader, and read the before/after wall-clock
speedup and GPU utilization straight from TraceML. One-click "Open in Colab"
badge; runs on a free T4 in a few minutes.

Companion to the blog post *The Half-Idle GPU*.

## Why

TraceML's differentiator is the automatic verdict, and the fastest way to feel
it is to run one. This is the reproduce-it artifact for the case-study post and
a self-contained TraceML demo on a real workload.

## Note / question for review

This is the **first in-repo notebook**. The tooling is already set up for it
(black-jupyter + nbstripout are configured in pre-commit), but there is no
`notebooks/` directory yet. Happy to keep it as `notebooks/` (standard for
Colab demos), or convert it to a `.py` example under `examples/` if you would
rather keep the repo scripts-only. Your call.

Details:
- Outputs are stripped; the notebook is meant to be run in Colab.
- The Colab badge points to `main`, so it activates once this merges.
- ruff / isort are scoped to `src/ | dev/ | docs/`, so they do not lint `notebooks/`.

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -1,0 +1,9 @@
+# Notebooks
+
+Runnable, Colab-ready notebooks that demonstrate TraceML on real workloads.
+Each opens in Google Colab (a free T4 GPU is enough) and runs top to bottom in
+a few minutes.
+
+| Notebook | What it shows | Open |
+|---|---|---|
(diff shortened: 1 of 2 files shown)


abhinavsriva:
I would add on main readme also changed somewhere 

[Open in Colab](YOUR_NOTEBOOK_LINK)

and may 1 line somewhere on top try now or something

PR #185, opened 2026-07-07
Step time cleaning

abhinavsriva:
## What changed

Base comparison: `version_0.3.4` at `d9531cd`.
Branch head: `step_time_cleaning` at `3ca44fe`.

Note: `3ca44fe` is a merge commit that records `version_0.3.4` as merged into
this branch. Its tree is identical to the previous branch head `2466792`, so
the code diff from `version_0.3.4` is unchanged.

This PR unifies Step Time analysis around one canonical selected-clock window
pipeline shared by live CLI rendering, the NiceGUI dashboard, final summary
reporting, and Step Time diagnosis.

Compared with `version_0.3.4`:

- Added `src/traceml_ai/utils/step_time_window.py` as the shared Step Time
  pipeline:
  `raw step events -> common-step alignment -> CPU/GPU clock selection ->`
  `per-step timing -> per-rank averages -> diagnosis/display metrics`.
- Step Time now selects one diagnosis clock for the whole aligned window.
  GPU timing is used only when the window has complete GPU event timings;
  otherwise explicit CPU timing is used.
- Live CLI, dashboard, and final summary now consume the same selected-clock
  Step Time metrics instead of maintaining separate timing paths.
- Step Time diagnosis now uses the same timing basis for `input_wait`, `h2d`,
  `forward`, `backward`, `optimizer_step`, `step_time`, and
  `residual_proxy`.
- `INPUT_BOUND` and input straggler attribution now use selected-clock
  `input_wait`, not CPU-only dataloader fetch time.
- The CLI Step Time table now shows diagnosis metrics, labels input wait as
  `IW`, shows per-step averages, preserves real zero timings as `0.0 ms`, and
  displays the selected diagnosis clock.
- The dashboard Step Time hero now reads the same diagnosis metrics as the CLI
  and leaves the verdict owned by the diagnosis engine.
- Final summary schema is updated to `1.6` and Step Time output now includes
  diagnosis-clock `input_wait_ms` and `step_time_ms`.
- `dataloader_ms` remains a public compatibility key for CPU dataloader fetch
  time. `total_step_ms` remains the CPU dataloader plus CPU step-envelope
  compatibility total.
- `duration_ms` remains stored compatibility timing, but is no longer used as
  Step Time display or diagnosis fallback.
- Removed old split/compatibility paths:
  - `src/traceml_ai/reporting/sections/step_time/alignment.py`
  - `src/traceml_ai/utils/step_time_diagnosis_clock.py`
  - duration-based Step Time row readers
  - summary-only alignment model
  - duplicate Step Time compatibility aliases
  - stale CPU/GPU duplicate summary fields
- Updated Step Time, final-summary, primary-diagnosis, renderer, HTML, and
  dashboard tests around the shared selected-clock contract.
- Updated docs and schema text to distinguish CPU dataloader fetch from
  selected-clock input wait.

## Why

CPU dataloader fetch time and GPU-visible input wait are different signals.
Before this branch, live display, diagnosis, and final summary could describe
Step Time from different timing paths, which made input-bound and straggler
results harder to reason about.

This PR makes the Step Time model explicit:

`SQLite event rows -> aligned Step Time window -> selected-clock metrics -> diagnosis/display/reporting`

The main result is that CLI, dashboard, final summary, and primary diagnosis
now agree on the same Step Time evidence.

## Not handled in this PR

- No sampler, transport, launcher, training-loop, or distributed execution
  behavior is changed.
- No SQLite table schema migration or backfill is added. This uses the existing
  `events_json` CPU/GPU timing payloads and keeps stored `duration_ms`
  compatibility data in place.
- No downstream consumer migration is included beyond preserving existing
  public compatibility fields and documenting schema `1.6`.
- No new NCCL, all-reduce, synchronization, checkpointing, logging, validation,
  or framework-overhead attribution is added. Those can still appear inside
  `residual_ms` and should be split out in later PRs only with stronger
  telemetry/profiler evidence.
- No broad dashboard redesign is included. The dashboard Step Time card changes
  its metric source and KPI semantics, but larger UX/layout work is left for a
  later PR.
- The current full test suite has one stable out-of-scope HF Trainer
  integration failure:
  `tests/integrations/test_hf_trainer.py::test_hf_trainer_callback_grad_accum_folds_microbatches`.
  The branch diff does not change the HF callback/runtime capture path; this
  should be investigated separately unless reviewers want it folded into this
  PR.
- No end-to-end multi-node or long-running training smoke was run in this pass.
  The automated coverage here is focused on reporting, diagnosis, renderers,
  HTML, and the Step Time dashboard section.

## How I tested

- `PYTHONDONTWRITEBYTECODE=1 pytest -p no:cacheprovider tests/reporting/summary tests/reporting/html tests/diagnostics tests/renderers tests/display/test_model_combined_section.py -q`
  - `222 passed, 1 warning`
- `git diff --check version_0.3.4..HEAD`
  - clean
- `PYTHONDONTWRITEBYTECODE=1 pytest -p no:cacheprovider tests -q`
  - `541 passed, 1 failed, 4 warnings`
  - Failure:
    `tests/integrations/test_hf_trainer.py::test_hf_trainer_callback_grad_accum_folds_microbatches`
  - Expected two `_traceml_internal:forward_time` events per optimizer step;
    observed `[0, 0, 0]`.

- [x] Tests added or updated
- [x] Existing tests run
- [ ] Manual smoke test
- [ ] Not run; reason:

## Runtime impact

- [x] No training-path impact
- [ ] Not sure

Notes:

This changes Step Time analysis, diagnosis attribution, trend notes, live
CLI/dashboard Step Time display, final-summary Step Time values/wording, and
primary-diagnosis evidence. It does not change sampler capture, stored event
payload shape, transport, distributed launch behavior, or training execution.

## Docs

- [x] Updated
- [ ] Not needed

Docs touched include the final-summary schema, diagnosis docs, quickstart, and
Step Time related user/developer guides.

## Extra context

This PR should be reviewed against `version_0.3.4`, not `main`.

Diff summary from `version_0.3.4..step_time_cleaning`:

- 33 tracked files changed
- 1192 insertions, 1465 deletions
- Added `src/traceml_ai/utils/step_time_window.py`
- Deleted `src/traceml_ai/reporting/sections/step_time/alignment.py`
- Deleted `src/traceml_ai/utils/step_time_diagnosis_clock.py`

Commits on this branch after `version_0.3.4`:

- `fdfa0f9` duration_ms removed from summary
- `605d67d` others changed
- `fa7a572` cli fix'
- `f6c3272` fix cli dashboard
- `7ad3b5f` rendered fix
- `729ebf7` test corected
- `184a432` single data pipelines for step_time
- `626f1da` cleanup
- `2466792` change summary mode
- `3ca44fe` merged

Public final-summary keys are preserved where possible. In particular,
`dataloader_ms` remains CPU dataloader fetch time, while `input_wait_ms` is the
selected-clock timing used by input-bound diagnosis.


Pendu:
Reviewed the branch against `version_0.3.4`. The unification into one selected-clock window is the right call, and I ran the test suite plus a set of probes against the actual pipeline to check the public output. I pushed one small fix directly to the branch and want to flag the rest as open points before merge, since most of them come out of the `total_step_ms` / residual / share semantics and I did not want to guess at the intent.

## Pushed to this branch

`f869b13` html report: fall back to `dataloader_ms` for pre-1.6 phase bar.

`traceml view --html` on a stored schema-1.5 `final_summary.json` was dropping the input phase from the SVG bar entirely, because `_PHASES` now keys only on `input_wait_ms` (svg.py:18) with no fallback and `phase_bar` re-normalizes the remaining phases to 100%. A 59% input-bound run rendered as compute-dominated. The fix falls back to `dataloader_ms` when `input_wait_ms` is absent, with a test. New 1.6 reports are unaffected.

## Two correctness items to settle before merge

**1. Phase shares can exceed 100%, and public `total_step_ms` is no longer additive.**
`public_step_time_metric_values` sets `total_step_ms = dataloader_fetch + step_time_cpu` (step_time_window.py:555, both CPU, no clamp), while `step_time_ms` and `compute_ms` follow the selected clock. In a GPU-clock window with a short CPU envelope (the CPU-runs-ahead case), phase values exceed the total. Probe: a window with a 10ms CPU envelope and 49ms of GPU phases yields `total_step_ms 10.5`, `step_time_ms 50.0`, `compute_ms 48.0`, and the text report Share column prints `Step Time 476.2%` / `Compute 457.1%`. Separately, `primary_diagnosis` divides each phase by `step_time_ms` (primary_diagnosis.py:232) while `input_wait` sits outside that envelope (`total_step = input_wait + step_time`, step_time_window.py:236), so `input_wait_pct` can exceed 100 (probe: 144.7%). The old additive contract `total = dl + h2d + compute + residual` is gone. What should `total_step_ms` mean now, and what denominator should the shares use so they stay bounded?

**2. Public `residual_ms` can contradict the diagnosis that cites it.**
Diagnosis residual is the average of per-step clamps (step_time_window.py:224 plus `_average_rank_timing`); public `residual_ms` is the clamp of the average (model.py:119, `compute_residual_avg_ms`). Base kept these identical by construction. Probe: alternating steps (100 / fwd 50, then 50 / fwd 100) make RESIDUAL_HEAVY fire at a 33% share while the public row and the primary evidence report `residual_ms: 0.0`. Which residual should be canonical, so the verdict and its evidence agree?

## Edge cases and design questions

**3. A dark dataloader stream degrades every public phase metric to CPU clock.**
`REQUIRED_GPU_METRICS` includes `input_wait` (step_time_window.py:81), so a loop that never goes through `DataLoader.__iter__` (manual fetch, some Ray loops) fails GPU election for the whole window and every public phase metric becomes CPU launch-wall time. Probe: identical GPU work with and without the dataloader event gives `compute_ms 550 -> 6`, `residual_ms 5 -> 554`; a compute-bound run reads as residual-heavy in the metrics themselves. Should `input_wait` be required only when present, with `step_time` as the sole hard GPU gate?

**4. Clock election makes headline numbers bimodal across identical runs.**
`gpu_ms` is nullable by design (async CUDA-event resolution may not finish before the sampler flush), and `_select_clock_from_events` elects GPU only if every rank and step in the window has it. One missing `gpu_ms` on one trailing step flips the whole window to CPU. Probe: `step_time_ms` 100 -> 30. The summary window spans up to 10,000 steps and the live window is 100, so the two can pick different clocks for the same run. The only signal is the nested `global.window.diagnosis_clock`. Worth surfacing the clock more prominently, or excluding the bad step rather than the whole window?

**5. Pre-0.3.4 payloads render the whole step-time section as zeros.**
The window reads only `cpu_ms` / `gpu_ms`, and every released client (PyPI <= 0.3.1) emits only `duration_ms` (`git tag --contains fd2786f` is empty). A mixed-version cluster or a report regenerated over an older DB now produces an all-zero step-time section with a NO_DATA diagnosis, and the removed all-zero guard means a zeroed rank can win median-rank selection. Given the stated v0.2.x to v0.3.1 back-compat constraint, is this intended? A short stderr notice when a window is `duration_ms`-only would keep old-client users from silently seeing zeros.

**6. `compare` has no schema-version gate across the 1.5 to 1.6 break.**
`validate_summary_payload` (compare/io.py:72) checks that `schema_version` exists but never reads its value. Comparing a 1.5 baseline against a 1.6 run mixes semantics (redefined `dataloader_ms`, clamped vs unclamped total, new keys) and reports phantom deltas with no warning, and `_dominant_phase` still ranks on CPU `dataloader_ms`. Want a version-mismatch warning in the compare output?

**7. `INPUT_BOUND` thresholds were not retuned for the new denominator.**
`input_bound_share = input_wait / step_time` is unbounded and warns at `w/s >= 0.25`, which is about 20% of end-to-end wall where the old `dataloader / total_step` semantics warned at 25%. `policy.py` is unchanged and `DIAGNOSIS.md` does not state the denominator. Intended sensitivity change?

## Doc and wording, downstream of the above

These three settle once the `total_step_ms` / residual / share semantics are fixed, so I left them rather than guess:

- SCHEMA.md `phase_share` example: `residual_ms = 1.6` does not match the code formula `residual = step_time_ms - h2d_ms - compute_ms` (162.6 for the shown inputs), and a self-consistent input-bound example produces `input_wait_pct = 144.7%` under the current share denominator. The right numbers depend on item 1.
- DIAGNOSIS.md `total_step_ms = input_wait_ms + traced_step_ms` describes the diagnosis-internal total but reuses the public field name, which the reporting layer defines as CPU dataloader plus CPU envelope. Depends on item 1.
- The `INPUT_BOUND` card and verdict wording ("Input wait took X of a Y step") reads as containment even though input wait sits outside the traced step. Depends on item 1.

## Also verified

The HF grad-accum test failure reproduces identically on `version_0.3.4` with a byte-identical test file, so it is pre-existing and not from this branch. Root cause: the test registers the callback without calling `init()`, so no patches install; calling `huggingface.init()` first lights every stream as asserted. Worth fixing separately, and it points at a gap where a user on the callback-without-init path gets only `step_time` with no warning.

Pendu (approved):

PR #197, opened 2026-07-13
Feat: Deepspeed support

ppraneth:
## How I tested

- [ ] Tests added or updated
- [ ] Existing tests run
- [ ] Manual smoke test
- [ ] Not run; reason:

## Runtime impact

- [ ] No training-path impact
- [ ] Not sure

Notes:

## Docs

- [ ] Updated
- [ ] Not needed

### DIFF
--- a/docs/user_guide/integrations.md
+++ b/docs/user_guide/integrations.md
@@ -10,4 +10,5 @@ job already runs.
 | Hugging Face Trainer | `TraceMLTrainerCallback` | [Hugging Face](integrations/huggingface.md) |
 | PyTorch Lightning | `traceml_ai.integrations.lightning.init()` + `TraceMLCallback` | [PyTorch Lightning](integrations/lightning.md) |
 | Ray Train | `TraceMLTorchTrainer` | [Ray Train](integrations/ray.md) |
+| DeepSpeed | `traceml.trace_step(...)` | [DeepSpeed](integrations/deepspeed.md) |
 | W&B / MLflow | `traceml.summary()` | [W&B / MLflow](integrations/wandb-mlflow.md) |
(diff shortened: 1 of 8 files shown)


Pendu:
Thanks for this. The recipe is right and the integration reads clean. I reviewed it and ran it on GPU to confirm the timing actually lands under real DeepSpeed, then pushed a few small follow-ups to your branch (commit 9abc6ca).

**What the GPU run showed**

I ran the example on a 4x T4 box with real DeepSpeed ZeRO stage 2 (deepspeed 0.19.2, torch 2.11):

- All five step-time streams populate on every step under real DeepSpeed: dataloader, forward, backward, optimizer, and step time. The dataloader stream lands whenever a real torch `DataLoader` is iterated, which the example does.
- No optimizer double-count. DeepSpeed's ZeRO optimizer wrapper is not a `torch.optim.Optimizer`, so only the base AdamW step is timed: one optimizer event per step.
- Multi-GPU works through `--nproc-per-node 2`, with per-rank telemetry. Residual share rises from about 45% on 1 GPU to about 71% on 2 GPU. That jump is the ZeRO reduce-scatter and all-gather folding into residual, exactly as your Limitations section says.

**Follow-up commit (9abc6ca)**

- The loop ran `float(loss.detach())` inside the `trace_step` block every step, which forces a device-to-host sync in the traced region. On a single GPU, removing it dropped the example's own reported step time from 17.4ms to 9.1ms and its residual share from 45% to 26%, so the diagnosis was no longer dominated by the loss logging. I moved the loss handling out of `trace_step` and accumulate on-device, syncing only when it prints, the same pattern the quickstart example uses.
- The test installed forward/backward/h2d/optimizer but not the dataloader patch, and asserted three of the five streams. I added the dataloader patch plus a real-`DataLoader` `dataloader_next` assertion and a `step_time` assertion, so it now covers all five.
- Scoped the DataLoader line in the doc to "when you iterate a real torch `DataLoader`", and added a gradient-accumulation note to Limitations (the config ships accumulation=1; with more, optimizer time only lands on boundary steps).
- Pinned the `deepspeed` extra to `>=0.10.0` to match the other framework extras.

The integration and summary test suites are green and lint is clean. Happy to change any of this if you would rather take it a different way.

PR #198, opened 2026-07-13
RFC-0001: surface and correctly attribute native training crashes (+ RFC process)

Pendu:
This PR adds a lightweight RFC process (`rfcs/`) and the first proposal, RFC-0001. It is opened as a Draft: the PR is the discussion thread (per `rfcs/README.md`), not a request to merge code yet.

## RFC-0001: Surface and correctly attribute native training crashes

When a training process dies from a native fault (a segmentation fault, a host or container OOM-kill, or any signal death), TraceML today does not capture the crash. The run waits out its finalization timeout (up to ~5 minutes), then writes a summary that attributes the stall to a TraceML rank-timeout rather than to the crash. That points the user at TraceML instead of at the fault they need to debug, which is the gap this RFC closes.

The proposal:
- Install `faulthandler` in the training process so a native fault leaves a C-level traceback.
- Write a structured `crash.json` on signal death, with a cause enum (`signal_death` / `likely_oom` / `scheduler_terminated` / `unknown`) plus rank, pid, and host.
- Make finalization crash-aware: a closed socket is a trigger to consult the launcher's `Popen.poll()` oracle, not a death verdict, so a dead worker no longer hangs the run or gets blamed on TraceML.
- Attribute OOM from TraceML's own memory samples read against the cgroup limit, not host RAM, and never claim OOM without evidence.
- Optionally capture the child's stderr tail for a post-mortem.

The background section cites the current code (`main` @ e7813c1); the design records its rejected alternatives, a test plan, and open questions. Feedback on the shape is the point of this PR.

## Child issues

Contributors are welcome on the ones labeled `help wanted` and `good first issue`:
- #199 Surface and correctly attribute native crashes (the correctness-critical core).
- #200 Persist the child's stderr tail in summary mode (`--capture-stderr`), `help wanted`.
- #201 Tests and docs: "what happens when my training crashes", `good first issue`.

RFC-0002 (epoch-level tracker logging) will follow in a separate PR.

### DIFF
--- a/rfcs/0001-surface-and-attribute-native-crashes.md
+++ b/rfcs/0001-surface-and-attribute-native-crashes.md
@@ -16,17 +16,18 @@
 ## 1. Summary
 
 When a training process dies from a native fault (a segmentation fault, a host or
-container OOM-kill, or any signal death), TraceML records nothing useful about the
-failure and actively misleads.
-The run appears to hang for several minutes, then TraceML writes an artifact that
-frames the failure as a TraceML rank-timeout rather than as the crash.
+container OOM-kill, or any signal death), TraceML does not currently capture the
+crash.
+The run waits out its finalization timeout (up to ~5 minutes), then writes a summary
+that attributes the stall to a TraceML rank-timeout rather than to the crash, which
+points the user at TraceML instead of at the fault they need to debug.
 We reproduced this end-to-end (a forced native crash under `traceml run --mode
 summary`) on both CPU and a single-GPU CUDA run.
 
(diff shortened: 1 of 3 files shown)


abhinavsriva:
Closing this draft RFC as superseded by the merged failure-output hardening design in PRs #415, #416, #423, #431/#433, #439, #443, and #444, with final internal-log consolidation under review in #448. The implemented design preserves native Python/torchrun evidence and launcher-owned raw streams instead of adding faulthandler files, crash.json, or inferred crash/OOM taxonomy. Remaining actual-SIGSEGV validation is still tracked by issue #201.

PR #223, opened 2026-07-17
Add a CPU-first TraceML demo container

### LINKED ISSUE #216: Add a demo container so a new user can try TraceML in one command
## Summary

There is no container image for a first run today. A small image that runs an existing example end to end lets someone see a TraceML diagnosis with a single `docker run`, without setting up a Python environment first. This is useful for a quick evaluation on a single machine.

## Proposed approach

- Add a `Dockerfile` (slim Python base, `pip install traceml-ai[torch]`).
- Default `CMD` runs an example that already ships:
  ```
  traceml run examples/diagnosis/dataloader_bottleneck_demo.py --args --scenario slow
  ```
- Add a "Try it in Docker" section to the README with the `docker build` and `docker run` commands, plus a `--gpus all` note for running the same example on a GPU.

## Why this example

`examples/diagnosis/dataloader_bottleneck_demo.py` is a good fit for a container:

- No dataset download and no GPU required. It uses synthetic data (`SyntheticInputDataset`) and guards CUDA use (`pin_memory=device.type == "cuda"`), so it runs on a CPU-only host.
- The `--scenario slow` run produces a clear input-bound result, so the container shows a bottleneck diagnosis, not just a successful install.

Most people evaluate on a single machine, so a CPU-runnable, no-download example keeps the barrier at zero while still showing the tool's headline input-bound diagnosis. `examples/quickstart.py` (a tiny CPU model) is an alternative if we prefer the default run to be a plain smoke check.

## Out of scope

`docker compose` and any multi-service stack. A single training container is enough for this demo.

## Acceptance

- `docker build` produces the image.
- `docker run <image>` on a CPU-only host prints TraceML's end-of-run summary with the input-bound diagnosis, using no extra flags.

## Target branch

Please base this on `version_0.3.5` so the later merge into `main` stays clean.

sahilmathur254:
## What changed

- Add a root Dockerfile based on Python 3.10 slim.
- Install TraceML from the checked-out source with `.[torch]`.
- Run the slow DataLoader bottleneck demo in summary mode by default.
- Exclude repository metadata and local or generated artifacts from the image.
- Document the CPU-first build and run commands, with GPU access presented only as optional guidance for configured NVIDIA environments.

## Why

TraceML did not have a container-based first-run path. This gives new users a single-image demo that requires no host Python environment, external dataset, or GPU and prints the input-bound diagnosis requested in #216.

## How I tested

- [ ] Tests added or updated
- [ ] Existing tests run
- [x] Manual smoke test
- [ ] Not run; reason:

Validation performed in an isolated Docker context:

- `docker build -t traceml-demo .`
- `docker run --rm traceml-demo`
  - exited with status 0
  - ran on `Device: cpu`
  - completed 60 training steps
  - printed `TraceML Verdict: INPUT-BOUND / CRITICAL`
- Verified the image excludes Git metadata, OS files, Python caches, virtual environments, generated logs, and pip build artifacts.
- `git diff --check`

No unit tests were added or run because the change is limited to the container entry point, build context, and documentation; the issue's end-to-end Docker acceptance path was exercised directly.

## Runtime impact

- [x] No training-path impact
- [ ] Not sure

Notes: The change only adds an opt-in container entry point and documentation. The locally resolved image is about 3 GB because the current Linux PyTorch dependency set is substantial.

## Docs

- [x] Updated
- [ ] Not needed

## Extra context

The default command follows the maintainer-confirmed CPU-first scope:

```text
traceml run examples/diagnosis/dataloader_bottleneck_demo.py --mode=summary --args --scenario slow
```

The optional `--gpus all` example does not guarantee CUDA support; it is conditional on NVIDIA Container Toolkit and a CUDA-capable PyTorch installation. GPU execution was not tested and is not an acceptance requirement.

Closes #216

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.git
+.DS_Store
+Thumbs.db
+.pytest_cache
+.mypy_cache
+.ruff_cache
+.tox
+.venv
+__pycache__
+*.py[cod]
+*.egg-info
+build
+dist
+logs
+site
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3.10-slim
+
+WORKDIR /app
+
+COPY . .
+
+RUN python -m pip install --no-cache-dir --timeout 120 ".[torch]" \
+    && rm -rf build src/traceml_ai.egg-info
+
+CMD ["traceml", "run", "examples/diagnosis/dataloader_bottleneck_demo.py", "--mode=summary", "--args", "--scenario", "slow"]
--- a/README.md
+++ b/README.md
@@ -46,26 +46,6 @@ TraceML produces actionable diagnostics with under 1% overhead in current benchm
 
 ## Quickstart
 
-### Try it in Docker
-
-Build the image from the repository checkout, then run the default CPU demo:
-
-```bash
-docker build -t traceml-demo .
-docker run --rm traceml-demo
-```
-
-The image installs TraceML from the checkout with PyTorch support and runs the
-slow DataLoader diagnosis demo in summary mode. The final report should identify
-the run as input-bound; no external dataset or GPU is required.
-
-If the host has the NVIDIA Container Toolkit configured and the installed
-PyTorch build supports CUDA, the same demo can optionally access the GPU:
-
-```bash
-docker run --rm --gpus all traceml-demo
-```
-
 ### 1. Install TraceML
 
 For the live browser dashboard:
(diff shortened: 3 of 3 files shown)


Pendu:
Thanks for this, it's a nice addition. Reviewed and tested end to end:

docker build -t traceml-demo . builds cleanly, image size 3.04GB.
docker run --rm traceml-demo exits 0 and prints TraceML Verdict: INPUT-BOUND / CRITICAL, running fully on CPU.
Verified the image excludes .git, caches, and build artifacts, and that TraceML installs into site-packages rather than an editable checkout.
The --mode=summary --args --scenario slow ordering in the CMD matches the CLI's own documented usage.

I pushed one small follow-up (5eff3ce): moved the new "Try it in Docker" section to after the numbered install walkthrough, so it lines up with how the README already places Or launch your script directly, the other alternative path, after the numbered steps rather than before them. No content changes, just the reorder.

PR #221, opened 2026-07-16
Abhinav/add runtime metadata

abhinavsriva:
## How I tested

- [ ] Tests added or updated
- [ ] Existing tests run
- [ ] Manual smoke test
- [ ] Not run; reason:

## Runtime impact

- [ ] No training-path impact
- [ ] May affect tracing, runtime, telemetry, or distributed behavior
- [ ] Not sure

Notes:

## Docs

- [ ] Updated
- [ ] Not needed

PR #228, opened 2026-07-18
Straggler simplified

abhinavsriva:
## What changed

This simplifies Step Time rank-straggler attribution for distributed training.

- Replaces the old clean-step/worst-vs-median attribution with culprit/victim attribution based on visible rank skew.
- Uses backward time as the visible skew anchor for DDP/default strategy, and  forward + backward for FSDP.
- Identifies the culprit as the rank that waited least in the visible phase and  the victim/reference rank as the upper actual median rank by visible phase.
- Attributes material culprit excess to input wait, H2D, or DDP forward time.
- Falls back to generic `STRAGGLER` when visible skew is sync-bound or  unattributed.
- Removes `RESIDUAL_STRAGGLER` from the rank-skew diagnosis set; residual  behavior remains covered by `RESIDUAL_HEAVY`.
- Abstains from rank-straggler diagnosis when visible-phase anchors or step  envelopes are missing, instead of treating missing instrumentation as a fast  rank.
- Updates final-summary primary diagnosis, Step Time report text, schema docs,  user docs, and developer docs to match the new attribution language.

## Why

The previous clean-step policy was hard to explain and could blur true culpritwork with synchronization wait observed on peer ranks. The new policy makes therank-skew story more direct:

- first detect visible wait cost,
- then name the likely culprit and victim/reference rank,
- then attribute only material culprit-side excess,
- otherwise leave the result as sync-bound or unattributed.

This also makes the FSDP behavior more honest. FSDP forward and backward caninclude sharding communication, so TraceML avoids emitting `COMPUTE STRAGGLER`from this policy without explicit collective timing.

## Markdown/docs delta vs `version_0.3.5`

Tracked Markdown files changed in the committed branch diff:

- `docs/developer_guide/architecture.md`
- `docs/guides/ddp-slow-training-rank-straggler.md`
- `docs/guides/low-gpu-utilization-pytorch.md`
- `docs/guides/slow-pytorch-training.md`
- `docs/user_guide/faq.md`
- `docs/user_guide/reading-output.md`
- `src/traceml_ai/diagnostics/DIAGNOSIS.md`
- `src/traceml_ai/reporting/SCHEMA.md`

The docs now consistently use culprit/victim/reference language, remove`RESIDUAL_STRAGGLER`, describe DDP vs FSDP attribution separately, and explain generic `STRAGGLER` as sync-bound or unattributed rank skew.

`mkdocs.yml` also adds a Developer Guide nav entry for`docs/developer_guide/rank-straggler-policy.md`. That file exists in the local worktree but is currently untracked, so include it when preparing the PR orremove the nav entry before publishing.

## How I tested

Focused tests for the changed Step Time and primary-diagnosis behavior:

```bash
  tests/diagnostics/test_step_time.py \
  tests/reporting/summary/test_primary_diagnosis.py \
  -q
```

Result:

```text
34 passed in 0.23s
```

- [x] Tests added or updated
- [x] Existing focused tests run
- [ ] Manual smoke test
- [ ] Not run; reason:

## Runtime impact

- [x] No training-path instrumentation impact
- [x] May affect live output, final summaries, reports, and diagnosis labels
- [ ] May affect telemetry collection or storage
- [ ] Not sure

Notes:

This changes diagnosis policy and presentation, not sampler timing collection. Existing telemetry can classify differently because rank-skew attribution no longer uses clean-step residual attribution and no longer emits `RESIDUAL_STRAGGLER`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 
 
 [![PyPI version](https://img.shields.io/pypi/v/traceml-ai.svg)](https://pypi.org/project/traceml-ai/)
+[![PyPI Downloads](https://static.pepy.tech/personalized-badge/traceml-ai?period=total&units=INTERNATIONAL_SYSTEM&left_color=grey&right_color=blue&left_text=downloads)](https://pepy.tech/projects/traceml-ai)
 [![CI](https://github.com/traceopt-ai/traceml/actions/workflows/ci.yml/badge.svg)](https://github.com/traceopt-ai/traceml/actions/workflows/ci.yml)
 [![Python 3.10+](https://img.shields.io/badge/python-3.10%2B-blue)](https://www.python.org/)
 [![License](https://img.shields.io/badge/license-Apache%202.0-blue)](./LICENSE)
--- a/docs/developer_guide/architecture.md
+++ b/docs/developer_guide/architecture.md
@@ -104,7 +104,7 @@ Architectural risks and known structural debt. Day-to-day bugs live in the issue
 | Step / trace_step | A training iteration; `with trace_step(model)` marks the boundary that phase timing is computed against. |
 | Residual (residual_proxy) | Residual step time, `max(0, step - h2d - forward - backward - optimizer)`. |
 | INPUT_BOUND / COMPUTE_BOUND | Input wait dominates iteration time, or compute dominates the traced step. |
-| INPUT_STRAGGLER / COMPUTE_STRAGGLER / H2D_STRAGGLER / RESIDUAL_STRAGGLER / STRAGGLER | One rank is slower than typical after accounting for backward waiting; the label names the dominant timing difference, or `STRAGGLER` when mixed. |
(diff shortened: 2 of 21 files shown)


Pendu:
Pushed a fix for the one blocker I found here (`7b68aa7`): the nav entry `developer_guide/rank-straggler-policy.md` had no matching page, so `mkdocs build --strict` aborted. Added that page, drawn from the Step Time policy in `src/traceml_ai/diagnostics/DIAGNOSIS.md` so it stays consistent with the code.

Verified on the pushed head:
- `mkdocs build --strict` builds clean (also confirmed by the pre-commit hook on push).
- `pytest tests/diagnostics tests/reporting` -> 253 passed.

Two notes, both your call:
- The branch is still behind `version_0.3.5`. Merging it up to date (as you did for #221) works cleanly now that #221 landed with its commits preserved.
- The "test corrected" commit dropped the only direct unit tests for `ResidualHeavyRule` / `ComputeBoundRule` (`test_step_time_rules_trigger_and_no_trigger_cases`). Both rules still get indirect coverage through `build_step_diagnosis_result`, so it is a coverage reduction rather than a regression. Optional to re-add.

PR #230, opened 2026-07-20
benchmarking: land overhead study, attribution harness, and campaign overview

Pendu:
## What's in this branch

Two independent overhead-benchmarking tracks, plus an index tying them together:

- **Wall-clock overhead study** (`benchmarking/`) — paired `time torchrun` vs
  `time traceml run` runs, no TraceML-specific tooling required to reproduce.
  Two dated campaigns so far (2026-06-11 on v0.3.1, 2026-07-19 on v0.3.5),
  both within the <2% overhead budget.
- **Attribution harness** (`perf_benchmark/`) — automated, JSON-config-driven
  benchmark with a static cost-model audit of every instrumentation hook
  (dataloader/forward/backward/H2D/optimizer/sampler thread), tagging
  critical-path vs off-thread cost. Methodology and tooling only; no
  campaign run yet.
- **`BENCHMARKING.md`** — a short overview at the repo root explaining how
  the two tracks relate and pointing to both.

## Commits

- `3c87329` — the original overhead study (extracted from `multnode_display`)
- `fc349e1` — the attribution harness (unedited, authored by @abhinavsriva)
- `0677855` — the overview doc
- `f0811dd` — the second wall-clock campaign (v0.3.5)

## Testing

No changes to `src/`. The new campaign's numbers were independently
re-derived from the raw run logs and cross-checked against the per-run
telemetry cards. `git merge-tree` against `main` is conflict-free.


Pendu:
Landed a 1/2/4-node scaling campaign on this branch: `benchmarking/analysis/2026-07-21_ondemand_1_2_4node_g4dn/`.

## Headline (tiny_mlp, `trace_auto` vs `never_init`)

| Topology | Baseline median | `trace_auto` overhead | `trace_manual` overhead |
|---|---:|---|---|
| 1 node | 246.39ms | +29.56% (real signal) | +10.46% (within noise) |
| 2 node | 247.98ms | +27.96% (real signal) | +6.66% (within noise) |
| 4 node | 216.16ms | +34.43% (real signal) | +12.72% (real signal) |

Not a clean monotonic scaling trend from these 3 single-campaign data points (2-node is actually lower than 1-node) — flagged in the report as needing repeat campaigns per topology before calling it a trend rather than noise. Full caveats (phase-vs-step timing-mode mismatch between the 1-node and 2/4-node runs, the unexplained faster 4-node baseline) are in the report.

## Worth a look: phase-attribution percentage artifact

`aggregate_results.py`'s phase-attribution rows divide by the `never_init` baseline for that phase. When that baseline is near-zero (e.g. `trace_context_enter_ms` at 0.0114ms), a legitimate ~10ms of real instrumentation cost becomes a reported **90,000%+** overhead. The absolute ms is real and worth reporting; the percentage isn't meaningful at that scale. This was flagged as a methodology concern before any GPU data existed, and it reproduces on real hardware now — planning to open an issue for it, let me know if you'd rather handle it differently.

## Raw data

Full per-rank JSON traces (multi-GB, not in git) are in S3 if you want to look yourself:
```
s3://traceopt-perf-benchmark-results-543486084785/aws_full_t4_phaseonly_20260721_074055/
s3://traceopt-perf-benchmark-results-543486084785/aws_2node-ondemand_t4_20260721_171556/
s3://traceopt-perf-benchmark-results-543486084785/aws_4node-ondemand_t4_20260721_173544/
```

abhinavsriva:
@Pendu : 

I reviewed against clean reruns on the same T4, and I think the new 1/2/4-node campaign cannot be used for normal-overhead as currently presented.

`run_benchmark.py` enables --gil-probe by default. That starts `GilVictim`, an infinite Python loop with no sleep. CPython threads share one GIL, so this is not a passive mechanism: it actively competes with the training thread for the GIL and CPU in every cell, including `never_init`.

**The effect is material.** On the same T4, clean step-mode never_init is ~1.38 ms/step with the probe off, enabling the probe produced ~221 ms/step before TraceML was enabled. This means the reported 246/248/216 ms baselines are results from a forced GIL-contention stress workload, not normal TraceML overhead.

The aggregator interval is intended to bound periodic collection work; it is not equivalent to a continuously CPU-bound Python thread. A GIL-stress test may still be useful, but it needs to be an explicit, separately labelled diagnostic, not the default overhead benchmark.

Recommendation:
- Validate the clean single-node / single-GPU matrix first.
- Run the clean 2-node matrix only after the single-node result is stable.
- Run the clean 4-node matrix only after the 2-node result is stable.
- At each topology, test workload scale separately, for example batch sizes 256, 512, and 1024, before making a scaling claim.

The clean single-GPU campaign is reproducible with:
`bash benchmarking/perf_benchmark/phase3_benchmark/run_clean_single_gpu_campaign.sh`

It uses matched launcher-disabled/manual/auto cells, step timing, GIL probe off, fresh repeats, rotated cell order, and per-repeat reporting.

The aggregation should treat process repeats, not pooled individual step samples, as the independent experimental units. Pooling step samples remains useful for within-run p95/p99, but it must not be presented as 10,000 independent benchmark trials. 

The cells should also be interleaved or randomized by repeat; running all baseline repeats, then all manual repeats, then all auto repeats (3-3-3) introduces time-of-run/system-drift bias.

Pending clean reruns, retain the current 2026-07-21 artifacts only as GIL-stress diagnostic data and supersede their normal-overhead/scaling headline.

abhinavsriva:
Clean 1-node / 1-T4 results, step timing, GIL probe off. Each cell has 10 fresh process repeats; values are medians of the per-repeat medians.

 | Batch size | Disabled baseline (ms) | TraceML manual (ms) | Manual overhead | TraceML auto (ms) | Auto overhead |
 |---:|---:|---:|---:|---:|---:|
 | 256 | 1.3846 | 1.6146 | +0.2300 ms / 16.6% | 1.9365 | +0.5519 ms / 39.9% |
 | 512 | 1.4968 | 1.6805 | +0.1837 ms / 12.3% | 2.0376 | +0.5408 ms / 36.1% |
 | 1024 | 2.3480 | 2.4311 | +0.0831 ms / 3.5% | 2.5393 | +0.1913 ms / 8.2% |

 The `tiny_mlp` batch-256 case is intentionally a **fixed-overhead stress case**. As compute increases, the percentage and, at batch 1024, the observed absolute wall-time cost falls substantially.

Pendu:
@abhinavsriva thanks for catching this — verified it independently and confirmed the same root cause: `run_benchmark.py:220` does `config.get("gil_probe", True)`, and none of the three configs used for this campaign (`single_gpu.json`, `multinode_2x1.json`, `multinode_4x1.json`) override it to `false`.

Worth noting for the record: there's no CLI flag for `gil_probe` at all, and nothing in the README flags it as a risk. So there was no signal anywhere — not in the docs, not in the configs, not in the command line — that would have told us to turn it off before running. Agreed this data should stand only as GIL-stress diagnostic, not normal-overhead evidence, until a clean rerun replaces it.

Pendu:
Landed a clean rerun of the 1/2/4-node campaign: `benchmarking/analysis/2026-07-22_clean_1_2_4node_g4dn/`. Marked the 2026-07-21 campaign as superseded (GIL-stress diagnostic only, not normal-overhead evidence — see #233/#240).

`gil_probe: false` explicit in every config this time, 10 repeats/cell (up from 3), and a batch-size sweep (256/512/1024) at every topology.

## Headline — leading with absolute ms, since percentage alone is misleading here

| Batch size | Topology | Baseline (ms) | `trace_auto` overhead | `trace_manual` overhead |
|---|---|---:|---|---|
| 256 | 1-node | 1.3775 | **+0.578 ms (42.0%)** — real signal | +0.244 ms (17.7%) — real signal |
| 256 | 2-node | 5.0724 | +0.396 ms (7.8%) — within noise | +0.115 ms (2.3%) — within noise |
| 256 | 4-node | 9.2348 | +0.378 ms (4.1%) — within noise | +0.080 ms (0.9%) — within noise |
| 512 | 1-node | 1.5060 | **+0.520 ms (34.5%)** — real signal | +0.222 ms (14.8%) — real signal |
| 512 | 2-node | 5.0185 | +0.447 ms (8.9%) — within noise | +0.195 ms (3.9%) — within noise |
| 512 | 4-node | 8.8797 | +0.431 ms (4.9%) — within noise | +0.277 ms (3.1%) — within noise |
| 1024 | 1-node | 2.4439 | +0.172 ms (7.0%) — real signal | +0.049 ms (2.0%) — within noise |
| 1024 | 2-node | 5.3442 | +0.390 ms (7.3%) — within noise | +0.128 ms (2.4%) — within noise |
| 1024 | 4-node | 9.2670 | +0.396 ms (4.3%) — within noise | +0.183 ms (2.0%) — within noise |

A pure-percentage table is misleading here: the absolute overhead stays in a narrow ~0.17-0.58ms band across all 9 conditions while the percentage swings 4-42% purely because the baseline step time grows (1.4ms → 9.3ms) with node count. Reads as "overhead shrinks dramatically" when it's really "roughly fixed cost, shrinking fraction of a growing baseline."

## Possible mechanism (hypothesis, not confirmed)

`trace_auto`'s cost looks like a fixed CPU-side bookkeeping cost per step. At 1-node it shrinks monotonically with batch size (256→512→1024: 0.578→0.520→0.172ms) — consistent with CUDA async overlap hiding more of that fixed cost as GPU compute-per-step grows. Matches your own observation from PR #230 about bs=256 being an intentional fixed-overhead stress case and absolute cost dropping at bs=1024. 2-node/4-node don't show as clean a trend, likely because baseline there is dominated by cross-node sync wait, not local compute, leaving less overlap headroom.

Important: "within noise" at 2/4-node isn't evidence overhead is absent at scale — the noise floor grew (1.4-4.6ms from cross-machine jitter) enough to swallow the same ~0.4ms signal, not because the cost went to zero. Would need a heavier workload than `tiny_mlp` to actually test this (overhead should keep shrinking with more compute/step if the hypothesis holds; staying flat would suggest something synchronous instead — not run here).

Same standing caveat as before: cell order is still block-ordered (no per-repeat rotation) for multi-node.

abhinavsriva:
@Pendu  Cool ! So it seems we are at most 1 ms per rank per step.

I suggest to create two branches:

- benchmarking: benchmarking code, methodology, results table, and benchmarking/README.md.
- benchmarking-data: original raw data, organized by campaign and link every table result to its exact data commit/tag.

We can remove the GIL test, it is not needed for this benchmark (?). 

Keep raw data in Git only if it is small; a separate branch still adds to repository size (benchmarking-data only if small). 

In README, I also suggest to add something like: 

`Measured overhead: In our current benchmark configuration, TraceML adds less than 1 ms per rank per training step. See the (benchmark methodology and results)[link] for the tested configuration (and raw data if kept).`

Pendu:
Agreed on the headline — **< 1 ms per rank per step** holds across everything tested (max observed was 0.578 ms at 1-node bs=256, everything else lower). All landed on this branch:

- **README line added** ([`README.md`](https://github.com/traceopt-ai/traceml/blob/benchmarking/README.md)): *"Measured overhead: in our current benchmark configuration, TraceML adds less than 1 ms per rank per training step"*, linking to [`benchmarking/README.md`](https://github.com/traceopt-ai/traceml/blob/benchmarking/benchmarking/README.md) — now the overview of both tracks with headline results and links to each campaign. Replaces the old "under 1% overhead" sentence; the ms framing is more honest than a percentage given how much the percentage swings with baseline step time.
- **Docs standardized while at it:** each folder now renders its own README (`benchmarking/` = overview, `workloads/` = wall-clock harness, `perf_benchmark/` = attribution harness, `analysis/` = campaign index with superseded status flagged). `BENCHMARKING.md` folded into `benchmarking/README.md`.
- **GIL probe: kept but made strictly opt-in** rather than removed. `run_benchmark.py` now defaults `gil_probe` to `false` (was `true` — the root cause of the contaminated campaign), and `perf_benchmark/README.md` documents it explicitly as a stress diagnostic that must never be on for a normal-overhead campaign. Rationale for keeping the code: it's the only tool we have for the GIL-contention investigation if that comes back, and as a clearly-labelled opt-in it can't contaminate anything again. If you'd still rather delete it outright, no strong objection — say the word.

On the branch split / raw data: **the raw data was never in git** — each campaign's multi-GB per-rank traces live in S3 (`s3://traceopt-perf-benchmark-results-543486084785/<run-id>/`, accessible with AWS credentials for this account; run-ids are in each report). Only the distilled `summary.csv` + `report.md` per campaign are committed under `benchmarking/analysis/`, a few hundred KB total — so repo size isn't currently a concern, and a `benchmarking-data` branch wouldn't change what a clone downloads anyway (all branches share the same object store). If the goal is public traceability for outside readers rather than repo hygiene, that's a real gap worth discussing — S3 needs credentials, so external users can't verify raw data today. Options if we want that later: git LFS, a public bucket path, or sharing data on request. Worth a quick sync on what we actually want there.

PR #250, opened 2026-07-23
Abhinav/issue236

### LINKED ISSUE #236: [Step Time] Normalize bottleneck shares to selected iteration time
Compute per-rank component shares and then their median. Use selected-clock `input_wait + step_time` as the denominator for input, residual, and later H2D. 

Remove metric-relative skew as a gate that suppresses a typical bottleneck. Keep skew as evidence only. Make `COMPUTE_BOUND` informational when no overhead component reaches 10%.

Minimal code surface:

- `diagnostics/step_time/policy.py`
- `diagnostics/step_time/context.py`
- `diagnostics/step_time/rules.py`
- `tests/diagnostics/test_step_time.py`
- canonical Step Time policy documentation

Acceptance:

- Median is taken over per-rank shares, not from a ratio of unrelated rollups.
- Input and residual use iteration share with 10%/20% severity.
- A typical bottleneck is not hidden because one rank is worse.
- Existing selected-clock, warmup, and FSDP gates still pass.

abhinavsriva:
## Summary

Normalize Step Time typical-bottleneck diagnosis around rank-aligned, selected-clock iteration shares.

This fixes a correctness issue where typical input and residual shares could be derived from separate metric rollups rather than from each rank's actual iteration cost. It also makes compute-bound output informational and removesobsolete threshold knobs that no longer affect policy.

## What Changed

- Compute input and residual shares per rank using:

  ```text
  iteration_r = input_wait_r + step_time_r
  component_share_r = component_r / iteration_r
  typical_component_share = median(component_share_r across ranks)
  ```

- Use shared `overhead_share_warn` / `overhead_share_crit` policy fields at   10% / 20% for both `INPUT_BOUND` and `RESIDUAL_HEAVY`.
- Keep rank skew as evidence in findings, but stop using it to suppress a  typical bottleneck that affects the median rank.
- Keep the existing compute-dominance gate, but make `COMPUTE_BOUND`  informational when no material input or residual overhead is present.  It can appear as secondary context beside an actionable rank straggler.
- Remove obsolete policy fields and summary overrides that no longer affected  diagnosis behavior.
- Update canonical and user-facing Step Time documentation.

## Why

In distributed training, independently taking a median input time and a median step time can combine values from different ranks. That ratio does not represent a real rank's iteration share and can cross a diagnosis thresholdincorrectly.

The new policy measures each rank's observed overhead fraction first, then takes the median. This preserves the selected-clock model and makes the diagnosis explainable: a typical bottleneck is not hidden just because one rank has more skew.

`COMPUTE_BOUND` now describes productive work rather than an actionable warning. Input, residual, and rank-straggler findings remain the actionable signals.

## Scope

- No instrumentation, telemetry, SQLite, transport, dashboard, or schema   changes.
- No `H2D_BOUND` diagnosis in this PR; the rank-share helper is intentionally reusable for that follow-up.
- No YAML configuration is added yet. The shared `overhead_share_*` fields are the intended future configuration surface.

## Verification

```bash
  tests/diagnostics \
  tests/renderers/test_step_time_compute.py \
  tests/renderers/test_step_time_renderer.py \
  tests/reporting/summary/test_step_time.py \
  tests/reporting/summary/test_step_time_card.py \
  -q
```

Focused coverage includes:

- median-of-per-rank-shares versus ratio-of-rollups;
- input and residual 10% warning / 20% critical boundaries;
- skew no longer suppressing a typical bottleneck;
- informational compute-bound behavior and its coexistence with a straggler;
- selected-clock, warmup, FSDP, renderer, and final-summary paths.

### DIFF
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -162,6 +162,8 @@ What to do next:
 Meaning:
 
 - input wait is taking a large share of iteration time
+- TraceML uses the median per-rank input share, so one unusually slow rank
+  does not hide a broad input bottleneck
 
 Common causes:
 
@@ -190,6 +192,7 @@ What to do next:
 Meaning:
 
 - model compute dominates the typical step
+- this is informational when no material input or residual overhead is visible
 
 In practice this means most step time is going into:
 
@@ -365,6 +368,11 @@ In TraceML:
 - `compute = forward + backward + optimizer`
 - `residual = step_time - h2d - compute`
 - `total_step = input_wait + step_time`
+
+TraceML evaluates residual as the median per-rank share of selected-clock
+iteration time (`input_wait + step_time`). Input and residual findings warn at
+10% and are critical at 20%; rank skew is supporting evidence rather than a
+gate that hides a typical bottleneck.
 
 This is residual unattributed time inside the traced step, not direct
 collective, NCCL, or all-reduce timing.
--- a/src/traceml_ai/diagnostics/DIAGNOSIS.md
+++ b/src/traceml_ai/diagnostics/DIAGNOSIS.md
@@ -73,9 +73,11 @@ Training-step timing.
   materially higher forward time than the victim rank.
 - `H2D_STRAGGLER`: the culprit rank has materially higher host-to-device
   transfer time than the victim rank.
-- `INPUT_BOUND`: selected-clock input wait dominates iteration time.
-- `COMPUTE_BOUND`: forward/backward/optimizer time dominates the typical step.
-- `RESIDUAL_HEAVY`: unattributed residual time is a material share of the step.
+- `INPUT_BOUND`: selected-clock input wait is a material typical iteration cost.
+- `COMPUTE_BOUND`: forward/backward/optimizer time dominates the typical step;
+  this is informational when no material overhead is visible.
+- `RESIDUAL_HEAVY`: unattributed residual time is a material typical iteration
+  cost.
 
 Step-time diagnosis uses one selected clock for the analyzed window. It uses
 GPU event timing when every rank/step has GPU timing for the step envelope,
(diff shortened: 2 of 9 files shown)


Pendu:
Reviewed and tested end to end. This reads clean and the correctness fix is the right one: median of per-rank iteration shares instead of a ratio of separate rollups, so the typical-bottleneck verdict reflects a real rank rather than a blend across ranks.

Verified:
- The new helper is keyed correctly: `local_per_rank_timing` is `{rank: {component: value}}` (`context.py:610`), and `_median_iteration_component_share` iterates per-rank dicts, so the median is over per-rank shares as intended.
- No dangling references to the removed policy fields (`input_share_*`, `residual_share_*`, `*_max_skew`, `compute_bound_share_crit`): zero hits across `src/`, `tests/`, `docs/`.
- `severity="info"` for COMPUTE_BOUND is an already-handled value, and `formatters.py` now derives the style from severity rather than a hardcoded kind, so the CLI color follows the engine.
- Checked the empty-`per_rank_timing` edge: the helper returns 0.0 with no per-rank data, but the window loader always populates rank 0 for single-rank runs (the selection tests cover this), and the only empty path is the stale/no-data fallback where no diagnosis is expected. No regression there.
- `pytest tests/diagnostics tests/renderers tests/reporting` -> 263 passed. The added tests check real boundaries, for example the residual 10/20 severity and STRAGGLER staying primary while COMPUTE_BOUND rides along as info.
- Lint clean at the pinned pre-commit versions on the source files.

One thing to confirm rather than change: RESIDUAL_HEAVY shifts on two axes at once, the denominator moves from `step_time` to `input_wait + step_time` and the threshold moves from 15/25 to 10/20. The net is that residual warns a bit more readily in low-input runs and less in high-input ones. That matches #236 and the DIAGNOSIS.md update, so this is only worth a sanity check that the new sensitivity is what you want.

One optional nit: `tests/reporting/summary/test_step_time_card.py` has isort-nonconformant imports. The isort hook is scoped to `src/|dev/|docs/`, so CI will not flag it; only mentioning it in case you run isort locally.

Looks good to merge.

PR #253, opened 2026-07-24
Abhinav/issue238

### LINKED ISSUE #238: [Step Time] Separate straggler impact from cause coverage
Depends on https://github.com/traceopt-ai/traceml/issues/236

Keep the current culprit/victim and DDP/FSDP visible-phase model. Use visible cost divided by victim iteration for severity. Name a component cause only when that component explains at least 80% of visible cost.

Minimal code surface:

- `diagnostics/step_time/context.py`
- `diagnostics/step_time/rules.py`
- `docs/developer_guide/rank-straggler-policy.md`
- focused straggler unit tests

Acceptance:

- Impact below 10% emits no straggler diagnosis.
- Impact at 10%/20% produces warning/critical subject to existing confidence
  gates.
- At least 80% input/H2D/compute coverage is required for a named straggler.
- Lower coverage produces generic `STRAGGLER` with coverage evidence.
- Severity is unchanged when only the attribution label changes.

abhinavsriva:
## Summary

build on top of issue 236

Build on `abhinav/issue237` by making rank-straggler attribution evidence-based. The existing culprit/victim selection and normalized straggler impact remain unchanged. A straggler is now given an input, H2D, or DDP-compute cause only when that measured component explains at least 80% of the visible wait cost.

When no measured component meets that bar, TraceML emits a generic `STRAGGLER` and retains the component evidence for investigation.

## What Changed

- Add `straggler_cause_coverage_min = 0.80` to `DiagnosisThresholds` as the  future configuration surface.
- Preserve current rank selection:
  - DDP uses `backward_ms` as the visible synchronization phase.
  - FSDP uses `forward_ms + backward_ms`.
  - The culprit is the rank with the smallest visible phase; the victim is the
    upper-median real rank by that phase.
- Preserve current impact and severity:

  ```text
  impact = (victim_visible_ms - culprit_visible_ms)
           / (victim_input_wait_ms + victim_step_time_ms)
  ```

  Impact below 10% emits no straggler finding; 10% warns and 20% is critical, subject to the existing confidence gates.

- Separate attribution from severity. For each measured component:

  ```text
  component_coverage = min(1, component_excess_ms / visible_cost_ms)
  ```

  The largest coverage wins, with deterministic `input`, `h2d`, then `compute` tie-breaking. A named straggler requires at least 80% coverage  otherwise the finding remains generic `STRAGGLER`.
- Keep FSDP conservative: it may attribute input or H2D, but never labels a  compute straggler without explicit collective timing.
- Include `component_coverage` alongside existing raw `component_excesses_ms` in public issue evidence and clarify summaries.
- Update the canonical diagnosis and developer rank-straggler policy docs.

## Why

Impact answers whether synchronization skew materially affects iteration time. Attribution answers whether TraceML has enough measured evidence to name its cause. Those are different decisions: a severe wait can be real while its measured input, H2D, and compute excesses do not justify a specific label.

This keeps severity stable while preventing overconfident root-cause labels.

## Scope

- No telemetry, transport, SQLite, dashboard, reporting schema, or YAML  configuration changes.
- No change to H2D-bound behavior or to existing `H2D_STRAGGLER` semantics.
- No change to culprit/victim selection, visible-phase model, impact formula,
  thresholds, or existing early-window/FSDP severity caps.

## Verification

```bash
  tests/diagnostics/test_step_time.py -q
```

Focused coverage includes:

- 9%, 10%, and 20% impact boundaries;
- input, H2D, and DDP-compute attribution at 79% and 80% coverage;
- generic stragglers with raw excess and normalized coverage evidence;
- coverage bounded at 100%;
- unchanged score and severity when only the attribution label changes;
- FSDP input/H2D attribution and its prohibition on compute attribution;
- existing culprit/victim, missing-anchor, warmup, and FSDP severity-cap
  behavior.

`git diff --check` passes.

## Base

Comparison: `abhinav/issue237...abhinav/issue238`

### DIFF
--- a/docs/developer_guide/rank-straggler-policy.md
+++ b/docs/developer_guide/rank-straggler-policy.md
@@ -20,7 +20,7 @@ window, as opposed to the single-rank bottleneck kinds (`INPUT_BOUND`,
 
 | Kind | Meaning |
 |---|---|
-| `STRAGGLER` | Visible rank skew exists, but input wait, H2D, and forward do not explain the likely culprit. |
+| `STRAGGLER` | Visible rank skew exists, but no measured component explains enough of the visible wait to name a cause. |
 | `INPUT_STRAGGLER` | The culprit rank has materially higher selected-clock input wait than the victim. |
 | `COMPUTE_STRAGGLER` | DDP / default strategy only: the culprit rank has materially higher forward time than the victim. |
 | `H2D_STRAGGLER` | The culprit rank has materially higher host-to-device transfer time than the victim. |
@@ -72,8 +72,10 @@ denom = input_wait_victim + step_time_victim
 score = (visible_victim - visible_culprit) / denom
 ```
 
-If `score < 0.10`, TraceML does not report a rank straggler. The gap is treated
-as normal run-to-run variation rather than a real straggler.
+If `score < 0.10`, TraceML does not report a rank straggler. At 10% it emits a
+warning; at 20% it emits critical severity when the existing confidence gates
+permit it. The gap below 10% is treated as normal run-to-run variation rather
+than a real straggler.
 
 ## Attribution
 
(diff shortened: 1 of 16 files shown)


Pendu:
Reviewed and tested. This completes the Step Time normalization set (input, residual, H2D, and now straggler), and it matches #238 closely: visible cost over victim iteration drives severity, and a component only names the cause when it covers at least 80% of visible wait, otherwise the diagnosis stays a generic STRAGGLER with coverage evidence.

Verified:
- The classification change is real, not just wording: subtype attribution moved from `component_excess / iteration >= 0.10` to `component_coverage = min(1, excess / visible_cost) >= 0.80`. So named stragglers are more conservative, and severity is driven only by the score. Coverage chooses the label, score chooses severity.
- No divide-by-zero on the new `excess / visible_cost`: if visible cost were zero the score would be zero and fall below the 0.10 gate, so the function returns before reaching that line. Victim iteration is also non-zero because eligible ranks require `step_time > 0`.
- Test coverage maps directly onto the acceptance criteria: impact below 10% emits nothing, 10% and 20% hit warning and critical, at least 80% coverage is required to name a component, coverage changing the label does not change severity, and coverage is bounded to 1.0.
- `pytest tests/diagnostics tests/reporting tests/renderers` -> 290 passed.
- Lint clean at the pinned pre-commit versions, and both DIAGNOSIS.md and the rank-straggler-policy dev guide document the coverage formula, the 80% threshold, and the label-versus-severity split.

One small note, not blocking: `component_coverage` can be negative for a non-dominant component where the culprit was actually faster on that phase, since `min(1, ...)` bounds only the top. It is harmless here because only the qualifying component is used and no negative is displayed, just flagging in case anything downstream reads that evidence expecting a 0 to 1 range.

Looks good to merge.

PR #272, opened 2026-07-27
fix(compare): rank H2D step-time statuses so regressions are flagged

### LINKED ISSUE #205: compare: straggler regressions mis-classified (H2D/RESIDUAL STRAGGLER missing from status-rank map)
## Summary

The step-time status rank map used by the compare verdict engine does not include two diagnosis statuses that the engine actually emits. Unmapped statuses silently default to rank 0 (the `NO DATA` / `WARMUP` tier), so `compare` misclassifies any run whose diagnosis is `H2D STRAGGLER` or `RESIDUAL STRAGGLER`.

Reproduced on `version_0.3.4`.

## Root cause

`src/traceml_ai/reporting/compare/policy.py` defines `_STEP_TIME_STATUS_RANK` as a closed set, and `step_time_status_rank()` looks up `dict.get(status, 0)`, so `H2D STRAGGLER` and `RESIDUAL STRAGGLER` both resolve to **0**.

`DiagnosisRegressionRule.evaluate()` in `src/traceml_ai/reporting/compare/verdict.py` flags a regression only when `rank(head) > rank(base)`, so:

- `BALANCED` (1) to a straggler (0): `0 > 1` is false, no regression is flagged.
- straggler (0) to `BALANCED` (1): `1 > 0` is true, so an actual fix is reported as `REGRESSION`.

These two statuses are live, not reserved: they are assigned in `src/traceml_ai/diagnostics/step_time/context.py` and documented in `diagnostics/DIAGNOSIS.md` (`H2D_STRAGGLER`: one rank has materially higher host-to-device transfer time; `RESIDUAL_STRAGGLER`: one rank has materially higher residual). They arise in multi-rank runs, where the comm/skew diagnosis matters most.

## Reproduction

Running the actual `DiagnosisRegressionRule` on a `version_0.3.4` checkout:

```
step_time_status_rank: RESIDUAL STRAGGLER -> 0, H2D STRAGGLER -> 0, BALANCED -> 1, INPUT STRAGGLER -> 3

BALANCED -> RESIDUAL STRAGGLER : no regression finding   (real regression missed)
BALANCED -> H2D STRAGGLER      : no regression finding   (real regression missed)
RESIDUAL STRAGGLER -> BALANCED : REGRESSION, "diagnosis worsened from RESIDUAL STRAGGLER to BALANCED"  (a fix, inverted)
BALANCED -> INPUT STRAGGLER    : REGRESSION   (control: the rule is correct when the status is in the map)
```

The control case (`INPUT STRAGGLER`, which is in the map) confirms the rule itself is correct; only the two missing keys cause the misclassification.

## Suggested fix

1. Add `"H2D STRAGGLER"` and `"RESIDUAL STRAGGLER"` to `_STEP_TIME_STATUS_RANK` at rank 3, matching the other single-phase stragglers.
2. Add a test asserting every diagnosis display-status the engine can emit is present as a key in the rank map, so a newly added or renamed status cannot silently regress to rank 0 again.

UTKARSH698:
Closes #205.

## Problem

`_STEP_TIME_STATUS_RANK` in `reporting/compare/policy.py` is missing two of the statuses the step-time diagnosis engine can emit. `step_time_status_rank()` looks the status up with `.get(status, 0)`, so an unmapped status silently ranks **0** — below `BALANCED` (1). `DiagnosisRegressionRule` flags a regression only when `rank(rhs) > rank(lhs)`, so a run that regressed to one of these statuses reads as *no change* and the verdict comes back `EQUIVALENT`.

The two missing keys on `main` are:

- **`H2D STRAGGLER`** — emitted by the straggler path (`diagnostics/step_time/context.py`), mapped from `H2D_STRAGGLER` in `_STATUS_BY_KIND`.
- **`H2D-BOUND`** — mapped from `H2D_BOUND` in `_STATUS_BY_KIND`.

Note: the issue also lists `RESIDUAL STRAGGLER`. That status no longer exists on `main` — it was renamed to `RESIDUAL-HEAVY`, which is already mapped (rank 3). So the live gap is the two H2D statuses; `H2D-BOUND` isn't called out in the issue but has the same bug.

## Change

Add both keys, placing them next to their peers so the ranking stays internally consistent:

- `H2D STRAGGLER` → **3** (same as `INPUT STRAGGLER` / `COMPUTE STRAGGLER`)
- `H2D-BOUND` → **2** (same as `INPUT-BOUND` / `COMPUTE-BOUND`)

## Tests

- **Drift guard** — `test_every_emittable_step_time_status_has_a_rank` imports `_STATUS_BY_KIND` (the single source of every display status the engine can emit) and asserts each value is a key in `_STEP_TIME_STATUS_RANK`. This is the durable fix: a future added or renamed status that isn't ranked fails here instead of silently regressing to rank 0.
- **Behavioral** — `test_compare_flags_regression_for_h2d_straggler` runs a full compare where the only change is `BALANCED → H2D STRAGGLER` (step times held equal) and asserts the verdict is `REGRESSION`.

Both tests **fail on `main`** (drift guard reports the two unmapped statuses; the behavioral test gets `EQUIVALENT`) and pass with the one-line map change. Full `tests/reporting/compare/` suite: 16 passed. `black`, `ruff`, `isort` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved step-time comparison reporting by recognizing H2D-bound and H2D straggler statuses.
  * Correctly identifies a transition to H2D straggler as a regression and includes the diagnosis in the report details.
  * Prevented unrecognized step-time statuses from receiving an incorrect default severity.


Pendu:
Thanks for this. It is a clean, well-reasoned fix, and the drift guard is a nice touch. I reviewed and tested it end to end and it is correct: the two missing rank keys, the non-vacuous tests that fail on the unpatched branch, and the guard that keys off `_STATUS_BY_KIND` so a future added or renamed status cannot silently rank 0.

One process note on where it lands. We currently integrate all changes on the `version_0.3.6` branch and merge to `main` in a batch, so we keep individual fixes off `main` directly. Because this branch was based on `main`, I ported your commit to `version_0.3.6` in #273 with your authorship preserved, so the fix ships on the active branch and reaches `main` in the next batch. Full credit to you for the fix and the guard.

I am closing this one in favor of #273. Thank you for the contribution, and for the care on the drift guard and on confirming that `RESIDUAL STRAGGLER` was already renamed rather than adding a dead key. Please keep them coming.

Issue #259, opened 2026-07-25
Show incomplete Step Time coverage in CLI and dashboard


abhinavsriva:
[Depends on Issue #257](https://github.com/traceopt-ai/traceml/issues/257)

### Problem

The live Step Time UI assumes all phase metrics are present. An incomplete window can therefore look like measured zeros, fail to update, or leave an old complete view visible.

### Proposed change

- Let CLI and dashboard consume sparse Step Time metrics.
- Render unavailable phases as `n/a` or omit them rather than displaying zero.
- Do not keep an old complete phase view when the current window is incomplete.
- Display the canonical `INCOMPLETE_DATA` diagnosis and missing-signal  explanation.
- Continue showing every valid observation from the current window.

### Acceptance

- CLI and dashboard distinguish missing values from measured zero.
- Incomplete windows show the same canonical diagnosis and explanation.
- Available phase values remain visible.
- A missing phase does not block unrelated valid metrics from updating.
- Complete windows retain current behavior.

Issue #258, opened 2026-07-25
Represent missing Step Time metrics in final summaries


abhinavsriva:
[Depends on Issue #257 ](https://github.com/traceopt-ai/traceml/issues/257)

### Problem

Final-summary rank rows and rollups currently convert unavailable Step Time phases to `0.0`. This makes missing instrumentation look like measured work and can distort averages, median/worst ranks, and card statistics.

### Proposed change

- Emit missing public Step Time metrics as `null`; retain measured zero as  `0.0`.
- Exclude missing values from average, median, worst-rank, and card  calculations.
- Include per-metric availability in final-summary evidence.
- Render missing values as `n/a`.
- Keep an H2D-only rank row instead of discarding or zero-filling it.
- Update the reporting schema version and documentation for nullable Step Time  metrics.

### Acceptance

- Missing phases are `null` in JSON and `n/a` in text.
- Measured zero remains `0.0`.
- Missing values do not affect rollups or rank selection.
- H2D-only rows retain their H2D observation.
- Compare treats a null metric as unavailable, not zero.

Issue #188, opened 2026-07-10
[Bug]: Step Time diagnosis treats missing phase timings as 0ms and can misattribute residual time


abhinavsriva:
### What happened?

When forward, backward, or optimizer timing events are missing, Step Time analysis currently treats those phases as 0.0ms and continues diagnosis without surfacing that the signals are missing. This can make compute time look artificially small and push real work into residual_proxy, leading to misleading diagnoses such as RESIDUAL_HEAVY or incorrect straggler attribution.

**Description**

Current behavior:
- Missing forward/backward/optimizer timing buckets default to 0.0ms.
- Diagnosis still runs on available signals like dataloader, H2D, and step time.
- There is no explicit user-facing indication that compute phase signals were missing.
- If step_time exists, missing compute phases are effectively absorbed into residual time.
- Final-report row usability also ignores H2D in the “anything usable?” check, so an H2D-only row can be dropped.

Expected behavior:
- Step Time should distinguish between “phase measured as zero” and “phase signal missing.”
- Reports should include coverage/missing-signal metadata for each phase.
- Diagnosis should avoid over-confident compute/residual conclusions when required phase timings are absent.
- Available signals like DL and H2D should still be analyzed, but with clear caveats.
- H2D-only rows should not be discarded if H2D timing is the available useful signal.

Acceptance criteria:
- Missing phase timings are represented explicitly, not silently as zero.
- Step Time summary/report includes missing signal information.
- Residual-heavy and compute-bound diagnoses account for missing compute coverage.
- H2D-only timing rows are handled consistently.
- Tests cover missing forward, backward, optimizer, and H2D-only scenario

PR #274, opened 2026-07-27
Preserve missing Step Time signals and gate diagnosis with INCOMPLETE_DATA

### LINKED ISSUE #257: Preserve missing Step Time signals and gate diagnosis
### Problem

The canonical Step Time window converts missing timing events to `0.0`. Diagnosis therefore cannot distinguish a measured zero from unavailable forward, backward, optimizer, H2D, input, or step timing. This can produce incorrect compute, residual, or straggler findings.

### Proposed change

- Keep measured zero as `0.0`; represent a missing internal metric by an absent  key.
- Derive `compute`, `residual_proxy`, and `total_step` only when their required inputs are present.
- Make each rule abstain when its required signals are unavailable:
  - input: input wait + step time;
  - H2D: GPU H2D + input wait + step time;
  - compute: all compute phases + input wait + step time;
  - residual: H2D + all compute phases + input wait + step time;
  - DDP straggler: backward + input wait + step time;
  - FSDP straggler: forward + backward + input wait + step time.
- Exclude unavailable components from straggler cause attribution.
-  Report `INCOMPLETE_DATA`, not `BALANCED`, when missing signals prevent areliable conclusion. Include the missing signal names in diagnosis evidence.

### Acceptance

- Missing and measured-zero timings remain distinguishable.
- Missing compute phases cannot produce `COMPUTE_BOUND` or `RESIDUAL_HEAVY`.
- Missing synchronization anchors cannot produce a rank-straggler diagnosis.
- Available input or H2D findings can still be emitted when their own inputs are complete.
- Existing complete-data behavior remains unchanged.

Pendu:
Closes #257. First of a three-PR stack (#257 -> #258 -> #259) targeting `version_0.3.6`; the stack should land together (rebase-merge or merge-commit, not squash).

## Problem

The canonical Step Time window converts missing timing events to `0.0`, so diagnosis cannot tell a measured zero from an unmeasured signal. Unmeasured compute leaks into `residual_proxy` (false RESIDUAL-HEAVY), missing input wait deflates iteration time, and a window with dark signals reports a confident BALANCED. Parent bug: #188.

## What changed

- Window builder: a metric measured in at least one aligned step of a rank's window is available; absent steps of an available metric count as zero work (gradient accumulation keeps working). A metric never measured stays an absent key, never `0.0`. Derived metrics gate on availability: `compute` needs all three phases, `total_step` needs input wait + step envelope, `residual_proxy` needs the envelope + all compute phases.
- H2D is occurrence-driven: a fully instrumented run with no host-to-device copies emits no h2d events, so absence means "no observed transfers" and contributes zero to the residual. It is never reported as a missing signal. This deviates from the issue text (which lists H2D as required for residual); taking it literally would kill RESIDUAL-HEAVY on every CPU run and flip healthy device-resident GPU runs to INCOMPLETE_DATA. Flagging for your call.
- Rules abstain when their required signals are unavailable; they still fire from ranks that measured their signals, so one dark rank does not silence a measured finding.
- New `INCOMPLETE_DATA` kind (status `INCOMPLETE DATA`, severity info): replaces BALANCED only when no rule fired and at least one abstained on missing signals. Evidence carries `missing_signals` and per-signal `signal_coverage` (`measured/observed` ranks). Metrics-only callers of `build_step_diagnosis` (no per-rank rows) keep returning BALANCED: availability is unknowable there.
- Rank stragglers: eligibility now also requires a measured input wait; cause attribution only considers components measured on both culprit and victim (measured zeros stay in the evidence maps).
- Registered everywhere a status is enumerated: CLI style, dashboard rail neutral kinds, HTML banner neutral kinds, compare rank map (tier 0, drift guard enforces), primary diagnosis routing (INSUFFICIENT_STEP_TIME_DATA with dedicated wording).
- Docs: DIAGNOSIS.md and the rank straggler policy updated in-step. CI now runs the step-time diagnosis suites.
- Demo pair: `src/dev/demo/mlp_incomplete_signals.py` (direct `model.forward()` call, so forward timing never records) fires INCOMPLETE DATA naming `forward`; the healthy baseline stays quiet.

## Notes for reviewers

- Runs with known-dark integration streams (for example the HuggingFace step-time-only stream) stop reporting BALANCED and start reporting INCOMPLETE DATA. That is the honest verdict, but it is visible.
- At this commit the dashboard hero still hard-gates on all seven metrics and will not render for windows without an h2d metric; #259 in this stack replaces that gate. Land the stack together.
- The pre-existing `test_registry.py` failures (4) and the HuggingFace conformance failure are unchanged by this stack and reproduce identically on the base commit.

## Verification

- `pytest tests/diagnostics tests/reporting tests/renderers tests/display --deselect tests/diagnostics/test_registry.py`: 367 passed.
- 23 acceptance tests in `tests/diagnostics/test_step_time_missing_signals.py`, each missing-signal scenario paired with a measured-zero or complete-data control twin; the behavioral tests fail on the base commit.
- Complete-data equivalence probed base-vs-head with hex-exact floats over 26 scenarios through the real entry points; the only intended deltas are the ones described here.
- Demo pair run through the real launcher on CPU: incomplete run reports `INCOMPLETE DATA` with `missing_signals: ["forward"]`, control reports `BALANCED`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,13 +2,13 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, 'version_*']
     paths-ignore:
       - 'docs/**'
       - '*.md'
       - '.gitignore'
   pull_request:
-    branches: [main]
+    branches: [main, 'version_*']
     paths-ignore:
       - 'docs/**'
       - '*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
             tests/core \
             tests/diagnostics/test_bands.py \
             tests/diagnostics/test_common.py \
+            tests/diagnostics/test_step_time.py \
+            tests/diagnostics/test_step_time_missing_signals.py \
             tests/telemetry/test_distributed.py \
             tests/telemetry/test_msgpack.py \
             tests/telemetry/test_sender_sequence.py \
--- a/docs/developer_guide/rank-straggler-policy.md
+++ b/docs/developer_guide/rank-straggler-policy.md
@@ -40,18 +40,22 @@ visible_r = backward_r              # DDP / default
 visible_r = forward_r + backward_r  # FSDP
 ```
 
-Only ranks with a measured visible-phase anchor and a measured step envelope are
-eligible:
+Only ranks with a measured visible-phase anchor, a measured step envelope,
+and a measured input wait are eligible:
 
 ```text
-DDP / default: backward_r > 0 and step_time_r > 0
-FSDP:          forward_r > 0 and backward_r > 0 and step_time_r > 0
+DDP / default: input_wait measured and backward_r > 0 and step_time_r > 0
+FSDP:          input_wait measured and forward_r > 0 and backward_r > 0
+               and step_time_r > 0
 ```
 
 If fewer than two ranks are eligible, TraceML does not report a rank straggler.
-Missing visible instrumentation makes the rule abstain rather than treat a
-zero as a fast rank. Component values of `input_wait == 0` and `h2d == 0` are
-still valid measurements and are kept.
+Missing instrumentation makes the rule abstain rather than treat a zero as a
+fast rank; when that abstention leaves the whole window without any finding,
+Step Time reports `INCOMPLETE_DATA` with the missing signal names instead of
+`BALANCED`. Component values of `input_wait == 0` and `h2d == 0` are still
+valid measurements and are kept: a metric measured somewhere in the window is
+never confused with one that was never measured at all.
 
 ## Culprit and victim selection
 
@@ -84,7 +88,8 @@ the victim to explain the visible wait:
 
 ```text
 input_excess   = input_wait_culprit - input_wait_victim
-h2d_excess     = h2d_culprit - h2d_victim
+h2d_excess     = h2d_culprit - h2d_victim           # only when H2D is
+                                                    # measured on both ranks
 forward_excess = forward_culprit - forward_victim   # DDP / default only
 ```
 
@@ -96,11 +101,14 @@ cause_coverage = min(1, component_excess / visible_cost)
 
 The component with the highest coverage names `INPUT_STRAGGLER`,
 `H2D_STRAGGLER`, or (DDP / default only) `COMPUTE_STRAGGLER` only when its
-coverage reaches 80%. DDP / default compute attribution requires measured
-forward time on both the culprit and victim. Otherwise the diagnosis stays
-`STRAGGLER` with a sync-or-unattributed component. The issue evidence includes
-both `component_excesses_ms` and `component_coverage`, keyed by `input`,
-`h2d`, and `compute`.
+coverage reaches 80%. Every candidate component must be measured on both the
+culprit and the victim rank: DDP / default compute attribution requires
+measured forward time on both, and H2D attribution requires measured H2D time
+on both, so an unmeasured phase can never name the cause. Otherwise the
+diagnosis stays `STRAGGLER` with a sync-or-unattributed component. The issue
+evidence includes both `component_excesses_ms` and `component_coverage`, keyed
+by `input`, `h2d`, and `compute`; components excluded for missing measurement
+are absent from both maps.
 
 Cause coverage chooses the diagnosis label. The straggler score above remains
 the sole source of warning or critical severity. Generic residual attribution
--- a/src/dev/demo/mlp_incomplete_signals.py
+++ b/src/dev/demo/mlp_incomplete_signals.py
@@ -1,0 +1,91 @@
+"""
+Incomplete-signals scenario for the Step Time INCOMPLETE_DATA diagnosis.
+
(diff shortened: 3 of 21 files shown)


abhinavsriva:
The overall direction looks good but  could you please check these cases?

- Intermittent coverage: availability becomes true when a metric appears in any step, after which missing occurrences are filled with 0.0. I reproduced forward timing present in only 1/30 steps being averaged as 1.33 ms, with the missing work producing RESIDUAL_HEAVY instead of INCOMPLETE_DATA. Should zero-filling be restricted to occurrence-driven metrics such as optimizer/H2D, or should we track per-step coverage for forward/backward/input/step?

**Fix:** make zero-fill metric-specific. Valid for `optimizer_step` (grad accumulation)  and `h2d` (occurrence-driven). Not valid for `forward` / `backward` / `input_wait` /  `step_time`, which must occur every step,  absence there is lost instrumentation and should drop availability, not synthesize a zero.
  

- Worst value/rank mismatch: with r0=step 100 ms + input available and r1=step 400 ms + input missing, the step metric reports worst_total=400 ms but worst_rank=r0. The value is calculated across all step-time ranks, then the rank is overridden using only total_step-eligible ranks. The value and rank should come from the same candidate set.

**Fix:** source value and rank from one ordering. Either drop the override, or carry the  `total_step` ranking as its own field instead of relabeling the `step_time` summary.

-  Final summary re-derives the residual it declined to compute  `public_step_time_metric_values` (`step_time_window.py:607-620`) collapses every missing key via `.get(key, 0.0)` and returns `Dict[str, float]`, so missing and measured-zero are   indistinguishable. Worse, `:617` — `residual_proxy` is absent precisely because   `_add_derived_step_metrics:261` refused to derive it without the compute phases, and the   fallback derives it anyway with `compute = 0`, giving `residual = step_time`.

  | Available | Diagnosis | Summary output |
  |---|---|---|
  | input 2 ms + step 100 ms | INCOMPLETE_DATA | compute 0, residual 100 |
  | input 40 ms + step 100 ms | INPUT_BOUND | compute 0, residual 100 |
  | step 100 ms only | INCOMPLETE_DATA | input 0, compute 0, residual 100 |

**Fix:** preserve absence through the projection (`Optional[float]` or omit the key) and drop the `residual is None` fallback, if the window declined to derive it, the summary must not.

abhinavsriva (changes requested):
There are some issues when step time has partial coverage. Please see comment in conversation.

Pendu:
Thanks, these were all correct. Fixed at head 0d03c4b; each fix has a test pinning it.

**1. Intermittent coverage.** Availability is now metric-specific rather than "present in any step". `optimizer_step` and `h2d` are occurrence-driven (available when measured in at least one aligned step, absent steps counted as true zeros), but `forward`, `backward`, `input_wait`, and `step_time` must be measured on every aligned step; intermittent presence there drops availability instead of averaging a fragment. So forward present in 1/30 steps now reports INCOMPLETE DATA with `forward` in the missing signals, not RESIDUAL_HEAVY. See `step_time_window.py` `_rank_metric_availability` (the `count == total or metric_key in _OCCURRENCE_METRICS` gate) and `test_intermittent_forward_drops_availability`.

**2. Worst value/rank mismatch.** I took the drop-the-override path. Each metric summary's `worst_total` and `worst_rank` now come from one ordering: the rank named is the rank that produced the reported value for that same metric. The run-level slowest rank by `total_step` is still computed, but as its own value threaded into the diagnosis (`_overall_worst_rank` in the diagnosis path), not by relabeling the step_time summary. `test_worst_value_and_rank_come_from_one_candidate_set` covers your r0/r1 case, and a self-consistency invariant asserts it holds across many presence patterns.

**3. Projection re-deriving the declined residual.** `public_step_time_metric_values` now returns `Dict[str, Optional[float]]` and preserves absence: a metric the window declined to derive is `None`, not a fabricated number. The `residual = step_time - h2d - compute` fallback is gone, so when the window withholds `residual_proxy` the projection reports `residual_ms = None` rather than `residual = step_time` with `compute = 0`. Covered by `test_public_projection_preserves_absence`.

I also added `test_step_time_invariants.py`, a property-test module that fuzzes the presence pattern each rule quantifies over (every required metric at partial coverage, self-consistency across two-rank presence combinations, the projection over each compute phase) plus a drift guard that fails if a new selected metric is added without an occurrence-vs-required classification. This pins the three classes, not just the reported instances.

Full suite is green at head; ready for another look.

abhinavsriva:
Thanks, #257’s scope is met.

 I verified the key missing, intermittent, measured-zero, rank-consistency, and complete-data cases by reproduction, and the new suites pass.
 
The remaining end-to-end work is correctly owned by:

#258: nullable final-summary rows, rollups, comparisons, and n/a rendering.
#259: sparse CLI/dashboard rendering and stale-view handling.

Please continue with #258 and #259 while preserving the canonical rule: absent means unmeasured; 0.0 means measured zero. #258 should include an end-to-end test through rank rows and rollups.

We can merge this into version_0.3.6 as the first stack layer. The overall change is complete only after #258 and #259 are merged and the full stack passes integration testing. 

**Do not release or merge version_0.3.6 onward before then.**

PR #277, opened 2026-07-28
Represent missing Step Time metrics as null in final summaries (schema 1.7)

### LINKED ISSUE #258: Represent missing Step Time metrics in final summaries

Pendu:
Closes #258. Depends on #257 (second PR of the stack targeting `version_0.3.6`).

## Problem

Final-summary rank rows and rollups convert unavailable Step Time phases to `0.0`, so missing instrumentation looks like measured work: averages dilute, median and worst-rank picks can land on ranks that never measured the metric, and an H2D-only rank distorts every total-step statistic.

## What changed

- Every public Step Time metric is now nullable: `null` in JSON means the signal was never measured in the analyzed window; a measured zero stays `0.0`. All ten metric keys are always present, so the section schema validator and existing consumers keep exact key parity.
- Rollups (`global.average`, `global.median`, `global.worst`) and median/worst-rank selection exclude null values. A rank with only some metrics measured keeps its row with `null` for the others; an H2D-only rank keeps its H2D observation and stays out of total-step picks. The single-rank card wording no longer depends on the rank winning a median/worst pick.
- The residual fallback that re-derived `residual_ms` from partial sums is gone: a residual the window could not derive stays `null` instead of silently absorbing an unmeasured phase.
- Text renders missing metrics as `n/a`; HTML renders an em dash; compare treats a null side as unavailable (no delta), never as a drop to zero.
- `compare` and the HTML phase bar no longer conflate "genuinely null" (schema>=1.6, signal never measured) with "key absent" (pre-1.6 payload): both now check key presence, not just value, before falling back to `dataloader_ms` for the input-wait figure. The old value-only check meant a schema-1.7 payload with a real, unmeasured input wait could silently render a different metric's number under the "Input" label.
- Schema version 1.6 -> 1.7 via a named `SCHEMA_VERSION` constant; SCHEMA.md, the compare version warning, the architecture glossary, and the user guides updated. CI now runs the summary reporting suites.

## Notes for reviewers

- Visible change: CPU-only runs now report `h2d_ms: null` (previously a fabricated `0.0`). H2D events only occur when transfers happen, so null is the truthful value there.
- A downstream consumer of `final_summary.json` needs a small companion change for 1.7 (new status registration plus one null guard); prepared separately.

## Verification

- `pytest tests/diagnostics tests/reporting tests/renderers tests/display --deselect tests/diagnostics/test_registry.py`: 410 passed.
- New acceptance tests drive real SQLite fixtures end to end: missing-vs-zero h2d, H2D-only rank row retention and rollup exclusion, all-ranks-without-total rank picks, partial compute triplet nullability, compare null handling, text `n/a` rendering, single-rank card wording with a null total. The db-driven tests fail on the #257 base.
- Complete-data payloads are byte-identical to the base through the summary entry point over five scenario databases.
- Real hardware: ran a healthy 4-GPU DDP baseline, a single-rank incomplete-signals scenario, and a new 4-GPU DDP scenario with genuinely asymmetric H2D coverage (one rank's data pre-placed on-device, so its H2D timer never arms) through the actual launcher on a `g4dn.12xlarge`. The device-resident rank reports `h2d_ms: null` and is correctly excluded from H2D rank selection; the diagnosis stays a confident verdict rather than falling back to INCOMPLETE_DATA on the strength of the other measured signals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,7 @@ jobs:
             tests/runtime/test_telemetry_publisher.py \
             tests/reporting/compare \
             tests/reporting/html \
+            tests/reporting/summary \
             -q
 
   smoke-test-install:
--- a/docs/developer_guide/architecture.md
+++ b/docs/developer_guide/architecture.md
@@ -108,6 +108,6 @@ Architectural risks and known structural debt. Day-to-day bugs live in the issue
 | RESIDUAL_HEAVY | A large window-wide share of step time is unattributed residual time. |
 | HIGH_PRESSURE / IMBALANCE | GPU memory is near capacity, or uneven across ranks. |
 | CREEP_EARLY / CREEP_CONFIRMED | Direction-confirmed GPU-memory growth across the run, early or confirmed. |
-| final_summary | The end-of-run `final_summary.{json,txt}`; the JSON carries `schema_version` (currently 1.6). |
+| final_summary | The end-of-run `final_summary.{json,txt}`; the JSON carries `schema_version` (currently 1.7). |
 | Wire envelope | The per-batch message, `{meta, body: {tables}}`, sent as a msgpack frame behind a 4-byte length prefix. |
 | NoOpRuntime | The inert runtime the system falls back to if instrumentation boot fails (fail-open). |
--- a/docs/user_guide/compare.md
+++ b/docs/user_guide/compare.md
@@ -193,6 +193,9 @@ That means:
 This helps keep compare useful across incremental TraceML releases.
 For Step Time input timing, schema 1.6 compares selected-clock `input_wait_ms`
 when present and falls back to legacy `dataloader_ms` for older summaries.
+Since schema 1.7, a Step Time metric can be `null` when its timing signal was
+never measured in the analyzed window; compare treats a null side as
+unavailable (no delta) instead of reading it as zero.
 
 ---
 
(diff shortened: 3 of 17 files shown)


abhinavsriva (approved):
Approved. 

The #258 acceptance criteria are met. Per-metric availability is derivable from the nullable values in each rank row of the payload rather than exposed as a separate aggregate coverage block;  ote this holds for final_summary.json and the HTML rank table, not the text evidence table, which shows aggregates only. If explicit measured/observed counts are needed later, we can add them separately without blocking this PR.

Issue #268, opened 2026-07-27
[RFC] Install friction: a dependency policy so pip install traceml-ai never fights the user's stack


Pendu:
## Summary

Issue #263 reported that `numpy<2` makes TraceML conflict with a modern scientific stack. Investigating it surfaced a class of problem rather than a single bad line, so this RFC collects the class and the policy that closes it.

Installing TraceML today does one of two things to a practitioner on a current stack:

- With pip: it uninstalls NumPy 2 and downgrades to 1.26.4 in place, with no error. Any package already compiled against the NumPy 2 ABI is now broken.
- With a resolver that will not downgrade (uv): it silently backsolves to `traceml-ai==0.1.6`, a December 2025 build. Every feature since then is invisible. The user does not get an error, they get an old TraceML.

`numpy<2` first shipped in 0.1.8 (2025-12-25) and is in every release since, so 25 of the 30 published releases are unreachable for anyone requiring NumPy 2.

A second, independent source of friction: a default install pulls 60 packages and 239 MB, because the NiceGUI web stack is a hard runtime dependency. Without it the same install is 9 packages and 65 MB.

## Why this is a category, not a bug

The pin was introduced by flipping `numpy>=2.0.0` to `numpy<2` inside an unrelated commit. Nothing could have caught it: CI installs with the same manifest it is validating, so the pin is always satisfiable, every leg resolves NumPy 1.26.4, and CI is green forever. The defect is only visible from outside, in an environment CI never constructs.

That is the general shape to defend against: a constraint that is invisible to us and load-bearing for the user.

## Policy proposed

1. No upper bound on a runtime dependency without a recorded, concrete incompatibility.
2. Every runtime dependency that the code calls a specific API on carries a lower bound.
3. Anything not imported by the core import path is an extra, not a runtime dependency.
4. CI keeps one leg that resolves the newest published dependencies, so ecosystem drift is heard from us rather than from a user.

## Children

- Unpin NumPy and add the upper-bound guard (PR ready)
- Remove `plotly`, which is never imported
- Add a `nicegui` lower bound for the `ui.context` API in use
- Make the dashboard stack a real extra instead of a core dependency

Pendu:
Children, now filed:

- #266 (PR) guard runtime deps against undocumented upper bounds
- #267 release 0.3.6, so the NumPy fix reaches users
- #269 remove plotly, which is never imported
- #270 add a nicegui lower bound for the ui.context API in use
[...]

PR #279, opened 2026-07-29
Add nicegui>=1.4.23 floor; ui.context is absent on older releases

### LINKED ISSUE #270: Add a lower bound for nicegui; ui.context is not available on older releases
`nicegui` is declared with no version floor. The display driver calls `ui.context` at:

- `src/traceml_ai/aggregator/display_drivers/nicegui.py:285`
- `src/traceml_ai/aggregator/display_drivers/nicegui.py:380`
- `src/traceml_ai/aggregator/display_drivers/nicegui.py:419`

`ui.context` was introduced in nicegui 1.4.23. Because `nicegui` has a large dependency surface (fastapi, starlette, uvicorn and others), an unrelated constraint elsewhere in a user's environment walks it backwards. With `fastapi==0.100.0` a resolver selects `nicegui==1.4.7`, where `ui.context` does not exist and the driver raises `AttributeError`.

Note `getattr(ui.context, "client", None)` does not protect this: the default applies to the attribute lookup on `ui.context`, not to `ui.context` itself.

Proposed: add a floor. `nicegui>=1.4.23` is the minimum for the API in use; a higher floor may be preferable given the ECharts and theme work.

Acceptance: resolving against an old fastapi either selects a NiceGUI with `ui.context` or fails loudly, instead of installing a version that breaks at runtime.

Part of #268.

Pendu:
Closes #270. Part of #268.

`nicegui` was declared with no version floor while the display driver calls `ui.context` at three sites. Because nicegui has a large dependency surface, an unrelated pin elsewhere in a user environment (an old fastapi, for example) walks nicegui back to a release where `ui.context` does not exist, and the dashboard raises `AttributeError` at runtime instead of failing at resolve time.

## Changes

- `nicegui>=1.4.23` in `dependencies` and the `dashboard` extra, with a comment recording why the floor exists.
- Regression test in `test_packaging_constraints.py` mirroring the numpy guard: fails if the floor is ever dropped.

## How the floor was chosen

Verified by execution, not changelog. All 15 nicegui APIs the display drivers touch (`ui.element`, `ui.label`, `ui.row`, `ui.html`, `ui.echart`, `ui.timer`, `ui.context`, `ui.run`, `ui.page`, `ui.link`, `ui.column`, `ui.add_head_html`, `app.on_startup`, `app.on_disconnect`, `app.add_static_files`) were checked for presence against installed copies of nicegui 1.4.7, 1.4.22, 1.4.23, and 2.0.0:

```
nicegui 1.4.7:  MISSING: ['ui.context']
nicegui 1.4.22: MISSING: ['ui.context']
nicegui 1.4.23: ALL 15 APIs PRESENT
nicegui 2.0.0:  ALL 15 APIs PRESENT
```

1.4.23 is exactly the minimum for the API surface in use.

## Verification

- `tests/core/test_packaging_constraints.py`: 13 passed (includes the new guard).

## Note

Trivial planned conflict with #278 on adjacent `pyproject.toml` lines (plotly removal). Merge in either order and take both changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pinned the minimum required NiceGUI version for runtime usage (including the dashboard option) to ensure the needed UI context behavior is present and avoid compatibility-related runtime errors.

* **Tests**
  * Added a regression test that verifies the package dependency specifier continues to enforce the expected NiceGUI “floor” version covering UI context support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,14 +47,17 @@ dependencies = [
     "numpy>=1.26",
     "msgspec",
     "pyyaml>=6.0",
-    "nicegui",
+    # Floor: the dashboard display driver calls ui.context, added in
+    # nicegui 1.4.23. Older nicegui imports fine and then raises
+    # AttributeError at runtime.
+    "nicegui>=1.4.23",
     "plotly",
 ]
 
 [project.optional-dependencies]
 
 dashboard = [
-    "nicegui",
+    "nicegui>=1.4.23",
     "plotly",
 ]
 
--- a/tests/core/test_packaging_constraints.py
+++ b/tests/core/test_packaging_constraints.py
@@ -174,3 +174,27 @@ def test_numpy_is_not_capped_below_version_2() -> None:
             f"numpy is capped below 2 by {requirement!r}. This downgrades "
             "NumPy in place for every user on a modern scientific stack."
         )
+
+
+def test_nicegui_floor_covers_ui_context() -> None:
+    """Regression guard for issue #270.
+
+    The display driver calls ui.context, which nicegui added in 1.4.23.
+    Without a floor, an unrelated constraint elsewhere in the user's
+    environment (an old fastapi pin, for example) walks nicegui back to
+    a version that imports fine and raises AttributeError at runtime.
+    """
+    for requirement in _runtime_dependencies():
+        if _requirement_name(requirement) != "nicegui":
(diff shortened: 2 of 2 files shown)


abhinavsriva (commented):
Is the floor good and we assume that we don't need to do earlier ?

Issue #263, opened 2026-07-26
[Bug]: NumPy<2 dependency


GANs-and-roses:
### What happened?

A lot of packages require NumPy>2 at this point. Thus, traceml conflicts with so many packages and use of traceml requires sacrificing many packages, sometimes the main AI library like Monai.

P.S.: Funnily, my dependency software, uv, installs traceml-ai==0.1.3 because it cannot downgrade NumPy. 

Would it be possible to upgrade to NumPy>2?

abhinavsriva:
Thanks for opening this, you are right. 

Newer TraceML releases use numpy<2, so uv backtracks to 0.1.3 in a NumPy 2 environment. That is an overly restrictive compatibility constraint on our side.

We tested against NumPy 2 and remove the upper bound.

Pendu:
Follow-up for anyone finding this later. #265 removed the cap on `main`, but the newest release on PyPI is still 0.3.5, whose metadata declares `numpy<2`, so the behaviour described here is still what an install gives you today. #267 tracks cutting the release that carries the fix.

@GANs-and-roses one clarification on the resolver behaviour you saw. In a clean environment the back-solve lands on `traceml-ai==0.1.6`, which is the newest release that still allowed NumPy 2. Reaching 0.1.3 additionally requires something in the environment to conflict with the nicegui stack that arrived in 0.1.5, so your 0.1.3 is consistent with this, just further back than the general case.

Wider follow-ups from this report are collected in #268.

GANs-and-roses:
I see. If it is helpful, here are the related parts from my pyproject.toml:
```
requires-python = ">=3.13"
dependencies = [
    "backports-zstd>=1.6.0",
    "bayesian-optimization>=3.3.0",
    "bottleneck>=1.6.0",
    "debugpy>=1.8.21",
    "einops>=0.8.2",
    "jinja2>=3.1.6",
    "kornia>=0.8.3",
    "matplotlib>=3.11.1",
    "mlflow>=3.14.0",
    "monai>=1.6.0",
    "numexpr>=2.14.2",
    "numpy>=2.4.4",
    "nvidia-ml-py>=13.610.43",
    "pandas>=2.3.3",
    "pandas-stubs>=3.0.3.260530",
    "pdbp>=1.8.2",
    "pudb>=2025.1.5",
    "pyarrow>=24.0.0",
    "pydicom>=3.0.2",
    "pyrsmi>=1.1.0",
    "python-calamine>=0.8.2",
[...]

PR #280, opened 2026-07-29
CONTRIBUTING: add the dependency policy from #268

Pendu:
Part of #268 (does not close it; the release, the yank, and the dashboard-extra decision remain).

Adds a Dependency Policy section to CONTRIBUTING.md, between "What We Are Not Looking For (Yet)" and "Releasing (maintainers)". It records the four rules from the RFC as contributor-facing policy:

1. No upper bounds without a recorded reason (enforced by the #266 guard test's allowlist).
2. Lower bounds follow the APIs we call, with the reason commented in `pyproject.toml` (the #270 pattern).
3. Not imported by the core import path means extra, not runtime dependency. The dashboard stack is called out honestly as the known open case, deferred to #268.
4. A scheduled newest-deps CI leg exists so ecosystem drift is our alarm, not a user's bug report.

Opens with the #263 story in two sentences as the motivation, and closes with two references (Henry Schreiner's upper-bound essay and SPEC 0), both link-checked before inclusion.

Docs-only change; no code touched.

## Summary by CodeRabbit

* **Documentation**
  * Added a new dependency policy section to the contributor guide, detailing how to set version constraints and document runtime upper bounds.
  * Clarified guidance for deriving lower bounds from required APIs and for organizing non-core dependencies as optional extras.
  * Documented CI checks to detect ecosystem drift by testing against the newest published versions, along with references for best practices.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -164,40 +164,35 @@ To keep the project focused, we are currently **not** accepting:
 
 ## Dependency Policy
 
-`pyproject.toml` is user-facing API. A wrong constraint does not break our
-CI; it breaks a stranger's environment, silently. A `numpy<2` cap shipped
-this way once: CI stayed green while `pip install traceml-ai` downgraded
-NumPy 2 in place, and resolvers that refuse to downgrade backsolved users
-to a months-old release instead (#263). These rules exist so that class of
-defect cannot ship again.
+Dependency constraints affect downstream package resolution and can create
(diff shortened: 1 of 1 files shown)


abhinavsriva (changes requested):
I agree with the policy and want to keep the safeguards. Since this is public contributor guidance, could we make the introduction a little more neutral? The current incident-focused wording feels slightly corrective.

abhinavsriva (approved):

Issue #267, opened 2026-07-27
Release 0.3.6: the NumPy fix has not reached users yet


Pendu:
#265 fixed the NumPy cap on `main`, and #263 is closed, but nobody installing TraceML has the fix yet. The newest release on PyPI is 0.3.5, and its metadata still declares `numpy<2`:

```
$ curl -s https://pypi.org/pypi/traceml-ai/json | python -c "import json,sys; d=json.load(sys.stdin); print(d['info']['version'], [r for r in d['info']['requires_dist'] if 'numpy' in r.lower()])"
0.3.5 ['numpy<2']
```

So the behaviour reported in #263 is still what a user gets today. With pip, installing traceml-ai uninstalls NumPy 2 and puts 1.26.4 in its place, with no error:

```
numpy BEFORE traceml: 2.5.1
  Attempting uninstall: numpy
    Found existing installation: numpy 2.5.1
    Uninstalling numpy-2.5.1:
      Successfully uninstalled numpy-2.5.1
numpy AFTER traceml: 1.26.4
```

With a resolver that will not downgrade, the conflict is invisible in a different way: `uv pip install "numpy>=2" traceml-ai` resolves to `traceml-ai==0.1.6`, a December 2025 build. The user gets an old TraceML rather than an error.

`numpy<2` first shipped in 0.1.8 and is in every release since, so 25 of the 30 published releases are unreachable for anyone who needs NumPy 2. Republishing does not help; the fix has to go out under a new version.

## Ask

Cut 0.3.6 from `main` so the fix reaches users.

## Before cutting

`version_0.3.5` still contains `numpy<2` and has commits that are not on `main`. When that branch merges forward, a conflict resolved the wrong way would put the cap back into a release. #266 adds a test that fails if that happens, which is worth landing first.

## Acceptance

- 0.3.6 is on PyPI and its metadata does not cap NumPy.
- In a clean environment with NumPy 2 installed, `pip install traceml-ai` leaves NumPy 2 in place.
- `uv pip install "numpy>=2" traceml-ai` resolves to the current release rather than 0.1.6.

Pendu:
Shipped as 0.3.5.1 (a metadata/packaging hotfix number rather than 0.3.6, which stays reserved for the feature branch in progress). All three acceptance criteria verified against live PyPI today:

- The release is on PyPI and its metadata does not cap NumPy: `numpy>=1.26`, alongside the plotly removal and the `nicegui>=1.4.23` floor.
[...]

PR #282, opened 2026-07-29
Release prep for 0.3.5.1: accept X.Y.Z.W versions, roll CHANGELOG

Pendu:
Prep for cutting 0.3.5.1 (#267): the patch release that carries the NumPy fix and the install-friction work to PyPI.

## Changes

- `release.yml`: the version validation accepted only `X.Y.Z`, so the "Cut a release" button would reject `0.3.5.1`. The regex now also accepts a four-component `X.Y.Z.W`, with the error text updated to match. Everything else in the workflow is untouched; the version still flows through the quoted `$VERSION` env var.
- `CHANGELOG.md`: the Unreleased section now lists everything queued for this release: the NumPy 2 uncap, the Windows aggregator fix, the plotly removal, the nicegui floor, the dependency guards, the dependency policy, the GitHub-Release step, and this version-format change.

## Merge order

Two Unreleased bullets describe #278 (plotly) and #279 (nicegui floor). Merge those first so the changelog states facts; if either is rejected, drop its bullet here.

## After merging

Actions -> "Cut a release" -> `0.3.5.1`. The run tags, builds, publishes to PyPI via Trusted Publishing, and creates the GitHub Release entry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,6 +56,38 @@ jobs:
           git tag -a "v$VERSION" -m "Release v$VERSION"
           git push origin "v$VERSION"
 
+      - name: Warn when runtime dependencies changed since the last release
+        # The conda-forge autotick bot only bumps version and sha256 in its
+        # feedstock PR; dependency changes must be applied there by hand.
+        # This step makes that need visible in the run summary at the exact
+        # moment a release is cut, instead of relying on someone remembering.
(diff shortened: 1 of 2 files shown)


abhinavsriva (dismissed):
Please modify contributing.md and merge that before this one.

abhinavsriva (approved):

PR #283, opened 2026-07-29
Show incomplete Step Time coverage in the live CLI and dashboard

### LINKED ISSUE #259: Show incomplete Step Time coverage in CLI and dashboard

Pendu:
Closes #259. Last PR of the three-PR stack targeting `version_0.3.6`; #257 (#274) and #258 (#277) are already merged.

## Problem

The live Step Time UI assumes every phase metric exists. A sparse window could freeze the dashboard hero on its last complete view, the CLI kept a dead run's final table on screen indefinitely, and incomplete coverage was invisible.

## What changed

- CLI: the renderer-side cache with no expiry is gone; the computer's own 30-second last-ok window is the only bridge for transient gaps. Once it expires, the panel says the window expired instead of re-serving a dead table. Before any data arrives the calm waiting message is kept (warm-up is not an alarm). Unmeasured phases are omitted from the table instead of rendering `0.0 ms`; the diagnosis block shows `INCOMPLETE DATA` with the missing signal names.
- Dashboard hero: the all-metrics-present gate is replaced with sparse rendering. Measured phases draw against the median iteration envelope (input wait + step envelope), so unmeasured time shows as empty ribbon space instead of stretching measured phases to fill the bar; for self-consistent complete windows the widths are unchanged. H2D is occurrence-driven, so its absence never counts as partial coverage, matching the engine.
- Dashboard hero, dead-run signal: `StepCombinedComputer`'s `had_ok` (already used by the CLI) is now also carried on the `StepCombinedTimeResult` payload itself, since the dashboard subscriber only ever sees the payload, not the computer. Once a run stops reporting and the stale window expires, the card visibly clears (ribbon, KPIs, window label) instead of freezing on the last complete view.
- Dashboard hero, unmeasured vs measured-zero: a non-H2D phase that was never measured this window (most importantly `input_wait`, the dataloader-fetch signal, see #88/#135) now renders as a fixed hatched sliver, never a proportional width. Without this, an absent `input_wait` alone left the other phases summing to exactly `step_time` (since `residual_proxy` is a remainder), so the ribbon filled to 100% and a dark dataloader stream was visually indistinguishable from a confirmed-fast one. The window label also names the missing signal(s) (`partial: FWD,RESIDUAL`) instead of a generic `partial signals`.
- Dashboard hero, sliver styling: the hatched sliver is drawn in the phase's own color (blue for forward, gold for residual), so it still identifies which phase is missing while the diagonal hatch marks "unmeasured, not zero". It carries no on-segment text (it would overlap the hatch illegibly at that width; the window meta already names the missing phases).
- Dashboard hero, legend: the phase legend spells out the labels ("Input wait", "Forward", ...) instead of the terse "IW"/"FWD"; the ribbon segments keep the compact abbreviations so a narrow slice never overflows.
- Docs: the live CLI and dashboard behavior notes in the reading-output guide. CI now runs the renderer and display suites.

## Notes for reviewers

- Startup behavior changed deliberately: an expired run now shows an expiry message instead of the last table (CLI) or a cleared card (dashboard); a fresh run still shows the waiting message.
- The sibling step-memory and model-diagnostics renderers keep their own unbounded caches with no dead-run signal. That is out of scope for this stack; a candidate follow-up issue.
- Two of the fixes here (the dashboard dead-run signal and the unmeasured-`input_wait` rendering, both described above) address pre-existing behavior rather than anything the first #259 commit introduced. Both are called out directly by the issue text ("leave an old complete view visible").

## Verification

- `pytest tests/diagnostics tests/reporting tests/renderers tests/display --deselect tests/diagnostics/test_registry.py`: 423 passed.
- Full suite (`pytest tests/ --deselect tests/diagnostics/test_registry.py`, GPU/multiprocess excluded): 675 passed, 20 pre-existing failures reproduced identically on base `upstream/version_0.3.6` (unrelated integrations/launcher/H2D-patch tests). Zero regressions.
- The stale-TTL bridge (the property the cache removal leans on) is pinned with a computer-level test: bridged within the TTL with a STALE status, empty after expiry; `had_ok` is asserted on both the computer and the returned payload, including a cold-start (never-had-data) control.
- New tests cover the sparse hero (missing phases, measured-zero twin, absent-H2D twin, step-time-only extreme, payload without a step envelope), multi-rank sparse CLI rendering, calm startup, dead-view expiry (CLI and dashboard), and the dark-`input_wait`-vs-measured-zero-`input_wait` pair. The load-bearing ones fail with the source changes reverted.

## Validation on real hardware

Setup: AWS `g4dn.12xlarge`, 4x NVIDIA Tesla T4 (16 GB each), Amazon Linux 2023
DLAMI, PyTorch 2.11. Each scenario ran through the real `traceml run` launcher
(not a unit test) with a small MLP under DDP. The live views show a rolling
window of the last 100 aligned steps; each run stepped continuously for a few
minutes so the window is stable. Purpose-built scenario scripts force each
coverage case.

### Dashboard (`--mode=dashboard`)

**1. Partial coverage, forward signal dark (1 GPU).** The scenario calls
`model.forward()` directly, so `nn.Module.__call__` is never hit and forward
timing is never recorded. The hero ribbon renders the unmeasured phases
(Forward, Residual) as hatched slivers in their own colors rather than as
zero-width segments, the window label reads `partial: FWD,RESIDUAL`, the
Residual Share shows `n/a`, and the verdict is `INCOMPLETE DATA`. GPU
utilization, step memory, and GPU temperature are populated from the real
device.
gpu_partial_1gpu

**2. Complete coverage (4 GPU, control).** Every phase is measured. The ribbon
is fully solid with no partial marker and a real Residual Share, which confirms
the sparse-rendering path does not change the healthy case. The `STRAGGLER`
verdict is a cross-rank finding on this hardware ("r1 slower than r2 by
~16%"), not related to this change.
gpu_healthy_4gpu

**3. No H2D anywhere (4 GPU).** Every rank's dataset is materialized on the
device, so no host-to-device copy ever happens and no H2D event is emitted.
H2D is occurrence-driven, so its absence is treated as a measured zero: the H2D
segment is a plain 0%-width solid segment (no hatch), there is no `partial`
marker, and the residual is still derivable. This is the frontend view of the
decision made in #257.
gpu_no_h2d_4gpu

### CLI (`--mode=cli`)

Rendered from the same runs' telemetry through the real CLI renderer.

**4. Partial coverage, forward dark.** The table omits the unmeasured Forward,
H2D, and Residual columns instead of printing `0.0 ms`, and the diagnosis block
reads `INCOMPLETE DATA` naming `forward`.
cli_partial

**5. Complete coverage (control).** The full phase table with a real verdict
and Residual Share.
cli_healthy

**6. Warm-up (never had data).** A calm waiting message, not an alarming
`NO DATA` verdict.
cli_startup
cli_expired

**7. Dead run (window expired).** Once a run stops reporting and the stale
window passes its TTL, the panel says the last window expired instead of
re-serving a stale complete table.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,6 @@ jobs:
             tests/telemetry/test_distributed.py \
             tests/telemetry/test_msgpack.py \
             tests/telemetry/test_sender_sequence.py \
-            tests/renderers \
-            tests/display \
             -q
 
   test-integration:
@@ -89,7 +87,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[torch]"
+          pip install -e ".[torch,dashboard]"
           pip install pytest
       - name: Run integration tests
         run: |
@@ -99,6 +97,8 @@ jobs:
             tests/reporting/compare \
             tests/reporting/html \
             tests/reporting/summary \
+            tests/renderers \
+            tests/display \
             -q
 
   test-latest-deps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,8 @@ jobs:
             tests/telemetry/test_distributed.py \
             tests/telemetry/test_msgpack.py \
             tests/telemetry/test_sender_sequence.py \
+            tests/renderers \
+            tests/display \
             -q
 
   test-integration:
--- a/docs/developer_guide/architecture.md
+++ b/docs/developer_guide/architecture.md
@@ -102,6 +102,7 @@ Architectural risks and known structural debt. Day-to-day bugs live in the issue
 | H2D | Host-to-device copy (CPU to GPU), timed by patching `Tensor.to`. |
 | Phase | A timed part of a step: input wait, H2D, forward, backward, or optimizer. |
 | Step / trace_step | A training iteration; `with trace_step(model)` marks the boundary that phase timing is computed against. |
+| Step Time availability | `StepTimeWindow` carries the expected rank universe and sparse per-rank metrics. Measured-rank populations and multi-metric eligibility are derived from those facts; consumers do not maintain a second coverage model. |
 | Residual (residual_proxy) | Residual step time, `max(0, step - h2d - forward - backward - optimizer)`. |
 | INPUT_BOUND / COMPUTE_BOUND | Input wait is material, or compute reaches 90% of typical selected-clock iteration time. |
 | INPUT_STRAGGLER / COMPUTE_STRAGGLER / H2D_STRAGGLER / STRAGGLER | Visible rank skew exists; the label names the culprit's material input, DDP-forward, or H2D excess, or `STRAGGLER` when the skew is sync-bound or unattributed. |
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -104,7 +104,10 @@ live step-time table rather than shown as `0.0 ms`, and the diagnosis block
 shows `INCOMPLETE DATA` with the missing signal names when no reliable
 conclusion is possible. The step-time dashboard card behaves the same way:
 unmeasured phases leave the ribbon, the window label adds `partial signals`,
-and an underivable residual shows `n/a`.
+and an underivable residual shows `n/a`. H2D is the exception on both
+surfaces: its events only occur when host-to-device copies happen, so an
+absent H2D means no observed transfers and never counts as partial
+coverage.
 
 ### Local UI
 
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -99,6 +99,13 @@ When launched with `--mode=cli`, the terminal shows live:
 Live CLI mode is intended for single-node runs, including single-node
 multi-GPU.
 
+Phases that were never measured in the current window are omitted from the
+live step-time table rather than shown as `0.0 ms`, and the diagnosis block
+shows `INCOMPLETE DATA` with the missing signal names when no reliable
+conclusion is possible. The step-time dashboard card behaves the same way:
+unmeasured phases leave the ribbon, the window label adds `partial signals`,
+and an underivable residual shows `n/a`.
+
 ### Local UI
 
 The local UI shows the same ideas in a more compact review format:
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -102,12 +102,15 @@ multi-GPU.
 Phases that were never measured in the current window are omitted from the
 live step-time table rather than shown as `0.0 ms`, and the diagnosis block
 shows `INCOMPLETE DATA` with the missing signal names when no reliable
-conclusion is possible. The step-time dashboard card behaves the same way:
-unmeasured phases leave the ribbon, the window label adds `partial signals`,
-and an underivable residual shows `n/a`. H2D is the exception on both
-surfaces: its events only occur when host-to-device copies happen, so an
-absent H2D means no observed transfers and never counts as partial
-coverage.
+conclusion is possible. The local UI's step-time hero card behaves the same
+way: an unmeasured phase renders as a hatched sliver on the ribbon, never as
+an empty (zero-width) segment, so a dark stream can never be mistaken for a
+confirmed-fast one; the window label adds `partial: <names>`, and an
+underivable residual shows `n/a`. If a run stops reporting entirely, both
+surfaces show the last window has expired instead of freezing on a dead
+run's numbers. H2D is the exception throughout: its events only occur when
+host-to-device copies happen, so an absent H2D means no observed transfers
+and never counts as partial coverage.
 
 ### Local UI
 
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -103,14 +103,20 @@ Phases that were never measured in the current window are omitted from the
 live step-time table rather than shown as `0.0 ms`, and the diagnosis block
 shows `INCOMPLETE DATA` with the missing signal names when no reliable
 conclusion is possible. The local UI's step-time hero card behaves the same
-way: an unmeasured phase renders as a hatched sliver on the ribbon, never as
-an empty (zero-width) segment, so a dark stream can never be mistaken for a
-confirmed-fast one; the window label adds `partial: <names>`, and an
-underivable residual shows `n/a`. If a run stops reporting entirely, both
-surfaces show the last window has expired instead of freezing on a dead
-run's numbers. H2D is the exception throughout: its events only occur when
-host-to-device copies happen, so an absent H2D means no observed transfers
-and never counts as partial coverage.
+way. The ribbon is one real representative rank from the common eligible
+cohort, named in the window label; it never combines phases from unrelated
+ranks. An unmeasured phase renders as a hatched sliver rather than a measured
+zero, while unavailable shares and cross-rank gaps render as `n/a`. If no rank
+has a coherent composition, the ribbon clears and independently valid KPIs
+continue updating. H2D is the exception throughout: its events only occur
+when host-to-device copies happen, so an absent H2D means no observed
+transfers and never counts as partial coverage.
+
+The live reader bridges short empty or failed reads using its last good
+window, then expires that bridge. Persisted SQLite rows remain a valid saved
+snapshot; an unchanged step number is not treated as proof that the producer
+stopped. Detecting producer liveness requires a separate heartbeat or source
+timestamp.
 
 ### Local UI
 
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_combined_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_combined_section.py
@@ -26,6 +26,11 @@ from traceml_ai.renderers.step_time.schema import (
 
 from . import theme
 
+# Fixed ribbon width for a non-h2d phase that was never measured this
+# window -- large enough to read as a deliberate marker, not a rounding
+# artifact of a real proportional segment.
+_UNMEASURED_SLIVER_PCT = 6.0
+
 
 def build_model_combined_section() -> Dict[str, Any]:
     seg_divs: List[Any] = []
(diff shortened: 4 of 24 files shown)


abhinavsriva (changes requested):
## Request changes

The sparse rendering direction is good, but three correctness cases still issue #259:

1. **Sparse → complete keeps the hatch.** NiceGUI merges styles, and the measured path updates only `width`, so a recovered phase remains visually marked as unmeasured. Restore the solid phase background and add a stateful transition test.

2. **Missing `step_time` freezes the previous view.** The early return at `model_combined_section.py:143` leaves the old ribbon and KPIs visible and prevents unrelated valid observations from updating. Clear the phase view or render an explicit unavailable-envelope state.

3. **Partial-rank coverage appears complete.** Aggregate metric presence does not mean every rank measured it. With Backward available on rank 0 but missing on rank 1, the canonical diagnosis reports `INCOMPLETE DATA`, while the dashboard label still says only `30 aligned steps`. Use `per_rank_timing` or canonical availability evidence.

One additional cleanup: stopping a producer does not currently trigger the advertised dead-run expiry because persisted SQLite rows continue producing successful results and reset `_last_ok_ts`. Please either add real source-freshness detection or remove/narrow the dead-run behavior and documentation in this PR.

The 423 passing tests do not cover sparse → complete, complete → missing envelope, or asymmetric rank coverage. Please add those cases with stateful NiceGUI-style merging.

Pendu:
All four are real. Fixed in `db62da8`, CI green. Each fix is pinned by a test that fails when the fix is reverted (verified by reverting the two source files and re-running).

**1. Sparse to complete kept the hatch.** Confirmed: `Element.style()` merges declarations, and the measured path only set `width`, so a recovered phase kept the hatched background indefinitely. The measured path now restates `background:{col}` on every update.

The test could not have caught this. The `_FakeSegment` double appended style strings to a list and assertions read the last call, so a surviving earlier declaration was unobservable. The double now merges into a dict like the real widget, exposes `background`/`width`/`hatched`, and is pinned against a real `ui.element` by `test_fake_segment_matches_nicegui_style_merge`. Added `test_step_time_dashboard_hero_recovered_phase_drops_the_hatch` (sparse, then complete, asserts the hatch is gone and the solid phase color is back).

**2. Missing `step_time` froze the previous view.** Confirmed, and the old test asserted that behavior on a fresh panel, so it passed while proving nothing. Without the envelope there is no denominator, so the card now clears the ribbon and KPIs and shows `step envelope unavailable`. Added `test_step_time_dashboard_hero_clears_when_step_envelope_missing`, which draws a complete window first and asserts the clear happens on the second update.

**3. Partial-rank coverage appeared complete.** Confirmed. The card read aggregate metric presence; the engine requires a signal on every observed rank. Coverage is now re-derived from `per_rank_timing` via `_partially_covered`, and rank-thin phases are named in the window label alongside fully-unmeasured ones. Added `test_step_time_dashboard_hero_flags_partial_rank_coverage` (backward on rank 0 only reports `partial: BWD`) with a symmetric control twin.

While fixing this I also imported the engine's occurrence-metric set instead of the local hardcoded `h2d` check. `optimizer_step` is occurrence-driven too, so an absent optimizer step no longer counts as a coverage gap, which matches `OCCURRENCE_METRICS`. That set is now exported from `step_time_window` so the live surfaces and the engine share one definition rather than drifting.

**4. Dead-run expiry never fired.** Confirmed, and this is the one I should have caught: I saw a live run stay on screen after I killed the producer and wrote it off as a test-environment artifact instead of the defect it was. Rows outlive the process that wrote them, so a stopped run recomputes the same window successfully forever and resets the stale timer every tick.

Added real source-freshness detection: the computer tracks the highest `completed_step` it has seen and the time it last advanced. A successful recompute no longer counts as liveness; only an advancing cursor does. Within the TTL the view is still served, past it the window expires, and it recovers if the producer comes back. `test_dead_run_expires_even_though_sqlite_rows_persist` drives that whole sequence with a frozen clock and fails without the fix.

**Coverage.** Suite is 679 passing (was 675), same 20 pre-existing failures as the base commit. The three cases you named are covered: sparse to complete, complete to missing envelope, and asymmetric rank coverage.

**Process change.** Three of these were transition bugs and every test I had rendered a single state, so nothing could catch them. I have added two test classes to the internal hardening checklist, stateful transitions and test-double fidelity, and the live-UI row now requires rank-asymmetric coverage even when a change touches no instrumentation.

abhinavsriva (changes requested):
## Request changes  3 items, all reproduced on `db62da8`

  **1. Absent `optimizer_step` renders as a measured zero **

  Line 208 swapped `key != "h2d"` for `key not in OCCURRENCE_METRICS`. That set governs presence in the window layer (seen once → zero-fill the gaps); it was never a
  rule about total absence. `_rank_metric_availability` drops a never-observed optimizer
  entirely, and `_missing_signal_report` exempts only h2d. Revert the predicate to h2d-only.

  **2. Residual Share shown when `input_wait` is unavailable**

  ```
  input_wait in window -> False
  residual_proxy       -> 10.0
  RESIDUAL SHARE KPI   -> 10%
  ```

  `residual_proxy` needs step_time + compute phases, not input_wait, so it survives while the
  denominator doesn't. `envelope = step_time + (input_wait or 0.0)` (line 214) substitutes zero,
  so the KPI reads a confident 10% against a denominator that's short by an unknown input wait.
  Should be `n/a` — the branch already exists at line 288, it just needs the guard.

  Reachable only because this PR removed the all-metrics gate.

  **3. Per-rank coverage omits `step_time`**

  ```
  win label -> "30 aligned steps"
  canonical -> INCOMPLETE_DATA: step_time
  ```

  `_partially_covered` gets `[key for _, key, _ in theme.PHASES]`, which has no `step_time`  it's  the envelope, not a phase. Pre-existing on `version_0.3.6`, but in #259's sparse-coverage scope.

Pendu:
All three confirmed and fixed in `6fc6691`, rebased onto the current `version_0.3.6` (so #286 and #289 are in). Each fix is pinned by a test that fails when the source is reverted.

**1. Absent `optimizer_step` rendered as a measured zero.** Correct, and this one was mine: the previous round swapped `key != "h2d"` for `key not in OCCURRENCE_METRICS` on the assumption the set meant "absence is fine". It does not. Confirmed against the availability rule:

```
never-observed optimizer  -> available? False
intermittent optimizer    -> available? True
```

The set governs intermittent presence, since `_rank_metric_availability` only ever considers metrics it actually observed, so a never-seen optimizer is dropped before the exemption applies. Predicate is back to h2d-only, matching `_missing_signal_report`. The `OCCURRENCE_METRICS` export added for that change is reverted too, so `step_time_window.py` is untouched by this PR again. Pinned by `test_absent_optimizer_step_is_not_a_measured_zero`.

**2. Residual Share shown when `input_wait` is unavailable.** Reproduced exactly as described: `input_wait` absent, `residual_proxy` 10.0, KPI reading a confident `10%`. The residual survives because it derives from step time and the compute phases, but the envelope substitutes zero for the missing wait, so the denominator is short by an unknown amount. The KPI is now gated on the denominator being whole, not just on the numerator existing. Pinned by `test_residual_share_is_na_when_input_wait_is_unavailable`, with a control twin asserting a real share still renders when `input_wait` is measured.

**3. Per-rank coverage omits `step_time`.** Correct. `_partially_covered` was handed `theme.PHASES`, which is the ribbon phase list and has no `step_time` in it, so a rank missing the envelope was invisible:

```
_partially_covered(PHASES)            -> []
_partially_covered(PHASES+step_time)  -> ['step_time']
```

`step_time` is now checked alongside the phases and named `STEP` in the window label, matching the CLI table's abbreviation. Pinned by `test_partial_rank_coverage_includes_step_time`.

**Coverage.** 687 passing against 667 on the current base, same 20 pre-existing failures, so 20 net-new tests and no regressions. All three new tests fail with the two source files reverted.

On the pattern in these two rounds: the first round was single-render state tests missing transitions, and this round was me widening a rule past what it actually meant. Both are cases of asserting the behavior I intended rather than the behavior the engine defines. The per-rank checks now derive from `per_rank_timing` and mirror `_missing_signal_report` rather than restating it, which is the part that kept drifting.

abhinavsriva (changes requested):
I’m limiting this review to the three correctness problems introduced or exposed by #259.

The sparse-metric direction is sound, and this does not require rewriting #257 or #258. Small additive changes to the shared Step Time models are explicitly in scope where the producer already has the canonical information. 

The goal is to carry that information forward instead of reconstructing it independently in each view.

### 1. Carry metric availability and a coherent composition cohort from the producer

`step_time_window.py` already computes `metric_ranks = sorted(values)` for every metric, but discards it. Please preserve the measured-rank population on the shared Step Time result/metric model and have the dashboard consume it. `_partially_covered` should not maintain a second interpretation of availability in the view.

The ribbon also must not combine aggregates calculated from disjoint rank populations. For example, forward measured only on rank 0 and backward only on rank 1 currently produces a composition that no rank measured. Build ribbon values from one coherent rank population.

If no coherent population exists, do not draw a proportional composition. Clear only the composition visualization, state why it is unavailable, and retain any independently valid KPI/text observations. `_clear_view` contains the existing ribbon-reset behavior, but it should not be called wholesale here because it also blanks every valid KPI.

`StepCombinedTimeSummary.skew_pct` is currently declared as `float` in `renderers/step_time/schema.py`. Preserving “cross-rank gap is uncomputable” therefore requires widening it to `Optional[float]`. That additive shared model change is in scope. When fewer than two ranks measured a metric, `step_time_window.py` should emit `None`, not force `0.0`, and all consumers should preserve that distinction.

### 2. Carry canonical component shares to the result consumed by renderers

The canonical `residual_share`, `compute_share`, and `input_bound_share` already exist on `StepTimeAnalysisContext`, but `StepCombinedTimeResult` does not carry them. Consequently, CLI, diagnostics evidence, and the dashboard reconstruct the shares independently and choose different denominators.

Please expose these canonical shares through `StepCombinedTimeResult`, or an embedded shared analysis object, and have consumers use them without re-deriving them.

Concrete reproduction:

- residual: 10 ms
- step envelope: 100 ms
- input wait: 100 ms
- canonical iteration share: `10 / (100 + 100) = 5.0%`
- diagnostics rail and Step Time CLI currently report `10.0%`

After this change:

- missing residual or gap is unavailable, not `0.0%`;
- measured zero remains `0.0%`;
- CLI, dashboard hero, and diagnostics evidence report the same canonical value.

These should be acceptance checks for the shared boundary, not additional view-specific guards.

### 3. Remove only the unsupported cursor-based expiry

Please remove the branch in `compute.py` that treats an unchanged `coverage.completed_step` for 30 seconds as evidence that the producer stopped.

`instrumentation.py:188-199` confirms why the cursor cannot establish liveness: `step_completed` is set only after the traced step body returns, and both `advance_step()` and `flush_step_events()` happen afterward. A legitimate long step, checkpoint, or barrier therefore leaves the cursor unchanged and can be falsely reported as an expired run.

Keep the existing last-good-result bridge for genuinely empty/transient computes. `test_computer_bridges_transients_then_expires` still represents useful behavior. The test that specifically blesses expiry while persistent SQLite rows remain readable should be removed. `had_ok` may remain if it continues to represent only the genuine empty-compute transition.

### Verification

Please add one stateful fixture that drives the same sparse/partial data through the shared Step Time result, CLI, diagnostics evidence, and dashboard.

Assert semantic outcomes—measured-rank populations, coherent cohort selection, canonical share values, and `None` versus measured `0.0`—rather than ribbon markup or formatted strings. Since these consumers require different extras, place the test in the integration job that installs `torch` and `dashboard` and runs the reporting, renderer, and display suites.

This keeps the solution modular: availability, cohort eligibility, and shares are established once upstream; consumers only render them. The separate cursor-liveness heuristic is removed because its input cannot support its conclusion.

abhinavsriva (approved):

PR #300, opened 2026-08-03
Add Hugging Face Accelerate example and guide

### LINKED ISSUE #190: [Feature]: Add a Hugging Face Accelerate example and guide
### What would you like TraceML to do?

Add a small, tested example showing how to use TraceML with a Hugging Face Accelerate training loop.

TraceML already supports Hugging Face Trainer through `TraceMLTrainerCallback`, but Accelerate users often write their own PyTorch loop with `Accelerator.prepare(...)` and `accelerator.backward(...)`. We should document that path clearly.

This should be an example/guide first, not a new callback API.

Suggested scope:

- Add `examples/integrations/accelerate_minimal.py`.
- Add `docs/user_guide/integrations/accelerate.md`.
- Link the guide from `docs/user_guide/integrations.md`.
- Link the example from `examples/README.md`.
- Keep `accelerate` optional; do not add it as a core dependency.
- Add a small import-skipped test or smoke test so CI does not require Accelerate by default.

The guide should also explain the launch path. For the first version, the documented path should be:
```
traceml run examples/integrations/accelerate_minimal.py --mode=summary
```
and for single-node multi-GPU:
```
traceml run examples/integrations/accelerate_minimal.py --mode=summary --nproc-per-node=2
```

Out of scope for this first issue:
- Making traceml run wrap accelerate launch.
- Supporting every Accelerate config mode.
- DeepSpeed/FSDP-through-Accelerate tuning.
- A new TraceMLAccelerateCallback API.

Acceptance criteria:
- The example runs on CPU or single GPU.
- The docs explain where traceml.init(...) and traceml.trace_step(...) go.
- The docs explain that TraceML should wrap the prepared model returned by accelerator.prepare(...).
- The docs mention that accelerate launch is not the first supported TraceML path; use traceml run for now.
- The docs call out gradient accumulation briefly, either with a simple note or by leaving it out of the minimal example.
- The final summary is produced successfully.

### Why would this help?

A lot of Hugging Face users use Accelerate when they want more control than `Trainer` gives them. Today TraceML has a clear Trainer path, but not a clear Accelerate path.

This would make TraceML easier to try for custom fine-tuning loops, small distributed experiments, and users who already have an Accelerate script.

Animesh452:
## What changed

- Adds `examples/integrations/accelerate_minimal.py`: a minimal Hugging Face
  `Accelerate` training loop instrumented with `traceml.trace_step(...)`.
- Adds `docs/user_guide/integrations/accelerate.md`: guide covering install,
  init/wrap ordering, launch, limitations, and troubleshooting — mirrors the
  structure of `integrations/huggingface.md`.
- Links the guide from `docs/user_guide/integrations.md` and the example
  from `examples/README.md`.
- Adds `tests/integrations/test_accelerate.py`: an import-skipped smoke test
  mirroring `test_deepspeed.py`'s pattern (fake `Accelerator` stand-in for
  unconditional coverage, one `skipif`-gated test against the real library).

No changes to `src/`. No new public API, no new core or optional
dependency — `accelerate` is already available via the existing `hf` extra.

## Why

Closes #190. TraceML has a documented path for Hugging Face `Trainer`
(`TraceMLTrainerCallback`), but no documented path for users who write their
own loop with `Accelerator.prepare(...)` / `accelerator.backward(...)`. This
adds that missing guide and example, as scoped in the issue — no new
callback API.

## How I tested

- [x] Tests added or updated
- [x] Existing tests run
- [x] Manual smoke test

```bash
traceml run examples/integrations/accelerate_minimal.py --mode=summary
pytest tests/integrations/test_accelerate.py -v
pre-commit run --files examples/integrations/accelerate_minimal.py \
  docs/user_guide/integrations/accelerate.md \
  docs/user_guide/integrations.md \
  examples/README.md \
  tests/integrations/test_accelerate.py
```

All pass. Also confirmed the smoke test degrades correctly without
`accelerate` installed: the two unconditional tests still exercise real
TraceML instrumentation via a fake `Accelerator` stand-in; only the test
against the real library skips.

Tested on WSL2/Ubuntu. For context: native Windows currently hits two
pre-existing, unrelated platform issues when running `traceml run`
end-to-end (`SO_REUSEPORT` — already fixed in #264 — and a separate
`torch.distributed.run`/libuv rendezvous issue that's still open). Neither
is related to this change; the example script itself was independently
verified correct via plain `python` execution, unaffected by either issue.

## Runtime impact

- [x] No training-path impact

Notes: Purely additive — example, docs, and test only. No changes to
tracing, runtime, telemetry, or distributed code paths.

## Docs

- [x] Updated

## Extra context

**Design decision worth flagging for review:** the guide documents
`accelerator.unwrap_model(model)` before passing to `trace_step()`, rather
than the raw object `accelerator.prepare()` returns. TraceML's step-memory
tracker keys off `id(model)` and reads device placement from
`model.parameters()`; under a distributed launch `prepare()` can wrap the
model (e.g. DDP), so the wrapped object isn't what actually runs
forward/backward. This follows the same `model.module`-unwrapping pattern
already used in `ddp_minimal.py` and `deepspeed_minimal.py`'s
`model_engine.module`.

Final summary output from a real run (WSL2, CPU):

```
+----------------------------------------------------------------------------+
|  TraceML Run Summary | duration 2.3s                                       |
+----------------------------------------------------------------------------+
|  TraceML Verdict: RESIDUAL-HEAVY / CRITICAL                                |
|  Why: Residual time is 25.2% of the typical cpu iteration time.            |
...
```

(Noted in the guide's Troubleshooting section as expected for this
minimal example — a millisecond-scale synthetic workload makes ordinary
loop overhead proportionally large, not a real regression signal.)

Unrelated observation, not addressed here: `tests/integrations/` doesn't
currently run in any CI job (verified via `grep -rn "tests/integrations"
.github/workflows/*.yml` — no matches). Pre-existing, affects
`test_hf_trainer.py`, `test_lightning.py`, `test_ray.py`, and
`test_deepspeed.py` identically, not specific to this PR.

## Summary by CodeRabbit

* **New Features**
  * Added Hugging Face Accelerate integration support for CPU, single-GPU, and multi-GPU training.
  * Added guidance for initializing TraceML, instrumenting training steps, and launching traced runs.
  * Added a runnable minimal Accelerate training example with per-step tracing.
* **Documentation**
  * Added the Accelerate integration guide, including setup, usage, limitations, troubleshooting, and examples.
  * Added Accelerate to the integrations and examples documentation.
* **Tests**
  * Added coverage for tracing, timing, data loading, and model handling with Accelerate.

### DIFF
--- a/docs/user_guide/integrations.md
+++ b/docs/user_guide/integrations.md
@@ -8,6 +8,7 @@ job already runs.
 |---|---|---|
 | Plain PyTorch loop | `traceml.trace_step(...)` | [Quickstart](quickstart.md) |
 | Hugging Face Trainer | `TraceMLTrainerCallback` | [Hugging Face](integrations/huggingface.md) |
+| Hugging Face Accelerate | `traceml.trace_step(...)` | [Hugging Face Accelerate](integrations/accelerate.md) |
 | PyTorch Lightning | `traceml_ai.integrations.lightning.init()` + `TraceMLCallback` | [PyTorch Lightning](integrations/lightning.md) |
 | Ray Train | `TraceMLTorchTrainer` | [Ray Train](integrations/ray.md) |
 | DeepSpeed | `traceml.trace_step(...)` | [DeepSpeed](integrations/deepspeed.md) |
--- a/docs/user_guide/integrations/accelerate.md
+++ b/docs/user_guide/integrations/accelerate.md
@@ -1,0 +1,170 @@
+# Hugging Face Accelerate Integration
+
+Use TraceML with a custom Hugging Face `Accelerate` training loop — for
+users who write their own loop with `Accelerator.prepare(...)` and
+`accelerator.backward(...)` instead of using `Trainer`.
+
+The integration is not a callback or a new API. It is the same
+`traceml.trace_step(...)` primitive used in any plain PyTorch loop, wrapped
+around the unwrapped model that `accelerator.prepare()` produces.
+
+## 1. Install
+
+```bash
+pip install "traceml-ai[hf]"
+```
+
+`accelerate` ships as part of the existing `hf` extra alongside
+`transformers` — there is no separate `accelerate` extra, and TraceML does
+not add `accelerate` as a core dependency.
+
+## 2. Initialize TraceML And Wrap The Prepared Model
+
+Create the `Accelerator` first, then call `traceml.init()`, then hand the
+model, optimizer, and dataloader to `accelerator.prepare(...)` before
+wrapping the *result* with `trace_step`:
+
+```python
+import traceml_ai as traceml
+from accelerate import Accelerator
+
+accelerator = Accelerator()
+traceml.init(mode="auto")
+
+model, optimizer, dataloader = accelerator.prepare(model, optimizer, dataloader)
+traced_model = accelerator.unwrap_model(model)
+
+for batch_x, batch_y in dataloader:
+    with traceml.trace_step(traced_model):
+        optimizer.zero_grad(set_to_none=True)
+        logits = model(batch_x)
+        loss = criterion(logits, batch_y)
+        accelerator.backward(loss)
+        optimizer.step()
+```
+
+`accelerator.unwrap_model(model)` matters even for a single-process run: it
+is what makes this code correct unchanged under multi-GPU too. TraceML's
+step-memory tracker keys instrumentation off `id(model)` and reads device
+placement from `model.parameters()`. If `accelerator.prepare()` wraps the
+model (which it does under distributed launch), passing the wrapped object
+into `trace_step` would hook an object that isn't the one actually running
+forward/backward. `unwrap_model()` is a no-op on CPU/single-GPU and strips
+the wrapper under distributed — so this line is what keeps the same code
+correct in both cases, not just a distributed-only concern.
+
+Note also what's *not* here: no `model.to(device)`, no `batch.to(device)`.
+`accelerator.prepare()` handles both — that's the actual value Accelerate
+adds over writing raw `DistributedDataParallel` by hand.
+
+## 3. Launch The Run
+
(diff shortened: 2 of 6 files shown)


Pendu:
Thanks for this, it is a strong contribution. The `[hf]` extra research is correct, the design reasoning for `unwrap_model()` matches how the DDP and DeepSpeed guides handle the same problem, and the troubleshooting section is honest about the example's own RESIDUAL-HEAVY verdict rather than hiding it.

I reviewed and tested it end to end, then pushed a few follow-ups to your branch (commit 4e85102):

- `float(loss.detach())` inside `trace_step` forced a per-step host sync on the training thread. Moved the loss accumulation to stay on-device (`torch.zeros((), device=accelerator.device)`, `+= loss.detach()`) and convert to a Python float only in the periodic print block, which already runs outside `trace_step`. On this example, removing the sync dropped the residual share from 38.6% to 27.2%. Added a short note to the troubleshooting section, worded so it does not overclaim about GPU-selected residual, since that comes from CUDA event timestamps rather than CPU wall time.
- `mkdocs build --strict` was failing: the example link in the guide escaped the docs directory. Switched it to inline code, matching how the DeepSpeed guide references its own example.
- The new page was not wired into `mkdocs.yml`, so it would not have appeared in the site nav even though the build itself succeeded. Added it next to Hugging Face, matching the order in your own overview table.
- Added the real-`Accelerator.prepare()` test CodeRabbit suggested, running the full recipe (real prepare, real dataloader, real backward) and asserting all five step-time streams land.

Verified on the pushed head: `pytest tests/integrations/` -> 25 passed, 1 skipped; `mkdocs build --strict` passes; `traceml run examples/integrations/accelerate_minimal.py --mode=summary` runs clean end to end; black and codespell clean.

Looks good to merge once the checks settle. Thanks again for the contribution.

PR #336, opened 2026-08-04
Make the Public API reference complete and behaviorally accurate

abhinavsriva:
## Summary

This PR makes the documented `traceml_ai` API match the runtime contract.
It documents every public export, corrects missing-aggregator behavior, and
adds tests that keep the API reference synchronized as the package evolves.

## What changed

- Rebuilt the Public API page around the top-level `traceml_ai` import path,
  covering all callable exports and `__version__`.
- Added complete public signatures and concise docstrings for the façade API:
  parameters, return values, side effects, duplicate-instrumentation errors,
  and no-op behavior.
- Corrected direct-launch behavior: an unreachable aggregator warns once and
  makes tracing a no-op by default; `raise` is opt-in through
  `on_missing_aggregator="raise"` or `TRACEML_ON_MISSING_AGGREGATOR=raise`.
- Made `traceml.start()` forward every `traceml.init()` option, including the
  missing-aggregator policy, so the existing alias has the same behavior.
- Separated stable core APIs from framework integrations and labelled
  `TraceMLTrainer` as legacy compatibility; `TraceMLTrainerCallback` remains
  the recommended Hugging Face integration.
- Added a docs-contract test derived from `traceml_ai.__all__`, plus focused
  runtime tests for missing-aggregator policy precedence and `start()` parity.

## Scope

This does not change instrumentation semantics, CLI behavior, aggregation, or
framework integrations. The only runtime API correction is that the existing
`traceml.start()` alias now accepts and forwards the same
`on_missing_aggregator` option as `traceml.init()`.

## Validation

```text
PYTHONDONTWRITEBYTECODE=1 pytest -p no:cacheprovider -q tests/sdk tests/docs/test_public_api_contract.py
```

Result: `52 passed`.

`git diff --check` also passes.

`python -m mkdocs build --strict` is currently blocked in this worktree by an
untracked `docs/README.md` that conflicts with `docs/index.md`; that file is
outside this PR's changes. Re-run the strict build in a clean checkout (or
after resolving that unrelated conflict) before merge.

### DIFF
--- a/docs/user_guide/public-api.md
+++ b/docs/user_guide/public-api.md
@@ -1,198 +1,241 @@
 # Public API
 
-The stable surface that user code imports and calls. Everything in this page is covered by TraceML's compatibility contract across v0.x minor releases.
-
-## Core API
+Import the stable core API from `traceml_ai`:
 
 ```python
 import traceml_ai as traceml
-
-traceml.init(mode="auto")
-
-with traceml.trace_step(model):
-    ...
 ```
 
-The old `import traceml` path remains available for compatibility, but emits a
-`FutureWarning` and may be removed in a future release. Do not import from
-decorator compatibility paths.
+The reference below documents every symbol in `traceml_ai.__all__`. The old
+`import traceml` path remains a compatibility import and emits a
+`FutureWarning`; new code should use `traceml_ai`.
 
-## Hugging Face integration
+## Stable Core API
 
-::: traceml_ai.integrations.huggingface.init
+### `traceml.__version__`
+
+A string identifying the installed TraceML version. It is useful when recording
+the environment for a run or bug report.
+
+### Lifecycle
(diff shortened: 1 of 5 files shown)


Pendu:
Pushed 5ec91fa with the drift check from the issue's acceptance criteria:
`test_public_api_page_documents_every_public_symbol` parses the
`::: traceml_ai.api.*` directives out of public-api.md and asserts
set-equality with `traceml_ai.__all__`, so it fails both when a new public
symbol lacks a page entry and when the page documents something that is no
longer public. Verified it trips on both mutations before pushing.

Everything else here checked out clean on my side: start/init signature
parity, the on_missing_aggregator resolution order (code, test, and docs
all agree), the per-helper Raises behavior, and `mkdocs build --strict`.

Issue #317, opened 2026-08-03
Step Time test fixtures hand-roll the SQLite schema instead of using the production projection writer


Pendu:
## Problem

The step_time read-side tests build their databases from hand-written DDL
rather than the production writer, so a writer-side schema change (column
rename, index change, events_json shape change) passes the whole suite green
and only breaks at runtime:

- `tests/step_time/scenarios.py:229` and `:245` hand-write
  `CREATE TABLE step_time_samples` / `runtime_environment`, and every
  repository/pipeline/contract-baseline test reads a database built from them.
- `tests/reporting/summary/test_step_time_nullable.py:60` and `:79` duplicate
  the same DDL a second time.
- The shared fixture that DOES call the production projection writers already
  exists (`tests/sqlite_fixtures.py`, wrapping
  `aggregator/sqlite_writers/step_time.py::init_schema` and friends) and is
  used by the reporting fixtures, but not by tests/step_time.
- The fixture `runtime_environment` table has 2 columns while the production
  table has many more (including `recv_ts_ns INTEGER NOT NULL`,
  `sqlite_writers/runtime_environment.py:102`), so the strategy read is only
  ever tested against a reduced table.
- The scenario databases create no indexes. The tail-scan cost test creates
  its own index named `step_time_rank_step_id`
  (`tests/step_time/test_sqlite_repository.py:265`) instead of the production
  `idx_step_time_samples_global_rank_step_id`
  (`sqlite_writers/step_time.py:214`), so nothing fails if the production
  index is dropped or renamed while the live tail query depends on it.

## Proposal

1. Route `scenarios.py` and `test_step_time_nullable.py` through
   `tests/sqlite_fixtures.py` (production `init_schema` + `build_rows`).
2. Add one test asserting the production index exists and is chosen by the
   live tail query (`EXPLAIN QUERY PLAN`).
3. Add a contract test pinning the writer's events_json encoding
   (`_normalize_events` output) against what the repository decoder expects,
   so the two cannot drift silently.

Pendu:
Scope extension from the 0.3.7 program review (tracking: #322).
Two items fold into this issue rather than standing alone, and they change
its acceptance criteria:

**1. Public-schema golden test (per-PR).** Once the production-writer fixture
exists, add a per-PR test: writer fixture -> real summary builder -> assert
the exact public key set AND a per-metric null-semantics table against a
committed golden. Per-metric matters: occurrence-driven metrics
[...]

PR #338, opened 2026-08-04
Test-suite hygiene batch: remove false coverage, restore sparse-UI tests, re-baseline HTML fixtures

### LINKED ISSUE #326: Test-suite hygiene batch: remove false coverage, restore deleted sparse-UI tests, collection preflight
Part of the 0.3.7 testing program (#322). First wave: everything
here is in-process, needs no design decisions, and makes every later test
PR's signal cleaner.

1. **Delete `tests/sdk/test_grad_accum.py`**: a leftover exploration script
   that runs Lightning training at module import (collection) time, prints,
   and asserts nothing. It costs runtime in every suite invocation and can
   fail collection; it pins no behavior.
2. **Fix or rename the manifest atomicity test**:
   `tests/runtime/test_launcher.py:759`
   (`test_run_manifest_write_and_update_are_atomic`) asserts merged content
   only; nothing tests atomicity (temp-file + rename, no torn manifest on a
   crash mid-write). Either implement the crash-injection check or rename to
   what it verifies; a test name that overclaims is negative documentation.
3. **Regenerate the HTML renderer fixtures at the current schema**:
   `tests/reporting/html/conftest.py:56` pins `schema_version: Any = 1.4`
   while the product emits 1.7, so every renderer test validates a
   three-versions-stale payload shape. Regenerate from the production-writer
   fixture path that #317 establishes (not a fresh hand-rolled copy), so
   this does not mint a fourth fixture lineage.
4. **Restore the CLI sparse-state renderer coverage** deleted in #302
   (`tests/renderers/test_step_time_sparse_ui.py`, 186 lines): port the
   assertions (partial-signal labels, n/a gating, absent-vs-zero rendering)
   onto the current renderer API.
5. **Collection preflight for the torch-free job**: add
   `pytest --collect-only` as the first step of the torch-free core job (or
   a pre-commit hook), so a torch import landing in a torch-free test file
   fails in seconds instead of at first execution (the #283 incident class).

Pendu:
Closes #326.

## What changed

1. **Deleted `tests/sdk/test_grad_accum.py`**: a leftover exploration
   script that ran real Lightning training at module import time (before
   any test executes), printed, and asserted nothing. Verified it added
   pure cost with zero pinned behavior.

2. **Restored `tests/renderers/test_step_time_sparse_ui.py`** (deleted in
   #302). Not a blind revert: I re-verified each of the four scenarios
   (omit-missing-phase + INCOMPLETE_DATA, dead-view-on-expiry, calm
   startup panel, multi-rank sparse table) against the current
   `StepTimeRenderer` and `tests/step_time/factories` API by rendering
   each payload and reading the real output before writing assertions.
   All four pass unmodified in substance from the pre-#302 version; only
   an unused `total_step` fixture key was dropped.

3. **Added `tests/runtime/test_atomic_io.py`** and renamed the misleading
   `test_run_manifest_write_and_update_are_atomic` to
   `..._merge_correctly`. The old test only checked the merged JSON
   content, never a crash mid-write, so it asserted nothing about
   atomicity despite the name. The new tests exercise
   `write_json_atomic`/`write_text_atomic` directly: a failure during the
   `os.replace` step must leave the target path exactly as it was and
   must not strand a temp file. Verified these tests actually bind by
   mutation: removing the temp-file cleanup on the failure path flips 3
   of 5 new tests to failing.

4. **Re-baselined `tests/reporting/html/conftest.py`** from schema 1.4 to
   1.7, and added the 7 metric names introduced in 1.6/1.7
   (`input_wait_ms`, `step_time_ms`, `h2d_ms`, `residual_ms`,
   `forward_ms`, `backward_ms`, `optimizer_ms`) to the default step_time
   fixture, set to `None` rather than a fabricated `0.0`. Every test that
   cares about a specific schema version or metric set already overrides
   the default explicitly; only one test relied on the bare default
   (`test_render.py::test_render_returns_parseable_html_document` and
   siblings), and none assert on exact metric counts. Verified the
   renderer handles the new null values without crashing and without
   leaking a raw `None` into the HTML output.

## Not implemented: the `--collect-only` preflight

The issue's fifth item proposed a `pytest --collect-only` step in the
torch-free core job to catch a stray `import torch` before it fails at
run time (the #283 incident class). I tested this empirically before
adding it: a module-level `import torch` in a file under the core job's
test paths already fails collection in 0.46s, before any test in the
job runs (`ERROR collecting ... ModuleNotFoundError: No module named
'torch'`, `Interrupted: 1 error during collection`). A separate
`--collect-only` step would repeat the exact same check pytest already
performs as the first phase of the existing `Run core tests` step, with
no new coverage and no meaningful speed difference. The actual #283
gap was running the wrong job's environment locally before pushing,
which is already the documented discipline. Not adding a step that
would be pure redundancy.

## Verification

- `pytest tests/config tests/core tests/step_time tests/diagnostics tests/reporting/test_view.py tests/telemetry -q`: 395 passed.
- `pytest tests/aggregator tests/integrations tests/runtime tests/sdk tests/reporting/compare tests/reporting/html tests/reporting/summary tests/renderers tests/display tests/test_h2d_timing.py -q`: 518 passed, 2 skipped (pre-existing).
- `black==24.10.0 --check` and `ruff check` on all touched files: clean.

GATE: full core+integration suites green (395+518), atomic_io tests mutation-verified, sparse-UI tests re-verified against live rendered output rather than replayed from the deleted version
TRANSITIONS: n/a (test-only changes, no runtime state machines touched)
RANKS: n/a (no telemetry or multi-rank code touched)
PLATFORM: n/a (no platform-conditional code touched)
TRIPWIRE: if a future `__all__`-style production change needs the HTML conftest's remaining synthetic (non-production-writer) fixture path, route it through #317's fixture unification instead of hand-editing this file again


abhinavsriva:
Follow-up fixes after the original branch commit:

- Removed the duplicate `tests/renderers/test_step_time_sparse_ui.py`. #302 had moved its four sparse-UI scenarios into `tests/renderers/test_step_time_renderer.py`; coverage was never lost.
- Corrected the HTML default fixture to derive the complete current 14-field schema-1.7 Step Time contract from `STEP_TIME_METRIC_NAMES`. It now uses canonical `step_time_ms` and `input_wait_ms`, removes retired `total_step_ms`/`dataloader_ms`, preserves nullable unmeasured fields, and asserts rendered HTML contains no raw `None`.
- Fixed atomic-I/O temp cleanup for failures before `os.replace`, including `fsync`: the temp path is now tracked immediately after creation, and a regression test confirms the existing target remains intact with no stranded temp file.

Verification:

- `pytest tests/reporting/html -q`: 65 passed
- `pytest tests/runtime -q`: 140 passed
- Full integration suite: 515 passed, 2 skipped

abhinavsriva (approved):

Issue #323, opened 2026-08-04
Examples validation harness: scripts/validate_examples.py + manifest + CI tiers


Pendu:
Part of the 0.3.7 testing program (#322). Prerequisite: #324
(step knobs). Absorbs the CI-execution side of #319.

## Problem

The 17 examples are the de-facto integration suite: they are the only runnable
artifacts that exercise the full launcher -> aggregator -> training pipeline
the way users do. No CI job runs any of them. During the 0.3.6 pre-merge
validation, a scripted sweep of the CPU-runnable examples caught more product
surface per minute than any test directory: verdict flips, per-rank rows,
null-vs-zero semantics, manifest lifecycle, stderr cleanliness.

## Proposal

1. `scripts/validate_examples.py`: runs manifest-listed examples sequentially
   under the real launcher, one TRACEML_LOGS_DIR per example, and asserts per
   run: exit 0; final_summary.json present with the current schema version;
   run manifest `completed`; `h2d_ms` null (never 0.0) on CPU; the expected
   diagnosis assertion (see tiers below); no `[TraceML]`-prefixed
   error/traceback lines in stderr.
2. A manifest (checked in, single source of truth) where each entry declares:
   script, args, hardware (cpu/gpu/multi-gpu), timeout (~300s cap), and an
   assertion tier:
   - `structural`: rc/schema/manifest/nulls only (default for anything
     timing-sensitive on shared runners),
   - `family`: verdict must be in an allowlist (only where an injected knob
     gives a large signal margin, e.g. the bottleneck demo's slow scenario),
   - `exact`: reserved for structurally-forced outcomes (e.g. a
     no-instrumentation example must produce INCOMPLETE_DATA).
3. CI wiring per the umbrella tier model: CPU subset on PRs targeting
   version_* branches (paths-filtered) + nightly full sweep; NOT per-PR
   (except the #319 DDP entry, which is this harness invoked on one manifest
   entry). Job hardening: OMP_NUM_THREADS=1, job timeout-minutes, file-based
   or port-0 rendezvous for the DDP entry, artifacts (logs dir + summary +
   stdout/stderr) uploaded on failure, one-line contributor-legible failure
   messages (example, assertion class, expected vs actual), and a post-run
   no-orphan-process check.
4. The examples/README expected-verdict table is GENERATED from the manifest
   (or CI asserts they agree); two hand-maintained copies of the same truth
   is how fixtures drifted before.

## Notes

[...]

Issue #320, opened 2026-08-03
Integration tests never start the runtime: add a real-runtime subset, a conformance negative control, and a summary-protocol round-trip


Pendu:
## Problem

`tests/integrations/conftest.py:18` monkeypatches `_start_runtime_for_init`
for the whole directory, so every `init()` in the integrations suite installs
patches but starts no runtime: no sampler thread, no exporter, no aggregator.
Consequences:

- `test_telemetry_conformance.py` asserts against the in-process step-time
  queue. Everything downstream of that queue (sampler tick, msgpack, TCP,
  aggregator, SQLite) can break while the conformance gate stays green.
- The conformance test's negative control (prove the harness detects a dark
  stream) is explicitly deferred in its own docstring, so a harness bug that
  drains nothing reports success for non-required streams.
- `lightning` is declared in the required-streams contract but its
  parametrization always skips (no harness), so the declaration is not
  enforced.
- The mid-run summary protocol (request file in `control/`, service poll,
  response, client return) has no test that drives the full round-trip;
  only the timeout path and a direct `_handle_request` call are covered.

## Proposal

1. Add a `real_runtime` fixture that starts a real aggregator on port 0 and
   does not stub `_start_runtime_for_init`; move a small marked subset of the
   HF conformance assertions from queue-drain to reading the aggregator's
   SQLite output.
2. Add the negative control: break one patch deliberately and assert the
   conformance test fails for that stream.
3. Add a Lightning harness (or drop lightning from the required contract
   until one exists, so the skip is explicit).
4. Add one component test driving the real summary request/response
   round-trip against a live service, asserting the returned card contains
   real values.

Issue #141, opened 2026-06-02
Harden instrumentation-state: make integration telemetry completeness self-verifying


Pendu:
**Status:** Stage 1 PR is open: #151 (rebased onto current `main`; additive, `359 passed / 2 skipped`). Self-assigned to @Pendu.

## Summary
TraceML installs its instrumentation via **process-global, mutable state** — sentinels like `nn.Module._traceml_forward_patched`, `Tensor._traceml_h2d_patched`, `Optimizer._traceml_opt_hooks_installed`, and the `DataLoader.__iter__` swap — set by **three overlapping paths** (the CLI launcher's `init()`, an explicit `init()`, and a legacy import-time side-effect) and **reset/owned by none**.

Two consequences fall out:
- An integration can be individually correct yet silently emit **incomplete telemetry** (a stream is dark, and nothing says so).
- Tests can pass **for the wrong reasons** (global state leaks across them).

This surfaced during the callback-based HF integration (#88 / #135): the callback path didn't *guarantee* DataLoader-fetch timing — it only worked *incidentally* because importing the integration triggers the legacy auto-init side-effect. The gap was caught by review, not by the system. We'd like completeness to be a **structural, self-verifying property**.

## Evidence (reproducible)
**A — import-time install is incidental, not guaranteed.** Pristine process, no `init()` call, no prior tests — importing the integration alone initializes the process and patches the DataLoader:
```bash
PYTHONPATH=src python3 -c "
import torch.utils.data as tud; before = tud.DataLoader.__iter__
from traceml_ai.integrations.huggingface import TraceMLTrainerCallback
from traceml_ai.sdk.initial import is_initialized
print('is_initialized:', is_initialized())                              # True
print('DataLoader patched by import:', before is not tud.DataLoader.__iter__)  # True
"
```
**B — global state never reset → order-dependent test failure** (reproduces on `main`):
```bash
PYTHONPATH=src python3 -m pytest \
  tests/integrations/test_hf_trainer.py \
  "tests/sdk/test_init_and_wrappers.py::test_wrap_optimizer_preserves_identity_and_times_step" -q
# -> wrap_optimizer test FAILS: "global optimizer step hooks are already installed"
```
Two symptoms, **one root cause** (ungoverned global instrumentation state): (A) a *missing-install* masked by an incidental side-effect; (B) a *never-uninstall* that poisons sibling tests.

## Why feature-level tests don't catch this
[...]

PR #344, opened 2026-08-06
 Publish an evidence-backed integration support matrix

abhinavsriva:
## Summary

Add a generated support matrix for all seven integration paths. It clearly separates current CI evidence from documented or experimental recipes.

## Changes

- Add a machine-readable manifest and generated docs table.
- Mark only real PyTorch, Hugging Face Trainer, and Accelerate CPU coverage as  `CI tested`; label unproven paths conservatively.
- Link integration limitations to the relevant guides and #141, #320, and  #323.
- Add a Docs CI check that fails if a CI-tested claim loses its cited extra,  example, test, or generated table entry.

## Validation

- `python tools/integration_support_matrix.py --check`
- `pytest -p no:cacheprovider tests/integrations/test_integration_support_matrix.py -q`
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Added an integration support matrix covering supported frameworks, capabilities, validation status, examples, and known limitations.
  * Updated Hugging Face integration guidance to use the current initialization and callback approach.
  * Added validation guidance for keeping integration documentation accurate.

* **Quality Improvements**
  * Added automated checks for documentation coverage and integration support metadata.
  * Added scheduled notebook smoke tests and documentation validation in CI.


Pendu:
Reviewed and tested this. The approach is right and the shipped state is correct: `--check` passes, the generated table is in sync with the manifest, and the docs job now fails if either drifts.

The part I checked hardest is whether the matrix over-claims anywhere, since that is the whole point of calling it evidence-backed. It does not. Every GPU cell is `Not claimed`, multi-node is `Experimental`, and `CI tested` appears only on CPU scopes, which is exactly what a CPU-only `ubuntu-latest` job can support. The `test-integration` job really does run all three cited tests through `tests/integrations` and `tests/runtime` with `.[torch,dashboard,hf]` installed, so the three `CI tested` rows are backed today. The "Evidence limits" note about Lightning, Ray, and DeepSpeed having mocked or skipped paths is the right call.

Verified: `python tools/integration_support_matrix.py --check` passes; `pytest tests/integrations tests/docs` -> 40 passed, 2 skipped; `mkdocs build --strict` exits 0; ruff, black, and codespell clean at the pinned versions. The narrow `docs/data/` gitignore exception works, the manifest is tracked, and a hypothetical `docs/data/other.json` is still ignored.

Three things worth hardening. None of them is wrong in the committed state, and all fail loudly rather than silently, so they do not block:

1. `coverage[*].status` is unvalidated free text (`tools/integration_support_matrix.py:132-142`). The manifest currently uses six distinct values: `CI tested`, `Documented recipe`, `Experimental`, `Not applicable`, `Not claimed`, `Unsupported`. None is checked against an allowed set, and `is_ci_claim` matches the exact string `"CI tested"`. A typo like `"ci tested"` would silently skip the `ci_evidence` requirement and still render a row that reads as CI tested. This is the one I would fix first, since it is the tool's core guarantee.

   Note on the bot's suggestion here: it proposes normalizing all values to the `validation_levels` casing, which is not right. `Not claimed`, `Not applicable`, and `Unsupported` are coverage states, not validation levels. The fix is a separate allowed coverage-status set, validated before the `is_ci_claim` check.

2. `limitations` sub-keys are never validated (`tools/integration_support_matrix.py:183-188` and the `REQUIRED_FIELDS` check at `:113`). Only the top-level key is required, so a manifest with `"limitations": {}` passes `validate_manifest` and then raises `KeyError('guide')` inside `render_matrix`. Reproduced by setting `limitations` to an empty dict on the first entry. Worth validating `guide`, `guide_label`, and `issues` alongside the existing guide-path check so a manifest edit gets the tool's own error message instead of a traceback.

3. The cited test is checked for existence, not for execution (`:154-157`). Accurate today, but if the pytest invocation in `ci.yml` ever drops `tests/integrations` while the file stays on disk, the row keeps its `CI tested` label with nothing behind it. Since the tool exists to catch exactly that drift, checking that the cited path or a containing directory appears in the job body would close it. It has to accept a parent directory, because the job lists directories rather than individual files.

Two smaller notes:

- The regex workflow parsing (`:67-84`) is brittle in the way the bot describes, but both failure modes are fail-closed: `ci_job_body` raises on an indentation change and `installed_extras` returns an empty set, which then reports missing extras. For a gate, a false alarm is the safe direction, so this is lower priority. `yaml.safe_load` would still be a cleaner read.
- The bot's B904 and RUF043 comments do not apply here. This repo's ruff config sets only `line-length` with no `select`, so the default rule set is active and neither rule is in it. `ruff check` at the pinned version passes clean on all three new files, so no change is needed for those two.

The `persist-credentials` comment is on `notebook-smoke.yml`, which came in with #343 and is already merged, so it is outside this PR's diff. Reasonable hardening, but it belongs in its own change.

Looks good to merge once you have decided on the three items above.

Pendu:
Pushed the three hardening items to this branch (commit `9399ba3`), since you were happy for me to take them.

1. Added a `COVERAGE_STATUSES` set and validated every `coverage[*].status` against it. It is deliberately separate from `VALIDATION_LEVELS`, because `Not claimed`, `Not applicable`, and `Unsupported` are coverage states rather than validation levels. A near-miss spelling like `"ci tested"` now fails with a clear error instead of silently bypassing the `ci_evidence` requirement.

2. Validated the `limitations` object: `guide`, `guide_label`, and `issues` are now required, the guide path is checked the same way as `entry["guide"]`, and `issues` must be a list. `"limitations": {}` previously passed validation and then raised `KeyError('guide')` while rendering.

3. `ci_evidence.test` is now checked for execution, not just existence. The cited path or a parent directory of it must appear in the job's commands, so dropping `tests/integrations` from the pytest invocation while the file stays on disk now fails the check.

Two details worth knowing about item 3, both found while testing it rather than by reading:

- Matching has to be token-aware. A plain substring check on the bare parent `tests` matched inside `tests/aggregator`, so an unrelated suite counted as evidence.
- The search has to skip YAML keys and comments. The step name `- name: Run integration tests` and the comment mentioning `timing tests` both satisfied a bare `tests` candidate. It now searches only the shell command lines.

Also added four tests: the happy path (committed manifest validates and the generated table matches `integrations.md`) plus one per new guard. Each was confirmed to fail without its fix.

Verified on the pushed head: `python tools/integration_support_matrix.py --check` passes; `pytest tests/integrations tests/docs` -> 44 passed, 2 skipped; `mkdocs build --strict` exits 0; ruff, black, and codespell clean at the pinned versions.

I left the two smaller notes alone: the regex workflow parsing still fails closed, and the B904 and RUF043 suggestions do not apply under this repo's ruff configuration.

PR #345, opened 2026-08-06
Redesign terminal run/watch summary cards (#342)

### LINKED ISSUE #342: Redesign terminal summaries for fast, trustworthy decisions
# Design: define the right terminal summary for `traceml run` and `traceml watch`

## Problem

TraceML now produces substantial diagnostic information, but the terminal is the first surface most users see. We need to decide what the terminal should show by default so that it is immediately useful without becoming a dense telemetry dashboard.

This is a design and discovery issue. It should establish the presentation policy before implementation begins; it should not assume that the current summary layout, or any proposed replacement, is the final answer.

## Why this needs a decision

TraceML has two modes with different information and different user promises:

- `traceml run` instruments training steps and can diagnose likely performance
  bottlenecks.
- `traceml watch` is lightweight System and Process monitoring. It has no
  step-time attribution and must not imply that it does.

The current final-summary JSON retains four detailed domains: System, Process, Step Time, and Step Memory. The current terminal text shows a promoted performance verdict, a status table for all four domains, and System and Step Time evidence tables. The detailed section cards and HTML report retain more information.

The open question is not whether TraceML should collect or preserve this information. It is how much of it belongs in the default terminal card.

### Available data

The final-summary payload already provides:

- `primary_diagnosis`: `status`, `severity`, `summary`, `action`, and typed
  evidence for the run-level performance conclusion.
- Section-local `diagnosis` and ordered `issues` for `system`, `process`,
  `step_time`, and `step_memory`.
- Section `global` aggregates, grouped rows, units, and human-readable
  section `card` text.
- Full JSON and optional HTML artifacts for detailed inspection.

Important semantic constraints:

- A `null` metric is unmeasured, not zero; a terminal renderer cannot invent values or treat missing timing as 0 ms.
- GPU utilization is a symptom. It is not sufficient proof of the cause o  slow training.
- Step Time is currently the only direct root-cause input to  `primary_diagnosis`; low/moderate GPU utilization is a fallback only when Step Time is balanced.
- Process and Step Memory findings can be important (for example, GPU-memory
  pressure, high RSS, or memory creep), but are not currently promoted to the
  performance verdict.

### Mode constraints

- `run` can use System, Process, Step Time, and Step Memory evidence.
- `watch` can use System and Process evidence only. It should not show  Step Time or Step Memory placeholders, including their no-data states.
- The live Watch surface and the end-of-run summary do not need to make the  same claim or use the same headline vocabulary.

### Existing source of truth

- Final-summary schema and metric contract:
  `src/traceml_ai/reporting/SCHEMA.md`
- Primary performance-diagnosis policy:
  `src/traceml_ai/reporting/primary_diagnosis.py`
- Current final text renderer:
  `src/traceml_ai/reporting/final.py`
- Watch versus Run CLI profiles:
  `src/traceml_ai/launcher/cli.py` and
  `src/traceml_ai/aggregator/display_drivers/cli.py`
- User-facing output expectations:
  `docs/user_guide/reading-output.md`

### 1. What is the single job of each terminal surface?

Define the first question each mode must answer within a few seconds.

- For `run`, is the job strictly “why was training slow?”, or does it also
  need to surface run-safety and scaling risks?
- For `watch`, should the job be “is the host/process healthy?”, “is the GPU
  being used?”, or both?
- What must a user be able to decide before opening JSON, HTML, or a profiler?

### 2. What information earns a place in the default card?

Choose an explicit budget for the default terminal output:

- maximum number of lines or visual blocks;
- maximum number of metrics;
- whether healthy metrics/statuses are shown at all;
- whether each mode needs a compact topology/coverage indicator;
- where the detailed artifact path should appear.

Evaluate the trade-off between:

- showing only the strongest conclusion;
- showing a small set of corroborating metrics;
- showing every section’s status; and
- showing only exceptional non-primary findings.

### 3. How should performance diagnosis and health warnings coexist?

Decide whether `run` needs a bounded secondary-warning concept separate from the performance verdict.

Questions:

- Should a Step Memory creep, GPU-memory pressure, high host memory, high  process RSS, or high temperature appear in the terminal if it is not the  reason for slow steps?
- Which severities qualify?
- How many findings may appear before the terminal becomes a list again?
- How should ordering work when performance and reliability/scalability risks  conflict?
- What language prevents a secondary resource warning from being mistaken for a root cause?

### 4. What should Watch say about low GPU utilization?

Watch can observe that a GPU is underutilized but cannot attribute the idle time to input, compute, H2D, synchronization, or user code.

Decide:

- whether low GPU utilization should be a Watch headline, a supporting metric,  or only a warning;
- the wording that makes its uncertainty clear;
- when the output should recommend `traceml run`;
- whether high CPU is similarly an observation or a Watch health finding.

### 5. Which evidence is enough for trust?

For each possible headline, identify the smallest concrete evidence that makes the statement credible:

| Finding type | Candidate evidence to evaluate |
| --- | --- |
| Step-time bottleneck | dominant phase time/share or rank comparison |
| Low GPU utilization | GPU utilization, with an explicit attribution boundary |
| Memory pressure | peak memory relative to known device capacity |
| Memory creep | trend/window evidence, not just a latest value |
| Process pressure | process RSS/CPU/GPU memory with rank context when applicable |

The design must avoid displaying redundant measurements merely because they are available in JSON.

### 6. How should unavailable and incomplete data look?

Specify terminal behavior for:

- warmup / insufficient Step Time data;
- incomplete timing instrumentation;
- no GPU visible;
- partial multi-node or multi-rank coverage;
- absent optional metrics;
- Watch runs, where Step Time is intentionally not collected.

The output must distinguish “not measured,” “not enough data,” and “measured zero” without exposing implementation jargon by default.

## Design artifacts requested

Before coding, produce:

1. A concise output contract for `run` and `watch`.
2. A line-budgeted terminal mockup for each important state:
   - healthy/expected `run`;
   - Step Time performance finding;
   - a performance finding plus a serious resource warning;
   - low-GPU-utilization fallback;
   - healthy `watch`;
   - low-GPU-utilization `watch`;
   - a System or Process pressure `watch`;
   - no-data/incomplete-data states.
3. A mapping from each displayed element to its existing JSON source field.
4. A statement of what remains available only in JSON/HTML and why.
5. Proposed user-facing wording that clearly separates observation, likely
   bottleneck, and next action.

## Decision criteria

The chosen design should:

- let a new user understand the next action without reading a table;
- let an experienced user assess whether the conclusion is evidence-backed;
- avoid claiming causality from System/Process telemetry alone;
- ensure serious memory/resource risks are not silently hidden;
- keep `watch` honest about its lack of Step Time attribution;
- preserve the full JSON/HTML report as the detailed and machine-readable
  source of truth;
- work for single-GPU, multi-GPU, and multi-node summaries;
- preserve existing availability/null semantics;
- be simple enough to lock down with stable golden-output tests.

## Non-goals

- Changing telemetry collection, diagnostic thresholds, or Step Time rank  alignment.
- Replacing the structured JSON schema.
- Making all health/resource findings compete with a proven Step Time performance diagnosis.
- Recreating a full dashboard in a terminal card.

## Acceptance criteria for this design issue

- [ ] Agrees on distinct, documented purposes for `run` and `watch`.
- [ ] Agrees on a default information and line budget for each mode.
- [ ] Decides how exceptional System, Process, and Step Memory    findings are surfaced without overloading the performance verdict.
- [ ] Documents how Watch communicates low GPU utilization and it     attribution limits.
- [ ] Provides mockups and field mappings for all requested states.
- [ ] The approved design identifies the presentation-only implementation    boundary and the golden tests needed for it.

Pendu:
Closes #342.

Presentation-only redesign of the end-of-run terminal summaries. The JSON schema is unchanged (still 1.7); only the `text` card and how it prints changed.

## What changed

- Verdict-first card: Verdict / Why / Next, then a hierarchical timing tree (Step Time = Input Wait + Traced Step Time; Compute, H2D, Residual indented beneath) with share bars when a phase-share finding fires and the culprit row marked.
- Four variants: run and watch, single-machine and multi-rank. Single-machine cards contain no rank/node/skew vocabulary. Multi cards show the median tree plus a worst-rank cell only where the skew is material.
- watch gets its own card: host and process health only, an explicit attribution boundary, and a pointer to `traceml run`. No step-time or step-memory placeholders.
- Missing is never zero: unmeasured signals are omitted and explained in one line ("H2D and step memory not measured (no GPU)"). No n/a floods.
- Header carries run name, topology, steps analyzed, and duration; the footer carries the artifact path.
- Severity-scoped ANSI color when stdout is a TTY (`NO_COLOR` and `TERM=dumb` respected; standard 16-color slots, so the user's terminal theme decides the palette). `final_summary.txt`, the JSON `text` field, CI logs, and piped output stay plain ASCII. The SDK print path (`traceml.summary()`) colorizes only when its rebuilt card matches the stored text byte-for-byte; otherwise it prints the stored text unchanged.
- Secondary warnings: at most two, warn/crit, resource findings only (system, process, step memory), under "Also, not the cause of slow steps". Competing step-time findings stay in JSON.

## Implementation

- New `src/traceml_ai/reporting/summary_card.py`: span-based renderer; the plain and ANSI renderings share identical visible geometry.
- `final.py` delegates to it; profile (run/watch) threaded from `TraceMLSettings` through the aggregator and the final-summary service. Rendering failures degrade to a minimal verdict card; shutdown never breaks on a card bug.
- 15 exact-text golden cards plus property tests: watch cards never mention step time; single-machine cards never contain rank/node/skew; every line is exactly 78 characters; ANSI stripped equals plain; no `n/a` on any card.

## Validation

- 1023 tests pass; black, ruff, and isort clean on all changed files.
- CPU end-to-end: quickstart, watch, 2-rank gloo DDP.
- GPU end-to-end (4x T4): single-GPU quickstart, 4-GPU input-straggler, 4-GPU balanced, watch. Every displayed value on the straggler card matches its own `final_summary.json` (median step 272.2 ms, worst input wait 201.2 ms on rank 0 vs 1.7 ms median, ratio 115x). Real GPU payload shapes are regression fixtures.

Example card from the 4-GPU straggler run:

```text
+----------------------------------------------------------------------------+
|  TraceML Run Summary                                                       |
|  session_20260806_101748_81187e · 4 GPUs · 1 node · 128 steps analyzed ·   |
|  39.4s                                                                     |
+----------------------------------------------------------------------------+
|                                                                            |
|  Verdict: INPUT STRAGGLER  (CRITICAL)                                      |
|  Why: rank 0 waits 201.2 ms per step for input; the median rank waits 1.7  |
|  ms. All 4 ranks then advance at rank 0's pace.                            |
|                                                                            |
|  Where a step goes (median rank, GPU clock)         worst rank             |
|  Step Time           272.2 ms  100%                 272.2 ms (r3)          |
|  ├─ Input Wait         1.7 ms    1%                 201.2 ms (r0)  ◀  115x |
|  └─ Traced Step Time 270.4 ms   99%                                        |
|     ├─ Compute       268.2 ms   99%                                        |
|     ├─ H2D             0.1 ms   <1%                                        |
|     └─ Residual        2.1 ms    1%                                        |
|                                                                            |
|  Next: inspect dataloader, collate_fn, preprocessing, and storage on rank  |
|  0.                                                                        |
|                                                                            |
|  Peak step memory: 1.1 GB reserved · even across ranks                     |
|                                                                            |
|  Full evidence: logs/session_20260806_101748_81187e/final_summary.json     |
|  (--html-report)                                                           |
+----------------------------------------------------------------------------+
```

The colored TTY rendering of the same card, captured live on the validation box:

colored terminal card, input straggler

## Evidence

GATE: conformance ran=true, 1 passed / 0 failed / 1 skipped (tests/integrations/test_telemetry_conformance.py; this diff is renderer-only and does not touch telemetry emission). Card lock-down gate: 15 golden cards plus invariants, 1023 tests total passed.
TRANSITIONS: the renderer consumes one finished payload, so there is no live-view state machine in this diff. Failure transitions tested on purpose: card build raising falls back to a minimal verdict card; an SDK colorize rebuild mismatch prints the stored plain text unchanged.
RANKS: per-rank asymmetry tested? yes. A real 4-rank straggler payload is a fixture (rank 0 input wait 201.2 ms vs 1.7 ms median; worst-rank cells and the ratio marker on the card), alongside single-rank and 4-rank balanced fixtures.
TRIPWIRE: made the failure paths fire on purpose? yes. Forced build_card_from_payload to raise (fallback card observed) and tampered final_summary.txt (colorization refused, stored text printed verbatim).
PLATFORM: EVIDENCE: n/a (no platform-specific branch added; rendering is pure string assembly).

### DIFF
--- a/src/traceml_ai/aggregator/summary_service.py
+++ b/src/traceml_ai/aggregator/summary_service.py
@@ -38,6 +38,7 @@ class FinalSummaryService:
         settle_telemetry: Optional[Callable[[float], bool]] = None,
         summary_window_rows: int = DEFAULT_SUMMARY_WINDOW_ROWS,
         write_html: bool = False,
+        profile: str = "run",
     ) -> None:
         self._logger = logger
         self._session_root = Path(session_root).resolve()
@@ -46,6 +47,7 @@ class FinalSummaryService:
         self._settle_telemetry = settle_telemetry
         self._summary_window_rows = int(summary_window_rows)
         self._write_html = bool(write_html)
+        self._profile = str(profile)
         self._last_request_id: Optional[str] = None
 
     def poll(self) -> None:
@@ -87,6 +89,7 @@ class FinalSummaryService:
                 print_to_stdout=False,
                 summary_window_rows=self._summary_window_rows,
                 write_html=self._write_html,
+                profile=self._profile,
             )
 
             self._write_ok_response(response_path, request_id=request_id)
--- a/src/traceml_ai/aggregator/trace_aggregator.py
+++ b/src/traceml_ai/aggregator/trace_aggregator.py
@@ -141,6 +141,7 @@ class TraceMLAggregator:
             settle_telemetry=self._settle_telemetry,
             summary_window_rows=int(settings.summary_window_rows),
             write_html=bool(settings.html_report),
+            profile=str(settings.profile),
         )
 
         # Display driver owns renderer selection and layout mapping.
@@ -267,6 +268,7 @@ class TraceMLAggregator:
                             self._settings.summary_window_rows
                         ),
                         write_html=bool(self._settings.html_report),
+                        profile=str(self._settings.profile),
                     )
                 except Exception as summary_exc:
                     # A summary-generation error is only fatal if it left no
--- a/src/traceml_ai/reporting/SCHEMA.md
+++ b/src/traceml_ai/reporting/SCHEMA.md
@@ -53,11 +53,13 @@ Sections:
 section diagnoses. It answers "why was training slow?" and is intentionally
 narrower than section-level health/resource diagnoses.
 
-`text` is a compact human-readable verdict report. It is presentation text for
+`text` is a compact human-readable verdict card. It is presentation text for
 the CLI/TXT artifact, not a structured contract for downstream parsers. It
-starts with `TraceML Verdict`, `Why`, and `Next`, then shows compact section
-status plus System and Step Time evidence tables. Detailed section prose
-remains in each section-local `card` field.
+leads with `Verdict` and `Why`, then a hierarchical step-timing tree, `Next`,
+and bounded corroboration. It is profile-aware: `traceml watch` renders a
+host-and-process health card and never shows step timing, because watch does
+not collect it. Detailed section prose remains in each section-local `card`
+field.
 
 ## Primary Diagnosis Shape
 
--- a/src/traceml_ai/reporting/final.py
+++ b/src/traceml_ai/reporting/final.py
@@ -27,11 +27,12 @@ from traceml_ai.reporting.sections.process import ProcessSummarySection
 from traceml_ai.reporting.sections.step_memory import StepMemorySummarySection
 from traceml_ai.reporting.sections.step_time import StepTimeSummarySection
 from traceml_ai.reporting.sections.system import SystemSummarySection
-from traceml_ai.reporting.summaries.summary_formatting import bytes_to_gb
-from traceml_ai.reporting.summaries.summary_layout import (
-    border,
-    row,
-    wrap_lines,
+from traceml_ai.reporting.summary_card import (
+    build_card_from_payload,
+    build_fallback_card,
+    card_hints,
+    card_to_plain,
+    colorize_stored_card,
 )
 from traceml_ai.sdk.protocol import (
     get_final_summary_html_path,
@@ -46,32 +47,7 @@ from traceml_ai.utils.atomic_io import write_json_atomic, write_text_atomic
 # A timing signal that was never measured remains null; a measured zero is 0.0.
 SCHEMA_VERSION = 1.7
 
-SUMMARY_WIDTH = 78
-SUMMARY_INNER_TEXT_WIDTH = SUMMARY_WIDTH - 4
-
-_SYSTEM_EVIDENCE_METRICS = (
-    ("CPU Util", "cpu_percent"),
-    ("GPU Util", "gpu_util_percent"),
-    ("GPU Memory", "gpu_mem_bytes"),
-    ("GPU Temp", "gpu_temp_c"),
-)
-_STEP_TIME_EVIDENCE_METRICS = (
-    ("Step Time", "step_time_ms"),
-    ("Traced Step Time", "traced_step_time_ms"),
-    ("Input Wait", "input_wait_ms"),
-    ("Compute", "compute_ms"),
-    ("Residual", "residual_ms"),
-    ("H2D", "h2d_ms"),
-    ("DataLoader Fetch (CPU)", "dataloader_fetch_cpu_ms"),
-)
-_STEP_TIME_SUPPLEMENTAL_METRICS = frozenset(("dataloader_fetch_cpu_ms",))
-_SEVERITY_LABELS = {
-    "crit": "CRITICAL",
-    "critical": "CRITICAL",
-    "warn": "WARNING",
-    "warning": "WARNING",
-    "info": "INFO",
-}
+DEFAULT_PROFILE = "run"
 
 
 def _log_final_report_error(message: str, exc: Exception) -> None:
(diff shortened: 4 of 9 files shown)


abhinavsriva:
Tested on 1 CPU and multiple CPUs 
Changed coloring and also flip the order of status unavaialble to make it consistent

PR #367, opened 2026-08-10
Make watch/run work on a torch-free install (#346)

### LINKED ISSUE #346: watch/run crash on a torch-free install
**Problem.** `pip install traceml-ai` without the `[torch]` extra crashes every `watch` and `run` invocation: `src/traceml_ai/samplers/process_sampler.py:14-15` imports torch at module scope, so the process sampler import fails before any CLI error handling. The FAQ positions watch as lightweight monitoring; a torch-free machine cannot run it at all. Reproduced in a clean venv.

**Fix.** Guard/lazy the torch import in the process sampler (torch-dependent metrics already degrade to null by design). Add a CI probe that imports `traceml_ai.runtime` in a torch-free environment so the class of bug cannot return.

**Acceptance.** `pip install traceml-ai` (no extras) → `traceml watch script.py` runs, GPU/step metrics null, card renders. CI has a torch-free import leg.

Pendu:
Closes #346.

## Problem

`pip install traceml-ai` without the `[torch]` extra crashed every `watch` and `run` invocation. The import chain reached module-scope `import torch` before any CLI error handling, so a torch-free machine could not run the lightweight monitoring path at all.

Reproduced in a clean virtualenv on the released 0.3.6: `traceml watch script.py` ends in `ModuleNotFoundError: No module named 'torch'`.

Working through it surfaced a second blocker the issue did not know about: the launcher always spawns training via `python -m torch.distributed.run`, which cannot exist without torch. Guarding the imports alone leaves the command hanging instead of crashing.

## Fix

Guarded torch imports on the watch/run entry chain, keeping the "unmeasured, never fabricated" contract for the metrics that need it:
- `samplers/process_sampler.py`: GPU memory metrics stay null without torch, CPU and RSS keep working
- `utils/step_memory.py`, `utils/timing.py`, `utils/cuda_event_pool.py`: CPU timing keeps working, CUDA event timing is only reachable from instrumented training code, which requires torch on the user side
- `sdk/__init__.py`: monitoring and post-hoc commands import cleanly. Calling a trace API without torch raises `RuntimeError` naming the fix (`pip install 'traceml-ai[torch]'`) instead of failing at import time

Launch path (`launcher/launch_config.py`): `to_command()` now spawns the single training process directly when torch is absent and the launch is 1x1, which needs no launcher. Multi-process launches still require torch and raise an actionable `ValueError`. Both existing launch sites go through this one method, so the fallback covers `run`, `watch`, and the disabled-passthrough path.

## Tests

`tests/runtime/test_torch_free_imports.py`, four subprocess cases that block torch (`sys.modules["torch"] = None`) because the test environment has torch installed:
- the watch/run import chain imports cleanly
- a 1x1 launch falls back to a direct interpreter spawn
- a multi-process launch raises with the install hint
- calling `trace_step` without torch raises the actionable error

CI: the existing torch-free `test-core` job gains an import probe over the same entry chain, so a new module-scope torch import fails there rather than in a user's terminal.

## Evidence

GATE: conformance ran=true, 1 passed / 0 failed / 1 skipped (`tests/integrations/test_telemetry_conformance.py`). Full suite: 1036 passed, 2 skipped. black, ruff, isort clean on all changed files.
TRANSITIONS: the state change that matters here is torch present to torch absent, which is exactly what the four new tests toggle. End to end, the same command was run repeatedly in a torch-free virtualenv until the whole chain survived: `traceml watch script.py` now runs the script, streams host telemetry, and prints the watch card with CPU and RAM populated, GPU absent, and a clean exit.
RANKS: per-rank asymmetry tested? Not applicable in the usual sense: without torch the only supported topology is a single process, and that boundary is the tested behavior (1x1 falls back, anything larger raises with the install hint). Multi-rank behavior with torch installed is unchanged and stays covered by the existing suite.
TRIPWIRE: made the failure fire on purpose? Yes, twice. The torch-free virtualenv reproduced the original crash before the fix, and each guard was validated by re-running that same command until the next unguarded import in the chain appeared, so every guard was added against an observed failure rather than a suspected one. The multi-process path was made to raise deliberately in `test_multi_process_launch_requires_torch`.
PLATFORM: n/a, no platform-conditional code in this diff. The launch fallback uses `sys.executable` and is platform-neutral.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,20 @@ jobs:
           pip install -e .
           # tomli backports tomllib to 3.10 for the packaging guard.
           pip install pytest 'tomli; python_version < "3.11"'
+      - name: Torch-free watch (end to end, no torch installed)
+        run: |
+          # This job installs TraceML without the torch extra, so it is the
+          # only leg that proves the documented torch-free path really
+          # works. Importing is not enough: the failure this guards against
+          # was a module-scope torch import inside the launch chain, which
+          # only surfaces when the command actually runs.
+          set -euo pipefail
+          echo "print('training done')" > tiny_train.py
+          python -m traceml_ai.launcher.cli watch tiny_train.py \
+            --mode=summary --run-name torch_free_ci | tee watch_out.txt
+          grep -q "training done" watch_out.txt
+          grep -q "TraceML Watch Summary" watch_out.txt
+          test -f logs/torch_free_ci/final_summary.json
       - name: Run core tests
         run: |
           # Keep this matrix torch-free; torch-dependent paths run in the
--- a/src/traceml_ai/api.py
+++ b/src/traceml_ai/api.py
@@ -32,7 +32,7 @@ def trace_step(model: nn.Module) -> ContextManager[None]:
     When tracing is disabled, the context manager is a no-op. Exceptions from
     the training body still propagate normally.
     """
-    from traceml_ai.sdk.instrumentation import trace_step as _trace_step
+    from traceml_ai.sdk import trace_step as _trace_step
 
     return _trace_step(model)
 
@@ -138,9 +138,7 @@ def wrap_dataloader_fetch(obj: Any) -> Any:
     RuntimeError
         If a torch ``DataLoader`` is already automatically instrumented.
     """
-    from traceml_ai.sdk.wrappers import (
-        wrap_dataloader_fetch as _wrap_dataloader_fetch,
-    )
+    from traceml_ai.sdk import wrap_dataloader_fetch as _wrap_dataloader_fetch
 
     return _wrap_dataloader_fetch(obj)
 
@@ -167,7 +165,7 @@ def wrap_forward(model: nn.Module) -> nn.Module:
         If automatic forward instrumentation is active or the instance cannot
         be wrapped safely.
     """
-    from traceml_ai.sdk.wrappers import wrap_forward as _wrap_forward
+    from traceml_ai.sdk import wrap_forward as _wrap_forward
 
     return _wrap_forward(model)
 
@@ -192,7 +190,7 @@ def wrap_backward(loss: Any) -> Any:
     RuntimeError
         If automatic backward instrumentation is active.
     """
-    from traceml_ai.sdk.wrappers import wrap_backward as _wrap_backward
+    from traceml_ai.sdk import wrap_backward as _wrap_backward
 
     return _wrap_backward(loss)
 
@@ -219,7 +217,7 @@ def wrap_optimizer(optimizer: Any) -> Any:
         If automatic optimizer instrumentation is active or the instance
         cannot be wrapped safely.
     """
-    from traceml_ai.sdk.wrappers import wrap_optimizer as _wrap_optimizer
+    from traceml_ai.sdk import wrap_optimizer as _wrap_optimizer
 
     return _wrap_optimizer(optimizer)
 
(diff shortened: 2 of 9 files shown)


src/traceml_ai/utils/timing.py
```diff
@@ -213,7 +220,11 @@ def timed_region(
     cpu_start = time.time()
 
     try:
-        if record_gpu_events and torch.cuda.is_available():
+        if (
+            record_gpu_events
+            and torch is not None
```
abhinavsriva: Is this required even after torch check is done ?

abhinavsriva (dismissed):
I confirmed the original Torch-free import problem, but the current tests do not fully validate the stated public contract. 

The documented traceml_ai.trace_step() path still raises ModuleNotFoundError; only the internal traceml_ai.sdk path is tested. 

The broad except ImportError can also mask unrelated import regressions as a missing-Torch error. Additionally, multi-process Torch-free configuration passes CLI validation and fails later, and CI only imports modules rather than executing a minimal Torch-free watch/run. 

Could we tighten the missing-Torch detection, test the public API, validate topology before launch side effects, and add one end-to-end Torch-free CLI test?

Pendu:
Thanks, all four points were right and are addressed. Rebased on the current base and pushed.

**Your inline question on `timing.py` ("is this required even after the torch check?"): no, and it led to a better fix.** Every `timed_region` caller is torch-dependent instrumentation (the auto-timer patches, the Lightning integration, `sdk/instrumentation`, `sdk/wrappers`), so that runtime guard was unreachable. The reason the module needed guarding at all was the sampler registry, which imported the step samplers eagerly for every profile, including `watch`.

So the registry now builds the two run-profile samplers lazily. `watch` no longer imports the torch-backed timing chain at all, which let me revert the guards in `utils/timing.py`, `utils/cuda_event_pool.py`, and `utils/step_memory.py` entirely. Three fewer guarded modules, and the boundary now sits where the profile split already was. A test asserts `traceml_ai.utils.timing` is absent from `sys.modules` after selecting watch samplers, so this cannot regress silently.

**1. The public API.** Confirmed: `traceml_ai.trace_step()` raised `ModuleNotFoundError` because `api.py` lazily imported the submodules directly, bypassing the package fallback. `api.py` now imports those six entry points from `traceml_ai.sdk`, so `trace_step`, `wrap_forward`, `wrap_backward`, `wrap_optimizer`, `wrap_dataloader_fetch`, and `wrap_h2d` all raise `RuntimeError` naming the install fix. Parametrized over all six. `wrap_h2d` was also missing from the sdk exports; it is exported now.

**2. Missing-torch detection.** A new `utils/torch_support.py` holds `is_missing_torch(exc)`, which checks `exc.name` for `torch` or a `torch.*` submodule. Both remaining guards use it and re-raise anything else. Tested two ways: a unit table over module names, and a case that blocks an unrelated module with torch installed, asserting the error propagates instead of the torch-free fallback being installed.

**3. Topology validation.** `validate_launch_args` now rejects a torch-free multi-process launch before any process starts, via `TorchrunLaunchConfig.requires_torch_launcher()`. The raise in `to_command()` stays as a backstop for a caller that skipped validation. Parametrized over 1x2, 2x1, and 2x4; verified by hand that the CLI prints one error and starts no aggregator.

**4. End-to-end CI.** The torch-free CI job now runs a real `traceml watch` on a small script and asserts the training output, the summary card, and the written artifact, rather than importing modules. The same run exists as a pytest case, which blocks torch through a `sitecustomize` module on `PYTHONPATH` so the spawned aggregator and training processes are torch-free too. That test is hermetic: it strips inherited `TRACEML_*` variables and passes its own `--logs-dir`.

Verification: 1062 passed, 2 skipped. black, ruff, and isort clean on the changed files. Reverting the public-API fix and the validation call turns exactly the matching tests red, so they fail for the stated reasons. Also re-checked by hand in a clean virtualenv without the torch extra: `traceml watch` prints the watch card, and single-process `traceml run` completes and reports no step data, which is the honest result without instrumentation.

abhinavsriva:
Thanks, this now cleanly separates the contracts: watch works **without Torch**, while run fails early with an actionable installation hint. Unrelated Torch import failures are no longer masked, and the focused tests pass.

PR #370, opened 2026-08-10
Always show GPU utilization beside a step-time finding (#349)

### LINKED ISSUE #349: Show GPU utilization unconditionally on GPU run cards
**Problem.** The run card shows GPU utilization on most states but suppresses the supporting line when a finding fires and utilization is >= 50%. Engineers think in nvidia-smi terms and expect the number on every run; its absence on exactly the interesting cards reads as a gap.

**Fix.** Make the GPU-utilization line unconditional on GPU run cards (value from `system.global.average.gpu_util_percent`, wording per state as today). One policy change + golden updates.

**Acceptance.** Every run-card golden with `gpus_observed > 0` contains a GPU-util line.

Pendu:
Closes #349.

## Problem

The run card hid GPU utilization on exactly the cards where a reader most wants it. The supporting line was suppressed whenever utilization was at or above 50%, so a card reporting a step-time finding on a busy GPU showed no utilization figure at all.

That is backwards. GPU utilization is the number most engineers reach for first, and its absence reads as missing data rather than as a deliberate omission. The healthy card already prints utilization, which made the gap inconsistent as well.

## Fix

Show the utilization figure for every step-time finding on a GPU run. Only the explanation stays conditional:

- below 50%, the line keeps its existing wording, because low utilization corroborates the finding above it (`GPU util 24% avg -- consistent with input starvation.`)
- at or above 50%, the number is reported without a claim attached (`GPU util 88% avg.`), since a busy GPU does not corroborate the finding and the card must not imply that it does

The threshold constant is renamed to say what it now gates, which is the explanation rather than the line. Behavior is unchanged for CPU-only runs (no utilization to report), for healthy runs, and for degraded states, which already print utilization through their own lines.

## Tests

- `test_gpu_utilization_is_shown_for_a_finding_at_any_level` asserts both halves: the explanatory wording at low utilization, and the bare figure with no claim at high utilization
- the `run_input_bound_with_also` golden gains the line it was previously suppressing (that fixture runs at 88% utilization)

## Evidence

GATE: conformance ran=true, 1 passed / 0 failed / 1 skipped (`tests/integrations/test_telemetry_conformance.py`). Full suite: 1033 passed, 2 skipped. black and ruff clean on both changed files.
TRANSITIONS: the transition this diff owns is low to high utilization across the explanation threshold, and both sides are asserted on real fixtures rather than one side being assumed. The card's other states (healthy, CPU-only, degraded, watch) are unchanged and stay locked by their existing goldens.
RANKS: per-rank asymmetry tested? Yes, through the existing multi-rank goldens: the straggler card keeps its median-phrased line (`GPU util median 14% -- ranks idle at the step barrier.`) and the multi residual-heavy card keeps its plain line, so the single and multi wordings are both pinned.
TRIPWIRE: made the failure fire on purpose? Yes. Restoring the old suppression and re-running turns both the new test and the updated golden red, so they fail for the reason the issue describes rather than passing by construction.
PLATFORM: n/a, no platform-conditional code in this diff.

### DIFF
--- a/src/traceml_ai/reporting/summary_card.py
+++ b/src/traceml_ai/reporting/summary_card.py
@@ -168,7 +168,9 @@ _WATCH_MULTI_WORST_COLUMN = 26
 _SKEW_MATERIAL_RATIO = 0.20
 _SKEW_MATERIAL_DELTA_MS = 1.0
 _MEMORY_SKEW_EVEN_RATIO = 0.10
-_SUPPORTING_GPU_UTIL_MAX = 50.0
+# Below this, GPU utilization corroborates a step-time finding and the card
+# says so. At or above it the number is still shown, but without a claim.
+_LOW_GPU_UTIL_MAX = 50.0
 
 _WATCH_BOUNDARY = (
     "Watch monitors host and process health only; it does not measure "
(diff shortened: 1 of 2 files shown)


abhinavsriva (approved):

abhinavsriva (changes requested):
This fixes the visibility issue, but should the formatter contain a separate 50% threshold? 

TraceML already defines canonical GPU-utilization bands—low below 30%, moderate through 70%, and healthy above 70%.

Keeping _LOW_GPU_UTIL_MAX = 50.0 duplicates policy and creates inconsistent behavior at 50%. 

Could we always display the measurement neutrally, or derive any interpretation from the existing diagnostic classification instead?

abhinavsriva:
I have opened #372, which addresses the broader Run-card evidence gap and explicitly requires GPU utilization to remain visible as part of the System block. Merging this PR would introduce an intermediate presentation and retain a separate 50% formatter threshold that #372 would replace. 

Let’s close this without merging and treat #349 as superseded by #372. We should carry the useful test requirement from this PR:GPU utilization is visible for every GPU Run state, into #372’s golden tests.

PR #377, opened 2026-08-12
fix(dashboard): hold last-known diagnosis on empty refresh (#353)

### LINKED ISSUE #353: Dashboard diagnosis panel blanks instead of holding last-known state
**Problem.** Since the 0.3.6 step-time rework, the dashboard diagnostics section dropped its `_last` payload cache (`model_diagnostics_section.py`, v0.3.5→v0.3.6 diff): any transiently empty diagnostics read (ranks settling, end-of-run) now blanks the panel to "NO DATA / Waiting for diagnostics" where it previously re-rendered the last-known finding.

**Reproduction.** Run a short multi-rank job with `--mode dashboard` and kill one rank (or watch the end-of-run transition): the diagnostics panel blanks instead of holding the last finding.

**Fix.** Restore last-known-state behavior (short-term: reinstate the cache in the section; long-term: the shared live-engine RFC's snapshot fallback owns it).

**Acceptance.** Kill one rank mid-run: the panel keeps the last finding with a staleness hint instead of blanking.

Pendu:
Closes #353.

**Problem.** #301 dropped the `_last` payload cache from the dashboard
diagnostics rail. Any transiently empty diagnostics read (ranks settling,
end of run, a rank dying) now blanks the panel to "NO DATA / Waiting for
diagnostics" instead of re-rendering the last finding the engine produced.

**Fix.** Reinstate the cache. A non-empty payload is stored; an empty read
re-renders the cached one and sets the hint to "Holding last diagnosis - no
update in the latest refresh". With no cached payload the NO DATA path is
byte-identical to today. The card adds no threshold, no staleness age, and no
severity of its own; it states only what the view itself observed.

**Testing.** New `tests/display/test_model_diagnostics_section.py` (6 tests):
render+cache, empty-after-cached holds with the hint, empty-with-no-cache
still shows NO DATA, a second empty refresh does not evict the cache, and a
recovered diagnosis replaces the held one. `tests/display/` 74 passed;
full suite 1048 passed, 2 skipped.

GATE: tests listed above run with exit codes shown (tests/display 74 passed
EXIT=0; full suite 1048 passed / 2 skipped EXIT=0; black/ruff/isort exit 0).
TRANSITIONS: empty→NO DATA covered by test_empty_refresh_without_a_cache_shows_no_data;
populated→held by test_empty_refresh_holds_the_last_diagnosis; held→held by
test_repeated_empty_refresh_does_not_evict_the_cache; held→recovered by
test_recovered_diagnosis_replaces_the_held_one.
RANKS: rank-agnostic. This is view code over the already-aggregated
model-diagnostics payload; it reads no rank field and changes no per-rank
behavior.
TRIPWIRE: with the fix reverted, test_empty_refresh_holds_the_last_diagnosis
fails with `assert 'NO DATA' == 'CRIT'` (5 of 6 red); restored, all green.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_diagnostics_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_diagnostics_section.py
@@ -74,7 +74,7 @@ def build_model_diagnostics_section(
             "font-family:var(--mono); font-size:11px; color:var(--muted); "
             "font-style:italic; padding-top:6px;"
         )
-    return {"overall": overall, "body": body, "hint": hint}
+    return {"overall": overall, "body": body, "hint": hint, "_last": None}
 
 
 def update_model_diagnostics_section(
@@ -82,6 +82,16 @@ def update_model_diagnostics_section(
 ) -> None:
     try:
         items = _items(payload)
+        held = False
+        if items:
+            panel["_last"] = payload
+        else:
+            # A transiently empty read (ranks settling, end of run, a rank
+            # dying) is not a new verdict, so re-render the last one the
+            # engine produced instead of blanking the rail (#353).
+            payload = panel.get("_last")
+            items = _items(payload)
+            held = bool(items)
         if not items:
             panel["overall"].text = "NO DATA"
             panel["body"].content = ""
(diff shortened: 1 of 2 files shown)


abhinavsriva:
@Pendu This change appears redundant and is not required on the current code path. `LiveStepTimeSession` already bridges transient empty/read-failure states, while expired or unavailable diagnostics produce explicit `NO DATA` items rather than an empty payload.

The new cache is therefore only reached on an exceptional renderer failure and can make the diagnostics rail disagree with the hero verdict. I recommend closing this PR and #353 as no longer reproducible.

PR #378, opened 2026-08-12
fix(dashboard): render the engine's action and note in the diagnosis rail (#357)

### LINKED ISSUE #357: Dashboard drops the diagnosis action line the terminal prints
**Problem.** The terminal's live diagnosis prints the engine's `Next:` action; the dashboard rail shows status + why but drops `action` (and `note`). Same engine verdict, different advice depending on surface.

**Fix.** Render `action` in the dashboard diagnosis rail (truncation rules as for the summary line).

**Acceptance.** The same live finding shows the same next-step text on both surfaces.

Pendu:
Closes #357.

**Problem.** The terminal prints the engine's `Next:` action for a live
verdict (`renderers/step_time/renderer.py:79`), but the dashboard's
diagnostics rail rendered only source, status, reason, confidence and
evidence chips. The same engine verdict gave a next step on one surface and
none on the other (#357).

**Fix.** Render-only change in `model_diagnostics_section._row_html`. The
payload already carried `action` and `note`
(`diagnostics/model_diagnostics.py:69-70`), so no engine or payload change
was needed. The row now prints `Next: <action>` verbatim under the same label
the terminal uses, and the `note` below it in muted 11px. Both are escaped
and length-capped through a new `_line_text` helper that reason now shares,
replacing the duplicated 130/127 literals with one `_MAX_LINE_CHARS`. Empty
or missing fields render nothing.

**Testing.** Five tests in `tests/display/test_model_combined_section.py`,
including a cross-surface parity test that takes one real INPUT_BOUND verdict
and asserts the same action string appears in `format_cli_diagnosis` and in
the rail HTML. `tests/display` 75 passed; full suite 1049 passed, 2 skipped.

GATE: repo-pinned black/isort/ruff hooks pass on both changed files; full
suite 1049 passed / 2 skipped, exit 0.
TRANSITIONS: present, absent, empty-string, whitespace-only and over-length
action all covered; note covered present/absent.
RANKS: view-only change, rank-agnostic; no per-rank or per-metric logic
touched.
TRIPWIRE: with the source change reverted, the four positive tests fail and
the negative cases stay green; restored, all eight pass.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_diagnostics_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/model_diagnostics_section.py
@@ -1,7 +1,9 @@
 """Model diagnostics rail — overall verdict + per-source severity rows (PR2).
 
-Renderer payload {overall_severity, items:[{source,status,reason,
-confidence_label,evidence}]} is unchanged; only the presentation changes.
+Renderer payload {overall_severity, items:[{source,status,reason,action,
+note,confidence_label,evidence}]} is unchanged; only the presentation
+changes. The row shows the engine's own ``action`` under the same "Next:"
+label the terminal uses, so one verdict reads the same on both surfaces.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


abhinavsriva (approved):
Approved. This PR cleanly renders the existing engine-provided action and note without changing diagnosis logic. 


Step-time trend notes are currently not reachable in the default live dashboard because its 100-point window is smaller than the trend engine’s 200-point minimum. That behavior should be evaluated separately in a diagnostics-focused issue/PR.

PR #379, opened 2026-08-12
fix(dashboard): watch profile hides step panels and states the watch boundary (#355)

### LINKED ISSUE #355: Dashboard is profile-blind: watch shows permanently dead step panels
**Problem.** The dashboard renders the full run layout regardless of profile (step panels built unconditionally at `nicegui_sections/pages.py:131,178,183`): under `watch`, Step Time and Step Memory panels sit permanently empty with no explanation. `serve` cannot set a profile at all.

**Fix.** Profile-aware layout: watch hides step panels and states what watch collects (mirroring the watch card's boundary line); serve gains profile plumbing.

**Acceptance.** `traceml watch --mode dashboard` shows system/process only + one boundary line; no empty step panels.

Pendu:
Closes #355.

## Problem

The dashboard built the full run layout for every profile.
`traceml watch --mode dashboard` therefore showed a Step Time hero and a Step
Memory panel that watch telemetry can never fill, with nothing on screen
saying why (#355). The terminal driver already selects its layout per profile
(`cli.py:150-152`); the dashboard did not.

## Fix

- `_sections_for_profile(profile)` in `nicegui_sections/pages.py` answers
  which section groups a profile's page builds. Watch drops the step-coupled
  groups (hero, step memory); every other profile, including an unknown or
  missing one, builds everything as before. Layout selection only: no
  thresholds, no severity, no diagnosis.
- Excluded panels are neither created nor subscribed, so no layout
  subscription is registered for step data that will not arrive. The shared
  diagnostics subscriber drives the rail alone when there is no hero.
- Under watch the page states the boundary once, under the header. The
  sentence is not written here: `pages.WATCH_BOUNDARY_LINE` is the summary
  card's own `_WATCH_BOUNDARY` constant, so the two surfaces state the same
  boundary by construction and no new user-facing copy is introduced.
- The Diagnostics rail, System, Process and GPU gauge stay in watch; the rail
  carries System and Process findings.
- The page body moved to a module-level `build_main_page(cls, profile)` so
  the layout is testable without a browser. Read the diff with `-w`: most of
  it is re-indentation (83 insertions / 13 deletions whitespace-insensitive).

## Testing

`tests/display/test_nicegui_profile_layout.py`, 19 tests: profile selection
(run / watch / deep / unknown / empty / None, plus case and whitespace
normalization), the boundary line, and the resulting subscriptions built
headlessly through the real driver (watch subscribes system + process +
diagnostics only; run adds the two step layouts; the watch diagnostics
subscriber updates the rail with no hero present).

- new file: 19 passed; `tests/display/`: 87 passed
- full suite: 1061 passed, 2 skipped
- black / ruff / isort clean

GATE: full suite green (1061 passed, 2 skipped) and the layout tests fail on
the base commit; verified by reverting pages.py (18 failed) and restoring
(19 passed).
TRANSITIONS: profile run / watch / deep / unknown / empty / None all covered;
watch matching is case- and whitespace-normalized the way the summary card
does it.
RANKS: rank-agnostic. This changes which panels are built for a profile, not
any per-rank payload, metric, or aggregation; no rank data is read on the
changed path.
TRIPWIRE: reverting the pages.py change turns the layout tests red (18
failed, 1 passed); restored, 19 pass. test_watch_boundary_comes_from_the_
summary_card additionally fails if the dashboard ever stops sourcing the
card's constant and words its own boundary sentence again.

Known, deliberately out of scope:

- `traceml serve` cannot express a profile (`_resolve_serve_settings` never
  sets it), so serve + dashboard is always the run layout. Needs a
  `--profile` flag decision.
- The dashboard still computes step payloads under watch
  (`nicegui.py:157-170`); this PR changes layout only.
- `reporting/html/sections.py` renders Step time / Step memory cards
  unconditionally, the same class of issue on the HTML report surface.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/pages.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/pages.py
@@ -9,6 +9,7 @@ from traceml_ai.aggregator.display_drivers.layout import (
     PROCESS_LAYOUT,
     SYSTEM_LAYOUT,
 )
+from traceml_ai.reporting.summary_card import _WATCH_BOUNDARY
 
 from . import theme
 from .model_combined_section import (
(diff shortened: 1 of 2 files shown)


abhinavsriva (changes requested):
The problem is valid, but this patch only hides parts of the run dashboard and leaves related NO DATA states and unnecessary background work.

I suggest closing this PR without merging and handling #355 through a dashboard redesign issue:

- Watch: a clean resource-health page using existing System, Process, and GPU data.
- Run: reuse the Watch sections and add Step Time, Step Memory, and training diagnostics.

This requires no new telemetry and would produce two clear, intentional views instead of patching the current layout. Please keep #355 open or link it to the redesign issue until the behavior is actually resolved.

Pendu:
Agreed, closing this without merging is the right call. The patch works at the wrong layer: the dashboard driver has no profile concept, so under watch it still refreshes step time on every tick (`_compute_model_payloads` in `display_drivers/nicegui.py`) and the diagnostics rail still builds step-time rows from that result. Hiding the panels touched neither, which is exactly the NO DATA and background-work residue you describe.

Two things worth carrying into the redesign so this class is resolved rather than moved: the driver should select what it computes per profile, the way the CLI driver already does (`display_drivers/cli.py`), and the diagnostics rail should not build step rows for a profile that never produces step telemetry. If useful, the section-selection helper and its tests on this branch (`_sections_for_profile` in `nicegui_sections/pages.py`, `tests/display/test_nicegui_profile_layout.py`) can be lifted into the redesign; the rest can go.

Keeping #355 open and linking it to the redesign issue sounds right, since the watch behavior is unchanged until then.

Issue #372, opened 2026-08-10
Complete the existing `traceml run` terminal card


abhinavsriva:
## Problem

The current Run card explains Step Time well, but it can hide the System and
Process measurements already present in `final_summary.json`. Step Memory is
reduced to one peak-memory line without its section status, and the Compute row
does not show forward/backward/optimizer values even when they are available.

The result can show a verdict without **enough visible evidence** to verify the
other section findings.

## Goal

Extend the current card. Do not replace its visual structure.

Keep unchanged:

- card border, title, metadata header, and ANSI styling;
- `Verdict`, `Why`, the existing timing tree, and culprit marker;
- `Next`, bounded secondary findings, and artifact path.

Add the missing evidence after the timing tree and before `Next`:

1. System status and core values.
2. Process status and core values.
3. Step Memory status alongside the existing allocated/reserved values.
4. Forward, backward, optimizer, and supplemental DataLoader fetch values in
   the existing Step Time block.

## Output rules

- Use only the in-memory payload that is also written to `final_summary.json`.
  Do not query SQLite, add thresholds, or recompute a diagnosis.
- Always show available core System values: CPU average, RAM bytes/percent,
  GPU-utilization average, GPU-memory bytes/percent, and average temperature.
- Always show available core Process values: CPU-capacity average, RSS
  bytes/percent, CUDA used bytes, and CUDA reserved bytes/percent.
- Show power, headroom, peak values, or growth values only when they support an
  existing displayed diagnosis.
- When System, Process, or Step Memory is non-normal, show its existing numeric
  trigger and structured scope. A general average must not be used to justify a
  peak- or rank-based warning.
- Preserve `null` as unavailable and `0` as measured zero. Omit a null metric;
  show one short reason if an entire section was not measured.
- System/Process values cover their own observation windows. Do not describe
  them as step-aligned.
- Replace, rather than duplicate, the current peak-memory line with the richer
  Step Memory line.
- Do not repeat a section's primary finding under `Other findings`. Keep that
  block only for additional warning/critical issues already present in
  `section.issues`.
- Keep the completed card within roughly seven logical lines of the current
  card, excluding unavoidable wrapping and additional critical findings.

### Multi-process correctness

- Multi-process means `meta.world_size > 1`, including one-node DDP. Fall back
  to section rank metadata/rows only when `world_size` is unavailable.
- Replace `Where a step goes (median rank, ...)` with
  `Representative step — rank R (node N) · ... timing`.
[...]

PR #381, opened 2026-08-12
Complete the end-of-run terminal summary card

abhinavsriva:
## Summary

Replace the sparse `traceml run` terminal card with a complete, payload-only summary that explains the promoted verdict and exposes the key timing, memory,
system, and process evidence without changing diagnostics or the JSON schema.

The Run card is now a fixed 2×2 layout:

```text
STEP TIMING     || STEP MEMORY
SYSTEM METRICS  || PROCESS METRICS
```

Watch output is unchanged.

## What changed

- Reworked the Run card to show `Verdict`, `Why`, and `Next` before the
  evidence panes.
- Added a coherent Step Timing tree with Forward, Backward, and Optimizer
  children; distributed timing reads every phase from the stored median-Step
  Time rank row.
- Added compact Step Memory, System, and Process tables using only stored
  averages and stored median/worst points.
- Added compact stored-evidence lines for non-normal resource findings and
  `Evidence: None` for normal/balanced sections.
- Added a shared identity formatter and distributed-card legend:
  `N = node · R = global rank · G = GPU index`.
- Preserved measured zeros, omitted unavailable metrics, and kept all tables
  scoped to their own observation windows.
- Moved terminal-card implementation from the monolithic `summary_card.py`
  into focused `terminal_card/` modules for layout, shared formatting, each
  Run section, compact evidence, and unchanged Watch rendering.
- Updated README, Quickstart, output-reading guidance, input-pipeline example,
  Accelerate CPU-only guidance, and the schema description to match the card.
- Added formatter golden/documentation-drift coverage for the supported Run
  topologies and diagnosis/resource states.

## Deliberately unchanged

- No JSON schema, SQLite, collector, diagnostic, threshold, or CLI/API change.
- No new attribution or health policy: the card only formats values and
  diagnoses already present in `final_summary.json`.
- Historical stored cards remain unchanged; `traceml view --re-render` uses the
  current renderer.
- Watch card rendering remains byte-for-byte unchanged.

## Validation

```text
  tests/reporting/summary tests/docs/test_terminal_summary_docs.py \
  tests/reporting/test_view.py -q
# 337 passed (one existing pynvml deprecation warning)

git diff --check
```

`mkdocs build --strict` reaches the build but stops on existing repository
configuration warnings: README/index conflict, unlisted `index.md`, and
missing Black/Ruff for signature formatting. This change adds no documentation
build errors.


Pendu:
Reviewed at 2b5d593, including the three commits from today. Verification first: reporting and docs suites pass at head (430 passed), and I also ran this branch merged with #377-#380 through a validation matrix on a 4x T4 box, 28 runs across the diagnosis states. The card held up well: it produced the first real H2D-BOUND and HIGH PRESSURE cards we have ever had, and a cross-surface check (card verdict vs `primary_diagnosis` vs HTML) came back clean on every healthy run.

Three things I had flagged locally are already fixed by today's commits, confirmed at head: the step-memory pane now reads Allocated and Reserved from one coherent selected row (264ebc0, matching the #372 rule), the `Evidence: None` fallback is gone (14ff9da), and no rendered text carries victim/culprit wording any more (2b5d593); the straggler sentences now read as flat per-rank measurements. Also confirmed: the #370 carry-over is pinned, with GPU utilization asserted in the goldens across straggler, compute-bound, and no-bottleneck GPU states.

One residue of the victim/culprit cut worth a deliberate decision rather than a drive-by: the vocabulary still lives in the JSON evidence contract (`victim_rank`, `visible_victim_ms`) and in the DIAGNOSIS.md prose. Anyone reading `final_summary.json` or the docs still meets it. Renaming the JSON keys is a schema-versioned change, so it probably belongs in the schema-governance lane rather than this PR; flagging it here so the cut does not look more complete than it is.

Two open questions, both yours to decide:

1. **Run-card width is now 156 columns** (`layout.py`: `RUN_CARD_WIDTH = CARD_WIDTH * 2`). On an 80-column terminal, the common default, every line wraps and the box border shatters. The watch card correctly stays at 78. Since the run card is the README money shot, worth deciding deliberately: keep 156 and document a terminal-width expectation, or detect width and fall back to the stacked 78-column layout.

2. **Pre-1.7 payloads render a contradiction.** A schema 1.5 `final_summary.json` (no `median` block) shows `STEP TIMING unavailable: no median rank was selected.` while the verdict above it still asserts step timing, e.g. "Compute took 99% of Step Time". Reproducible with `traceml view --re-render` on any 0.3.5-era summary, which is a live user path since #369 exists specifically to rebuild older cards. Suggest either falling back to the average-based tree when the median block is absent, or suppressing the timing claim in the verdict for coherence.

One coordination item, mine not yours: this PR deletes `reporting/summary_card.py`, and my #379 imports `_WATCH_BOUNDARY` from it. Git merges the two cleanly and then the dashboard fails at import. The one-line repoint to `terminal_card/watch.py` is ready on my side; simplest sequence is you merge #381 first and I rebase #379 on it.

Two smaller notes from the matrix runs, neither blocking: the card footer prints the evidence path relative to the logs root (for example `Full evidence: logs/run/final_summary.json`), which resolves from neither the invoking directory nor absolutely when `--logs-dir` points elsewhere; and on FSDP runs the communication time lands in Residual, which reads fine but is documented nowhere and will surprise FSDP users.

abhinavsriva:
Watch is changed in the next issue 373. This is only for run 

This belongs to --re-render compatibility. Older schemas use different timing meanings, so --re-render should not apply the current card renderer to payloads below schema 1.7; it should show the stored card instead. I updated it now with a simple guard.

Issue #362, opened 2026-08-07
Design: watch-specific verdict promotion


Pendu:
**Problem.** The promotion policy pins every watch run's headline to `INSUFFICIENT_STEP_TIME_DATA` (v1 policy, `reporting/primary_diagnosis.py:40-46`); real system/process findings can never head a watch summary. The watch card currently derives its health line presentation-side as a workaround.

**Proposal.** A watch profile promotion policy: promote the top system/process finding (severity, then impact) to the watch verdict; run profile unchanged. The card then consumes the engine verdict on both profiles identically.

PR #384, opened 2026-08-17
Add complete Process evidence to the `traceml watch` card

### LINKED ISSUE #373: Add complete Process evidence to the existing `traceml watch` card
## Problem

The current Watch card already has useful host-health and System evidence, but
Process evidence is incomplete: a one-process card shows little beyond RSS,
and a multi-process card shows node-level System comparisons without the
corresponding rank-level Process comparisons.

One-node multi-process Watch is also treated like a one-process card because
the current layout decision is based on node count rather than process count.

## Goal

Add a compact, explicit Process block, fix one-node multi-process routing, and
use the same fixed 156-column System/Process layout as Run.

Keep unchanged:

- title, metadata header, ANSI styling, and artifact path;
- current System and Process diagnostic meaning;
- Watch's resource-only attribution boundary.

Changed intentionally:

- Watch now shares Run's 156-column System/Process layout instead of retaining
  the previous compact layout;
- the separate `Host health` and `Observation` lines are removed because the
  System and Process headings now surface those findings directly;
- Watch ends with the short transition
  `Next: Wrap your step with trace_step(model), then traceml run.`

## Output rules

- Watch shows System and Process evidence only. It must never show analyzed
  steps, Step Time, Step Memory, or claim a cause for low GPU utilization.
- Use only the existing final-summary payload. Do not query SQLite, apply new
  thresholds, or create a Watch-only diagnosis.
- Keep the current System values, but label averages accurately; average
  temperature must not be called `max`.
- Add Process status plus CPU-capacity average, RSS bytes/percent, CUDA used
  bytes, and CUDA reserved bytes/percent when available.
- For a non-normal System/Process status, show the stored numeric trigger and
  structured node/GPU or rank/node scope. Core averages alone do not explain a
  peak-based warning.
- Preserve null versus measured zero. Omit individual null metrics; show one
  short reason when an entire block is unavailable.
- Avoid duplication: System and Process headings carry the resource findings
  without a separate `Host health` headline.
- Show the scope legend only when visible card content uses an `N`, `R`, or
  `G` identity.

### Multi-process behavior

- Multi-process means `meta.world_size > 1`, even when `nodes_observed == 1`.
- On one node, keep one compact System block and add median/worst Process
  ranks.
- On multiple nodes, show the median/worst System-node table and the
  median/worst Process-rank table.
- Use compact `R<rank>/N<node>` rank scope. Each metric keeps its own stored
  median/worst `idx`; do not imply that one rank represents the whole table.
- When bytes and percent appear in one cell, select by the percent point and
  read both values from that same grouped row.
- Label `meta.gpus_observed` as `GPUs observed`, not GPUs used.
- Resolve Watch rank coverage from Process telemetry and node coverage from
  System telemetry. Do not invent observed rank counts from `world_size`.

## Expected output

The examples below omit border padding; the rendered card keeps the shared
156-column Run geometry.

### One process

```text
TraceML Watch Summary
watch-single · 1 rank · 1 GPU observed · 5m 12s

SYSTEM METRICS: NORMAL             || PROCESS METRICS: NORMAL
CPU               18% avg          || CPU capacity    14% avg
RAM used          6.2 GB (19%)     || RSS used        3.1 GB (10%)
GPU util          76% avg          || CUDA used       2.9 GB
GPU memory/device 10.3 GB (64%)    || CUDA reserved   3.2 GB (20%)
GPU temperature   61C avg          ||

Next: Wrap your step with trace_step(model), then traceml run.
Full evidence: logs/watch/final_summary.json  (--html-report)
```

### Multiple processes and nodes

```text
TraceML Watch Summary
watch-multinode · 12/12 ranks · 12 GPUs observed · 3/3 nodes · 12m 40s

Scope: N = node · R = global rank · G = GPU index

SYSTEM METRICS: VERY HIGH GPU MEM · 3/3 nodes || PROCESS METRICS: HIGH PROCESS RSS · 12/12 ranks
Evidence: GPU memory 96%, N0/G3                 || Evidence: Process RSS 84%, R7/N1

CPU               24% median · 71%, N1         || CPU capacity  18% median · 61%, R7/N1
RAM used          18.4 GB median · 38.2 GB, N1 || RSS used      3.2 GB median · 5.8 GB, R7/N1
GPU util          88% median · 62%, N2         || CUDA used     8.1 GB median · 9.7 GB, R2/N0
GPU memory/device 71% median · 93%, N0         || CUDA reserved 56% median · 67%, R2/N0
GPU temperature   64C median · 81C, N0         ||

Next: Wrap your step with trace_step(model), then traceml run.
Full evidence: logs/watch/final_summary.json  (--html-report)
```

For one-node multi-process Watch, use the same Process comparison with rank
scope ending in `/N0`; do not fall back to the one-process layout.

## Existing payload sources

| Content | Source |
| --- | --- |
| Header/topology | `meta`, System node coverage, Process rank coverage |
| System/Process status | existing section `diagnosis` and ordered `issues` |
| One-process values | `system|process.global.average` |
| Multi-node System values | `system.global.median|worst` plus grouped node identities |
| Multi-process values | `process.global.median|worst` plus grouped rank identities |

## Acceptance criteria

- [ ] Watch uses the shared 156-column System/Process layout used by Run.
- [ ] Process status and core evidence are visible for one and multiple
      processes.
- [ ] One-node multi-process Watch uses the multi-process layout.
- [ ] Every displayed non-normal status includes its stored numeric trigger and
      structured scope when available.
- [ ] Watch rank coverage comes from Process telemetry without inventing
      observations from `world_size`.
- [ ] System node coverage remains visible when Process telemetry is degraded.
- [ ] The scope legend appears only when visible content uses compact scope.
- [ ] One-process, one-node multi-process, multi-node, normal,
      low-utilization, pressure, partial-coverage, CPU-only, null, and
      measured-zero cases are covered by golden tests.
- [ ] Watch does not render a performance verdict, Step Time, Step Memory,
      analyzed-step counts, or step-time attribution.
- [ ] Watch ends with
      `Next: Wrap your step with trace_step(model), then traceml run.`
- [ ] Plain and ANSI renderings contain identical information and geometry.
- [ ] Rendering failures produce a truthful fallback with a full-evidence
      pointer.

## Non-goals

- Changing Watch's resource-only diagnostic meaning.
- Step-time measurement or attribution in Watch.
- New JSON fields, metrics, thresholds, or collection.
- Per-GPU comparison outside an existing structured System diagnosis.
- A responsive or narrow-terminal layout in this PR.

abhinavsriva:
Closes #373

## Summary

Complete the existing `traceml watch` resource summary by adding explicit
Process status and measurements for both single-process and distributed
workloads.

Watch now renders the same fixed-width System and Process panes used by Run,
while preserving Watch's resource-only boundary: it does not show a performance
verdict, step timing, step memory, or root-cause attribution.

## What changed

- Added a Process pane with CPU-capacity average, RSS bytes/percent, CUDA-used
  bytes, and CUDA-reserved bytes/percent when those values are available.
- Route Watch by process/rank coverage, so one-node multi-process workloads use
  the distributed layout instead of falling back to the single-process card.
- For distributed workloads, show per-metric median and worst rank averages
  with compact `R<rank>/N<node>` identities. Each metric keeps its own stored
  selector rather than implying that one rank represents the entire table.
- Keep paired RSS and CUDA-reserved bytes/percent on the same selected rank row,
  while preserving measured zero and omitting unavailable individual metrics.
- Show compact stored evidence and structured scope for non-normal System and
  Process findings; normal sections leave the evidence row blank.
- Use Process metadata for Watch rank coverage and System metadata for observed
  GPU and node coverage. GPU counts are labeled as observed, not used.
- Share the System/Process pane builders, fixed divider, scope legend, and
  footer between Run and Watch.
- Update the output guide, FAQ, and reporting-schema documentation to describe
  the profile-aware card behavior and the observation-window semantics.

## Watch boundary preserved

- Watch displays only System and Process evidence.
- It does not mention `Verdict`, `Why`, `Next`, Step Time, Step Memory, analyzed
  steps, or performance causes.
- No new metrics, thresholds, diagnostics, payload fields, SQLite reads, or
  collection behavior are introduced.
- Run keeps its existing performance verdict and Step Timing/Step Memory panes.

## Test coverage

Golden and plumbing coverage now includes:

- healthy single-process Watch;
- one-node multi-process low-GPU-utilization Watch;
- multi-node System/Process comparisons;
- host-memory pressure with stored evidence;
- CPU-only and partial-rank coverage;
- null versus measured-zero handling;
- fixed 156-column geometry and plain/ANSI parity;
- assertions that Watch never emits step or performance-attribution language;
- parity between the System/Process panes rendered by Run and Watch.

## Validation

```text
  tests/reporting/summary tests/docs/test_terminal_summary_docs.py \
  tests/reporting/test_view.py -q
# 352 passed (one existing pynvml deprecation warning)
```


Pendu:
Reviewed at bb47adb36. Verification first: the summary, view, torch-free e2e, and display suites all pass at head (886 tests), lint is clean at the pinned versions, and I audited the merge resolution directly: the pre-1.7 re-render guard in view/command.py survived, and evidence.py is a lossless refactor of run_evidence.py with the type helpers moved to common.py. I also replayed stored watch summaries from real single-GPU and 4x T4 runs through the new card: all render the 156-column layout with no step-time language, and real system findings (HIGH GPU PWR on a busy box, LOW GPU UTIL on an idle one) now reach the section headings, which the old card never surfaced. A sweep for consumers of the deleted watch.py and run_evidence.py found zero dangling references.

Three small fixes worth making before merge, none of which touches the design:

1. The header can invent rank coverage. `_rank_coverage_text` falls back to bare `world_size` when `global_ranks_used` is missing (card.py:112-113), and on watch the Process section is the only rank source since the step samplers are run-only. With a degraded Process section the header reads `4 ranks` while the Process pane on the same card reads `NO DATA · 0/4 ranks`. The docstring already states the right rule ("without inventing missing counts"); returning None, or `?/4 ranks`, would honor it.

2. Watch multi-ness ignores nodes. `resolve_multi_process(run_meta, (process,))` at card.py:413 consults only the Process section, while `build_system_pane` computes its own node count at system.py:238. A 2-node watch whose Process section degraded renders no node coverage in the header, `2/2 nodes` in the System heading, and can print `N0`/`N1` tokens with the `Scope:` legend suppressed. The converse also ships: the `watch_cpu_only_partial` golden carries the legend with zero scope tokens on the card. Passing `(process, system)`, or gating the legend on whether a pane actually emitted a token, closes both directions.

3. The watch fallback card is empty. `build_fallback_card` skips the verdict for watch (card.py:482-484), so the shutdown-time fallback prints a title and a footer and nothing else. It only fires when rendering raises, but when it does, a line with the System/Process statuses, or an explicit "summary rendering failed, see final_summary.json", would keep it truthful.

One-line hardening: evidence.py:263-264 calls `str()` on `note`/`summary`, so a non-string value renders a Python repr into the card, and this path is now shared by run and watch. An isinstance guard covers it.

Two questions, both product calls rather than defects:

1. The watch card loses the "Watch monitors host and process health only; it does not measure step time" boundary and the `Next: traceml run <your-script>.py` pointer, and both are now banned by the golden test at test_card_golden.py:4278-4297, while the card doubles from 78 to 156 columns. The PR body's "Watch boundary preserved" section reads as keeping that boundary statement, so I want to confirm the deletion is intentional: it removes the watch-to-run conversion path, and 156 columns no longer fits a default 80-column terminal.

2. `card_profile_from_text` (card.py:528-533) scans the first four lines of the stored text, which include the user-controlled run name, so a run named "TraceML Watch Summary" re-renders as a watch card under `traceml view` (view/command.py:75-80 has no equality guard, unlike colorize_stored_card). Low likelihood, but the durable fix is storing the profile in the payload meta rather than sniffing text, which is a schema-versioned change, so your call on timing.

Not this PR's bug, noted for completeness: a watch run's final_summary.json and HTML banner still headline INSUFFICIENT_STEP_TIME_DATA while this card correctly says nothing about step time. That is the standing promotion question and can stay with #362.

abhinavsriva:
We made one intentional change: Watch now uses the same 156-column
System/Process card as Run. This keeps the output consistent across
single-process, multi-process, and multi-node runs.

Watch still reports resource health only and does not explain step time. It
also keeps a simple next step for users:
`Next: Wrap your step with trace_step(model), then traceml run.`

The wider layout is an accepted trade-off. Narrow-terminal support can be
handled separately if needed.

Issue #325, opened 2026-08-04
Live-surface test debt: dashboard section update paths + one real NiceGUI server-start test


Pendu:
Part of the 0.3.7 testing program (#322). The CLI sparse-state
renderer tests deleted in #302 (tests/renderers/test_step_time_sparse_ui.py,
186 lines) are restored under #326; this issue covers what never existed.

## Problem

The dashboard sections other than the combined Step Time hero
(system, process, GPU gauges, step memory) have only build-time smoke
coverage: nothing exercises their update paths, the code that runs on every
2s tick with live payloads. And no test ever starts the NiceGUI server: the
driver tests pin the port-conflict short-circuit only, so a template/wiring
regression that breaks page assembly ships silently.

## Proposal

1. Update-path tests for the sections that carry real conditional logic
   (staleness, missing-GPU, per-rank rows), in-process against the real
   section update functions; scoped, not exhaustive.
2. ONE server-start test in the existing integration job (the dashboard
   extra is already installed there): start the driver on port 0, poll until
   ready with a deadline (no fixed sleeps), fetch the page, assert all five
   section headers are present, shut down cleanly. Enters at nightly if it
   shows any timing sensitivity; per-PR after two weeks green.

## Explicit non-coverage

This does NOT guard #315-class layout/viewport regressions: in #315 the
sections were present in the DOM but unreachable at ~1280x720. That class
stays with the manual two-viewport visual pass in the release checklist
(see the umbrella's traceability table).

PR #387, opened 2026-08-21
Dashboard: escalate a server that never binds to ERROR with diagnostics; add real server-start test

Pendu:
## What

- `start()` already probes the socket and returns within 10 s. When that probe times out, a daemon watchdog now keeps probing for a grace period (60 s). A late bind logs INFO ("ready after Ns"); a server thread that exits without binding, or one still not listening after the grace, logs ERROR with diagnostics: port, thread alive, lifespan state, nicegui/uvicorn versions, and the server thread's live Python stack.
- The "Dashboard ready" line now names the serving stack versions.
- New `watch_server_startup()` in `server_readiness.py`, same injected-clock shape as `wait_for_server_ready()`, same foreign-listener guard.
- Tests: four pure-function tests for the watchdog outcomes; a driver test that forces the never-binds path and asserts the ERROR names the stuck function from the captured stack; a real server-start test that launches the driver in a child process on a free port and asserts HTTP 200 (the NiceGUI server-start test asked for in #325).

## Why

NiceGUI prints its "ready" banner from the ASGI lifespan, which runs before the socket is bound, so a dashboard that never binds looks healthy on stdout. The driver logged that case once at WARNING, and the aggregator's error log only keeps ERROR, so a run on a GPU box that printed the banner and never listened left empty logs behind. This change does not claim to know why that box failed to bind; it makes the next occurrence self-diagnosing (the stack shows where the server thread is stuck) and gives CI a floor against the whole class.

## Scope

Touches `display_drivers/nicegui.py` and `display_drivers/server_readiness.py` only (outside `nicegui_sections/`, as a serving-floor fix, no representation change). No behavior change on the healthy path beyond the extra versions in the ready line.

## Verification

- `pytest tests/display tests/core/test_packaging_constraints.py tests/runtime/test_launcher.py`: 137 passed (py3.12, nicegui 3.12.1).
- New test file on py3.13 + nicegui 3.16.0 and py3.12 + nicegui 3.16.0: 14 passed each, including the real bind.
- black / isort / ruff clean at line length 79.
- On a fresh DLAMI g4dn.xlarge (py3.13.13, nicegui 3.16.0, uvicorn 0.47.0 + uvloop): `traceml run --mode dashboard` with default flags bound in 1-6 s on 4/4 launches, HTTP 200. The original box failure does not reproduce on the AMI.

## Evidence

- GATE: conformance ran=true, `tests/integrations/test_telemetry_conformance.py` 1 passed / 0 failed / 1 skipped on this branch (the change touches no stream or payload).
- TRANSITIONS: tested timeout->ready-late, timeout->thread-died, timeout->still-not-listening with injected clocks (`test_server_readiness.py`), plus the forced never-binds path end to end through `start()` (`test_nicegui_server_start.py`).
- RANKS: EVIDENCE: n/a (serving-floor change; no per-rank rendering or aggregation is touched).
- TRIPWIRE: yes. Forced the watchdog to fire on purpose (server stub sets the lifespan flag and never binds, grace 1 s) and asserted the ERROR carries the stuck function's name from the captured stack.
- PLATFORM: EVIDENCE: n/a (no platform branch added; the diff has no `sys.platform` / `os.name` condition).

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui.py
@@ -32,6 +32,11 @@ Important reliability notes
 - Locks are held only for snapshots / swapping payload dicts:
   - We do NOT compute while holding the lock.
   - We do NOT update UI while holding the lock.
+- NiceGUI prints its "ready" banner from the ASGI lifespan, which runs
+  BEFORE the socket is bound. A server that never binds therefore looks
+  healthy in stdout. start() probes the socket, and if the bind is not
+  confirmed in time a watchdog keeps probing and logs an ERROR with the
+  server thread's stack, so the failure reaches the error log.
 
 Aggregator contract (required by trace_aggregator)
 --------------------------------------------------
@@ -42,6 +47,8 @@ Aggregator contract (required by trace_aggregator)
 
 from __future__ import annotations
 
+import importlib
+import sys
 import threading
 import time
 import traceback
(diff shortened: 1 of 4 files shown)


abhinavsriva (approved):
Still unnresolved startup diagnostics during shutdown for runs shorter than 70 seconds.

PR #388, opened 2026-08-21
Dashboard: context strip (run identity + observed coverage); watch profile without step panels

Pendu:
# Dashboard: context strip (run identity + observed coverage), watch profile without step panels

> **Review scope: the context strip only.** Everything below it in the screenshots (System, Process, GPU gauge, Step Time, Step memory, Diagnostics) is today's rendering, untouched by this PR and still work in progress; those blocks are rebuilt one PR at a time after this one. Please read the top strip and treat the rest of the page as background.

## What

- New `nicegui_sections/context_section.py`, a build/update pair like every other section. The strip replaces the old header and shows: the profile (`run` / `watch`), run name and script basename from the launcher manifest, a strategy chip (DDP / FSDP; nothing when the detector recorded no classification), a liveness word derived from data age (`live · updated 1s ago` / `stale · 47s since last data`, green / red dot), a coverage line (`ranks 3/4 reporting · 4 GPUs observed · 1 node · 2m 26s`), and the path to the run's `final_summary.json`.
- `pages.py` composes by profile: a watch session gets System | GPU gauge and Process, and no step-time hero, step-memory or diagnostics panes (those have no data under watch and rendered as permanently dead panels, #355). Run pages keep today's layout under the strip.
- A small `renderers/context/` package feeds the strip through its own layout key: `ContextDB.fetch_context` reads the existing tables (system, process, runtime environment), `ContextComputer` wraps it with the same stale-fallback the other computers use, and `ContextRenderer` is registered in the NiceGUI driver next to System / Process / Step memory, so `CONTEXT_LAYOUT` is computed on every tick in both profiles and `context_section.py` subscribes to it directly. Nine observed facts: world size, node GPU count and hostname from the newest system sample; ranks reporting; node count; GPUs observed across nodes; training strategy; first and last sample clocks. Every value is observed from the database, never copied from configuration; each query degrades on its own so an older database never blanks the strip. The System payload is untouched (its `ctx` keeps the three keys it had before).
- The rank rule is on clocks: a rank counts as reporting while its newest sample is within three ticks of the newest sample anywhere, with the tick estimated from the fastest rank's own cadence (floored at the default 2 s). Not on `seq`: on a 2-node, 3-hour capture the two ranks' seq counters ended 10 apart while their clocks ended 0.4 s apart, so a seq rule calls a live rank dead. `world_size` only ever appears as the denominator.
[...]

PR #389, opened 2026-08-21
Dashboard: System block as four tiles, two charts, and per-GPU rows that open on spread

Pendu:
# Dashboard: System block as four tiles, two charts, and per-GPU rows that open on spread

> **Review scope: the System block only.** The context strip above it is #388, already merged. The Process, Step Time, Step memory and Diagnostics panes in the screenshots are today's rendering, untouched by this PR and still work in progress; they are rebuilt one PR at a time after this one. System and Process now sit side by side at equal width without height stretching; Process will match the System block's height once it takes the same skeleton in its own PR, so the pair is uneven in these screenshots on purpose. That half-width layout is also why the power chart will not hold 5 to 10 GPUs on one node; see the second open item under Scope.

## What

The System block is rebuilt to the design we settled on 19 and 20 August: four number tiles, two small charts, and per-GPU rows behind a disclosure that opens on its own when the GPUs disagree.

- Tiles, one plain number each. `gpu util` is the across-GPU average, shown as the median of the last ~100 samples (`avg of 4 GPUs`). `gpu mem` is the GPU with the most memory in use, against its capacity (`6.3 / 16.1 GB`, `max GPU`). `gpu temp` is the hottest GPU (`52 °C`, `max GPU`). `host ram` is used against total (`5.9 / 200 GB`). Levels always carry their denominator; rates name their estimator in words.
- Host CPU chart: a small zero-anchored series with its one number (the window median) in the label, gridlines at 0 / half / top, and the x axis labelled in wall-clock time. The ceiling adapts (4 / 10 / 20 ... 100 %) so a 2 % line stays visible. The label reads `host cpu util · avg across cores`, because this is `psutil.cpu_percent()`: the mean across logical cores, where 100 % is every core saturated. That qualifier matters on this page, since the Process block reports per-rank CPU from `psutil.Process.cpu_percent()`, which sums across cores and can exceed 100 %.
- GPU power chart: one trace per GPU, with two dashed reference lines. The board limit in red (`70 W limit`), read from the payload rather than a constant, and the lowest draw seen in the run in grey (`30 W lowest seen`), which on any run whose sampler was up before the first step is the idle draw. Between them is the band the GPUs actually work in, so a trace sagging toward the grey line reads as idling rather than as a small number. The floor line is dropped when it sits within 10 % of the limit, where it would say nothing. Red belongs to the limit line, so no GPU trace uses it. Any aggregate power cell is the max GPU, never a sum.
- Per-GPU rows (`gpu` · `util` · `power trend` · `mem GB` · `temp °C` · `W / limit`), one sparkline per GPU, collapsed by default. They open on their own when the across-GPU utilisation spread (max - min, window p95) crosses 20 points. The header states the utilisation range and nothing else: `4 GPUs · util 0 to 100% · click to close`, `4 GPUs · util 99 to 99% · click to open`; the tail always reflects the rows' real open state, which a click may have changed. The 20-point trigger lives in the code, not on the screen. The rows on the minority side of the split are tinted (the one busy GPU of four, or the one idle GPU of four). Opening happens on the rising edge only, so a reviewer who closes the rows is not fought every tick.
- No verdict words anywhere. The renderer's `Hot` / `Warm` temperature word is no longer shown, and there is no high-power chip: on T4 every healthy run we measured averages 89-98 % of the 70 W limit against the rule's 80 % bar, so that chip would be permanently on. Judgement comes from the diagnosis engine or it does not appear.
- The four tiles are one CSS grid of equal columns (four, dropping to two then one as the card narrows), and every tile always carries its qualifier line, so the boxes are identical in width and height on every host: measured 147x80 / 233x80 / 183x80 at 1400 / 1100 / 900 px viewports. Before this, a single-GPU host left the temperature tile's qualifier blank ("max GPU" says nothing with one GPU) so that box was a line shorter, and a narrow card wrapped the RAM tile onto its own row where it stretched to its cap.
- No GPU: on a CPU-only host every slot keeps its place and height (one shape to learn on every host); the three GPU tiles read `n/a` with `no GPU` under them (the same word the Process card uses, so one page does not spell absence two ways), the power-chart slot shows `no GPU reported` at chart height, the rows slot is the one line `per-GPU rows · no GPU`; the RAM tile and the CPU chart work as on any host. Unreported is never zero: a NULL column stays a dash, and the sampler's all-zero fallback row (what it writes when NVML fails on a device) is treated as unreported, so it cannot render as a `0 W limit` line, `0 °C` or `0.0 GB`; the limit line is the largest limit seen in the window, so one unreported tick does not make it flicker.
- One node per block. System telemetry is per machine. When the window holds more than one host, the block shows the lowest `node_rank` host alone and its header says `node 0 of 2`. Before this, a 2-node run rendered one CPU line that zig-zagged between the two hosts and merged both nodes' `gpu0` into one row. Per-node rows and a node selector are the platform's job; this keeps the single-node block honest until then.
- The GPU utilisation gauge is gone: the util tile plus the rows carry that headline, and the gauge showed the raw last tick, which reads `0 %` at the moment someone inspects a just-finished run (both campaign runs end on `gpu_util.now = 0.0` with `p50 = 100 / 25`).

### Two views of the same chart: the sample window, and the whole run

Both charts render in one of two modes, and the label always says which one is on screen. This is the part of the PR most worth a careful read, because it changes what the chart means, not just how it looks.

**The switch.** `run_span > window_span * 1.2` in `system_section.py`. If the run is more than 20 % longer than what the last 100 system samples cover, the charts draw the whole run. Otherwise the window already is the whole run and there is nothing to gain. Every run passes through both: it starts windowed and flips a few minutes in.

**Window mode** draws the raw samples, exactly as before, and names the real extent of the data it has: `last 41 s`, `last 2 min`. Nothing is smoothed or bucketed, because at 100 points there is nothing to smooth away.

**Whole-run mode** reads a second, decimated query per chart, capped at 120 points however long the run gets, so the payload does not grow with run length. The two charts summarise differently, on purpose:

- Host CPU is a **rolling mean** over a window that scales with the run (30 s / 1 min / 2 min / 5 min, about a fiftieth of the run, rounded to a step a person recognises), computed in SQL with a window function and then sampled every Nth row. A rolling mean over an hour-long run shows the slow drift; tumbling buckets of the same width show the same drift with a sawtooth on top, which reads as noise. The label spells the window out (`rolling 2 min`) rather than naming the technique. Peaks are not drawn, but they still set the axis ceiling, so a spike the mean smooths away can never be drawn off the top of the chart.
- GPU power is a **mean and a floor per bucket, per GPU**: the mean solid, the bucket's minimum faint beneath it. Mean says whether the GPU is being fed. The floor says whether it went idle inside that bucket, which is what a stall looks like in watts and what a mean alone hides.

**Why not the peak.** A peak trace saturates. On every GPU-busy capture we own the busy GPUs' p95 draw is 85.7 W, 86.8 W and 91.2 W against the same 70 W board limit (NVML reports instantaneous draw, which overshoots the cap briefly), while their means are 67.7 W, 68.3 W and 66.2 W. So a peak trace sits above the limit line on any healthy run, which leaves the limit line with nothing to be compared against, and it reads about the same for three unrelated runs. The three captures also show what idle looks like on this hardware: their lowest samples are 33.0 W, 30.1 W and 30.7 W, and on the 1-busy-of-4 captures the three idle GPUs sit flat at about 33 W. That is the value the floor trace and the floor line are measured against. We do not yet own an input-starved GPU capture, so the floor's ability to separate a stalling run from a healthy one is argued from that idle draw rather than measured against a stalling run; the capture that would settle it is on the list.

The x axis of both charts is labelled in wall-clock time (`18:38`, and `HH:MM:SS` under ten minutes), because that is what logs and the Process block beside it are keyed on, and because a whole-run view spanning hours is not locatable otherwise. The two charts also share one anchor and one span, so a vertical read across the pair lands on the same moment in both. Without that they drift apart by a window, since the rolling mean drops its first partial window and the power buckets do not.

The run's length is not repeated on the charts: the context strip states it once (`ranks 1/1 reporting · 1 GPU observed · 1 node · 1h 36m`), so the chart labels name the view (`whole run`, `last 2 min`) and not the duration.

The window could not show the thing that justified having these charts. Host CPU on the 96-minute reference run climbs from 25.2 % to 32.1 % (+6.9 points, +27 % relative), and on the 3-hour two-node capture from 82.7 % to 92.0 %, comparing the first and last 5 % of samples. A 3-minute window renders both as a flat line at the final value.

Cost, measured as medians of 40 renders per fixture, CPU time rather than wall clock, expressed as a share of one core at the 2 s render interval, measured on the same machine with nothing else running, before (`642cbc1`, window only) against after (this head):

| fixture | before | after | payload before | payload after |
|---|---|---|---|---|
| 4-GPU, 40 s | 0.66 % | 0.88 % | 3.2 KB | 4.3 KB |
| 4-GPU, 23 min | 0.37 % | 0.37 % | 9.3 KB | 24.5 KB |
| 1-GPU, 96 min | 0.13 % | 0.42 % | 6.5 KB | 14.3 KB |
| 2-node, 3 h | 0.20 % | 0.89 % | 6.5 KB | 13.8 KB |

So at most about +0.7 points of a single core, on the longest capture we own. Browser main-thread time is unchanged within noise (75.2 ms per tick before, 66.2 ms after, 20 ticks each via CDP `Performance.getMetrics`), because the charts only re-send when the window moves. Nothing here runs per rank or per step, so the training path the overhead budget covers is untouched.

## Why

One number hid the failure we exist to find: with one GPU of four busy, every surface read ~25 % and nothing on the page said that three GPUs were idle, although the payload computed the spread (`gpu_delta`) every tick and rendered it nowhere. The average stays because it describes the node; it is safe only because the rows open at the same moment. Power was the one system metric missing from the dashboard and the only one that moves on every run we measured, so it gets the chart; everything else sits still for whole runs and gets a number. The per-metric reasoning (which estimator, why a number or a chart) is written up per metric in the block's design record; this PR description is the short form of it.

## Scope

Data path: `system_section` <- `SYSTEM_LAYOUT` <- `SystemRenderer.get_dashboard_renderable` <- `SystemDashboardComputer` <- `system_samples` + `system_gpu_samples`. Everything new reads System tables only and rides the System payload; the section formats and does not query; no other section's payload is read (the strip keeps its own `CONTEXT_LAYOUT`, #388).

`nicegui_sections/`: `system_section.py` (rewritten), `theme.py` (span-axis chart builders, reference-line builder, block CSS), `pages.py` (hero row, then System | Process side by side at exactly half the width each with `items-start`, so heights are never stretched to match; no gauge), `context_section.py` (section list without the gauge).

Outside the directory, stated on purpose because the rebuild rule is to change representation and not data: the dashboard payload grows by six fields, all additive, no existing key changes.

- `rollups.gpu_power = {now, p50, limit, floor}`: max across GPUs at the newest tick, median of that over the window, the largest positive limit reported anywhere in the window, and the lowest draw seen anywhere in the run (the minimum across every whole-run bucket minimum), which is what the grey reference line draws. Each is `None` when nothing reports it; with no power at all the chart hides and the label reads `gpu power · not reported`.
- `rollups.gpus[]`: per GPU, `gpu_idx`, `util_now`, `util_p50`, `mem_used`, `mem_total`, `temp`, `power`, `power_limit` from the newest tick, plus util as the window median. A GPU missing from the newest tick keeps its row with `None` values rather than vanishing.
- `rollups.gpu_mem.total`: capacity of the GPU shown in `gpu_mem.now`.
- `series.gpu_power[]`: `{gpu_idx, values}` aligned with `x_time`; a tick without a row for that GPU is `None` (a gap), never `0`.
- `rollups.ctx.system_node = {hostname, node_rank, nodes_in_window}` and a `hostname=` filter on `SystemMetricsDB.fetch_recent_system_samples`, used only when a window contains more than one named host (rows without a hostname are not a node of their own).
- `series.cpu_run = {t, avg, max, span_s, window_s}`: the whole-run CPU view from `fetch_cpu_run_history`, a rolling mean and rolling max over `window_s` computed with SQL window functions (`AVG/MAX OVER (ORDER BY sample_ts_s ROWS BETWEEN n PRECEDING AND CURRENT ROW)`, SQLite 3.25 and newer, which the packaging test already requires), then sampled every Nth row to at most 120 points. Partial windows at the start of the run are excluded rather than drawn short.
- `series.gpu_power_run = [{gpu_idx, t, avg, min, max, span_s, window_s}]`: the whole-run power view from `fetch_gpu_power_run_history`, disjoint buckets of `window_s` grouped per GPU, carrying each bucket's mean, minimum and maximum, likewise capped at 120 buckets. Both reads take only System tables and both follow the node scope above, so a two-host capture charts one machine rather than a blend, and both are skipped entirely while a run is still short enough to be drawn from its window.

Estimator per item, so the two surfaces can be compared later:

| item | form | estimator |
|---|---|---|
| gpu util tile | number | across-GPU mean per tick, window median (p50) |
| gpu mem tile | number + denominator | newest tick, GPU with max used, its own total |
| gpu temp tile | number | newest tick, max across GPUs |
| host ram tile | number + denominator | newest tick |
| host cpu chart, window mode | series + number | raw samples; label = window median |
| host cpu chart, whole run | series + number | rolling mean over `window_s`; axis ceiling from the rolling max |
| gpu power chart, window mode | per-GPU series | raw samples per GPU |
| gpu power chart, whole run | per-GPU series x2 | per bucket, mean (solid) and minimum (faint), per GPU |
| power reference lines | two dashed lines | limit = max reported in the window; floor = minimum across all bucket minima |
| rows: util | number | per-GPU window median |
| rows: mem / temp / W | numbers | per-GPU newest tick |
| rows trigger | hidden | `gpu_delta` window p95 > 20 points |

The rows header used to say `1 of 4 GPUs busy, 3 idle`, counting a GPU as idle below 20 % utilisation. That bar was invented in the display layer, and it disagreed with the diagnosis engine, whose `GPUUtilizationBands` calls anything under 30 % low. One page would have carried two thresholds and a word this layer has no standing to say, so the header now reports the range it observes and leaves the verdict to the engine. The trigger that opens the rows is unchanged and stays off the screen.

Three things for review rather than decided here:

1. Trigger statistic. p95 of the spread over the window opens the rows about five ticks (10 s) after one GPU goes idle, which is the responsiveness we want; the cost is that an 11-second, 4-rank FSDP run opened them on `spread 30 > 20` from its own start-up ramp (a 6-tick window, so p95 is the max). Now that the header reports the range rather than a verdict, that capture reads `4 GPUs · util 98 to 100% · click to close`: open rows beside a range that gives the reader no reason for them, which is the clearest argument yet for the median. A window median would ignore ramps but take ~100 s to react to a GPU going idle. The code carries p95; say the word and it is one constant.
2. Known limit, many GPUs on one node: in whole-run mode the power chart draws two traces per GPU (mean and floor) and the rows one line per GPU. Every fixture we own is 1 or 4 GPUs; at roughly 5 to 10 GPUs on a node the chart gets cluttered (traces overlapping in one 150 px slot at half the page width), and past that it is not usable as drawn. We know it and it needs a revision, not a fix in this PR. Candidates in the order we would try them: give System and Process the full page width one below the other instead of the side-by-side half each; aggregate the chart (a band for the GPU range plus the max trace) and leave per-GPU detail to the rows; make the traces selectable; or move the many-GPU form to its own scrollable page, which the design record already anticipates for the 32 to 100 GPU case. Layout only, nothing about what the block computes; decide it with a real many-GPU capture in front of us.
3. Cross-surface power: the run card sums power across GPUs (`cli_compute.py`, `power_total`); the dashboard shows the max GPU and per-GPU values. The card's form is untouched here; which one the card should follow is part of the open card-vs-dashboard estimator question.

Not in this PR: the least-headroom refinement of the memory tile (open), per-node rows (platform lane), the driver's step-time refresh under watch (#355 compute side).

## Verification

- `pytest tests/display tests/renderers tests/reporting tests/aggregator tests/core/test_packaging_constraints.py tests/integrations/test_telemetry_conformance.py`: 629 passed, 1 skipped, on `version_0.3.7` at `b05fb73` (which includes #388).
- 17 tests in `tests/display/test_system_section.py`: span words, denominators, disclosure text against the trigger and against the rows' real open state, rising-edge opening, power axis bounds, CPU axis ceilings, sparkline gaps, the rows table with the odd-ones-out tint and dashes for absent values, no GPU trace in the limit's red, every tile keeping a qualifier line, the rolling window spelled out rather than named, the two whole-run charts sharing one clock axis, both power reference lines with their positions, and build + update without a browser (CPU-only state, a user-closed disclosure, a vanished limit, an all-unreported tick, the no-resend guard, the node header).
- 12 tests in `tests/renderers/test_system_dashboard_gpu_payload.py`: 1-busy-of-4 reads `p50 = 25` with `delta p95 = 100`, all-busy reads spread 0, a missing GPU row is a `None` gap, no-GPU and unreported-power payloads, empty database schema, a two-node window keeps node 0 only, NULL util and temp stay `None`, the sampler's zero fallback counts as unreported, rows without a hostname are not a node of their own.
- Replays of every fixtured scenario cell through the live dashboard (gallery below), plus live local sessions through the real launcher: a CPU-only single-process run captured in both modes (windowed at `last 51 s`, then whole run at `rolling 30 s`), a CPU-only two-process run, and an unnamed watch session.
- The chart formatters are JavaScript strings, so they were executed rather than eyeballed: the axis formatter returns `14:00` at the span start and `14:59` at the newest sample, and the hover formatter returns `14:00 · 59 min ago`, `14:58 · 45 s ago` and `14:59 · now` for the same three positions.
- Charts only re-send when the window moved (the UI timer runs faster than the data arrives), and the reference lines are cleared explicitly when a limit disappears, since ECharts merges options rather than replacing them.
- black / isort / ruff clean at line length 79, the repo's pinned black applied. The five ruff findings elsewhere in `tests/` predate this branch and are left alone.

## Scenarios (screenshots)

Each image is the System block alone, the only block this PR changes; the rest of the page is today's rendering and out of scope here.

**B · B_xf_1gpu_long** · 96 min, 15k steps, the metrics-tour run; CPU drift visible  
`System / GPU UTIL / 99% / 1 GPU / GPU MEM / 6.3 / 16.1 GB / used / total / GPU TEMP / 48 °C / 1 GPU / HOST RAM / 2.8 / 16.6 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 2 MIN / 32% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 2 MIN / per-GPU rows / 1 GPU · click to open`  
B B_xf_1gpu_long

**B · B_cnn_1gpu** · 4 min cnn, 1000 steps  
`System / GPU UTIL / 93% / 1 GPU / GPU MEM / 4.6 / 16.1 GB / used / total / GPU TEMP / 47 °C / 1 GPU / HOST RAM / 3.1 / 16.6 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 27% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 30 S / per-GPU rows / 1 GPU · click to open`  
B B_cnn_1gpu

**C · C_M1_bert** · bert-base, 220 steps, COMPUTE-BOUND, 1 rank on a 4xT4 node  
`System / GPU UTIL / 25% / avg of 4 GPUs / GPU MEM / 9.1 / 16.1 GB / max GPU / GPU TEMP / 49 °C / max GPU / HOST RAM / 5.9 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 2 MIN / 2% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 2 MIN / per-GPU rows / 4 GPUs · util 0 to 100% · click to close`  
C C_M1_bert

**C · C_sysB_1of4** · 19-Aug campaign run B: 1 busy GPU of 4, 800 steps; the aggregate-hides-idle case  
`System / GPU UTIL / 25% / avg of 4 GPUs / GPU MEM / 6.3 / 16.1 GB / max GPU / GPU TEMP / 52 °C / max GPU / HOST RAM / 5.9 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 2% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 30 S / per-GPU rows / 4 GPUs · util 0 to 100% · click to close`  
C C_sysB_1of4

**D · D_S2_ddp4** · 4-rank DDP, healthy, 40 s  
`System / GPU UTIL / 97% / avg of 4 GPUs / GPU MEM / 1.9 / 16.1 GB / max GPU / GPU TEMP / 44 °C / max GPU / HOST RAM / 7.7 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 41 S / 8% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 41 S / per-GPU rows / 4 GPUs · util 97 to 97% · click to open`  
D D_S2_ddp4

**D · D_F1_fsdp4** · 4-rank FSDP (strategy chip)  
`System / GPU UTIL / 100% / avg of 4 GPUs / GPU MEM / 1.0 / 16.1 GB / max GPU / GPU TEMP / 41 °C / max GPU / HOST RAM / 8.0 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 12 S / 7% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 12 S / per-GPU rows / 4 GPUs · util 98 to 100% · click to close`  
D D_F1_fsdp4

**D · D_sysA_ddp4_3k** · 19-Aug campaign run A: 4-rank DDP, 3000 steps, 23 min  
`System / GPU UTIL / 100% / avg of 4 GPUs / GPU MEM / 6.3 / 16.1 GB / max GPU / GPU TEMP / 53 °C / max GPU / HOST RAM / 8.8 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 8% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 30 S / per-GPU rows / 4 GPUs · util 100 to 100% · click to open`  
D D_sysA_ddp4_3k

**D' · Dd_S2_rank3_dead** · S2 with rank 3's last 10 ticks removed  
`System / GPU UTIL / 97% / avg of 4 GPUs / GPU MEM / 1.9 / 16.1 GB / max GPU / GPU TEMP / 44 °C / max GPU / HOST RAM / 7.7 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 41 S / 8% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 41 S / per-GPU rows / 4 GPUs · util 97 to 97% · click to open`  
D' Dd_S2_rank3_dead

**E · E_xf_2x1_long** · 2 nodes x 1 GPU, 178 min, 30k step rows; the seq-drift case  
`System / node 0 of 2 / GPU UTIL / 100% / 1 GPU / GPU MEM / 6.3 / 16.1 GB / used / total / GPU TEMP / 45 °C / 1 GPU / HOST RAM / 2.9 / 16.6 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 5 MIN / 92% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 5 MIN / per-GPU rows / 1 GPU · click to open`  
E E_xf_2x1_long

**E · E_cnn_2x1_fsdp** · 2 nodes x 1 GPU FSDP, 5 min  
`System / node 0 of 2 / GPU UTIL / 90% / 1 GPU / GPU MEM / 4.6 / 16.1 GB / used / total / GPU TEMP / 46 °C / 1 GPU / HOST RAM / 3.1 / 16.6 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 65% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 30 S / per-GPU rows / 1 GPU · click to open`  
E E_cnn_2x1_fsdp

**E' · Ed_2x1_rank1_dead** · 2x1 with node 1's rank cut 10 ticks early  
`System / node 0 of 2 / GPU UTIL / 90% / 1 GPU / GPU MEM / 4.6 / 16.1 GB / used / total / GPU TEMP / 46 °C / 1 GPU / HOST RAM / 3.1 / 16.6 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 65% / GPU POWER · PER GPU VS 70 W LIMIT · WHOLE RUN · MEAN AND FLOOR OF EVERY 30 S / per-GPU rows / 1 GPU · click to open`  
E' Ed_2x1_rank1_dead

**W · W_W1_busy** · watch, 4 GPUs, 1 busy (98/0/0/0)  
`System / GPU UTIL / 25% / avg of 4 GPUs / GPU MEM / 0.8 / 16.1 GB / max GPU / GPU TEMP / 51 °C / max GPU / HOST RAM / 5.0 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 2 MIN / 2% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 2 MIN / per-GPU rows / 4 GPUs · util 0 to 100% · click to close`  
W W_W1_busy

**W · W_W2_idle** · watch, 4 GPUs, all idle  
`System / GPU UTIL / 0% / avg of 4 GPUs / GPU MEM / 0.6 / 16.1 GB / max GPU / GPU TEMP / 39 °C / max GPU / HOST RAM / 5.0 / 200 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 2 MIN / 0% / GPU POWER · PER GPU VS 70 W LIMIT · LAST 2 MIN / per-GPU rows / 4 GPUs · util 0 to 0% · click to open`  
W W_W2_idle

**A · live, CPU only, 1 process, window mode** · the first minutes of any run, before it outlives its window
`System / GPU UTIL / n/a / no GPU / GPU MEM / n/a / no GPU / GPU TEMP / n/a / no GPU / HOST RAM / 6.9 / 17.2 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 51 S / 50% / GPU POWER / no GPU reported / per-GPU rows · no GPU`
A cpu windowed

**A · the same live run, eight minutes in, whole-run mode** · same session, after the switch
`System / ... / HOST CPU UTIL · AVG ACROSS CORES · WHOLE RUN · ROLLING 30 S / 42% / GPU POWER / no GPU reported / per-GPU rows · no GPU`
A cpu whole run

**A2 · live, CPU only, 2 processes**
`System / GPU UTIL / n/a / no GPU / GPU MEM / n/a / no GPU / GPU TEMP / n/a / no GPU / HOST RAM / 7.2 / 17.2 GB / used / total / HOST CPU UTIL · AVG ACROSS CORES · LAST 38 S / 52% / GPU POWER / no GPU reported / per-GPU rows · no GPU`
A2 cpu 2 processes

## Evidence

GATE: conformance ran=true, `tests/integrations/test_telemetry_conformance.py` 1 passed / 0 failed / 1 skipped on this branch.
TRANSITIONS: tested closed->open on the rising edge of the spread (`test_rows_open_on_the_rising_edge_only`; the 1-busy-of-4 replays open reading `util 0 to 100%`, the healthy 4-rank replays stay closed reading `util 99 to 100%`), window mode -> whole-run mode on one live session (the same CPU run captured at `last 51 s` and then at `whole run · rolling 30 s`), GPU -> no GPU on the same panel (`test_section_builds_and_updates_without_a_browser` and the live CPU sessions), and single-node -> multi-node window (`test_two_node_window_shows_the_leader_node_only`; the 2x1 replay reads `node 0 of 2`).
RANKS: per-GPU asymmetry tested? yes: 100/0/0/0 reads `25 % avg` with the rows open and gpu0 tinted (unit test plus three replays); a GPU absent from one tick is a gap in its trace, not a zero (`test_missing_gpu_row_is_a_gap_not_a_zero`). Rank death does not touch this block, since the system sampler outlives the rank, which the D' and E' replays confirm.
TRIPWIRE: made the spread trigger fire on purpose? yes, the 1-busy-of-4 payload in the tests and the M1, sysB and W1 replays; and the inverse, all-busy (3000 steps) and healthy DDP stay collapsed at spread 0 and 1. The two-mode switch was also fired both ways rather than assumed: the same live run was captured on both sides of it.
SCENARIOS: A single-node CPU 1 process ok, captured live in both modes. A2 single-node CPU 2 processes ok, captured live. B single-node 1 GPU ok (96-min and 4-min runs: `99 %` / `93 %`, whole-run charts with `rolling 2 min` / `rolling 30 s`). C single-node 4 GPUs 1 process ok (`25 %`, rows open, `util 0 to 100%`; one capture windowed at `last 2 min`, one whole run). D single-node 4 GPUs 4 ranks ok (`97 %` and `100 %`, both closed with a narrow range; the 12-second FSDP capture opened on its start-up ramp, see the first open item under Scope). D' one rank dead ok, block unchanged. E 2 nodes x 1 GPU ok and honest (`node 0 of 2`, no pooled series, 178-min capture on `rolling 5 min`); E' 2 nodes with one rank dead ok. F 2 nodes x N GPUs: no capture exists. W watch ok (busy: `25 %`, rows open, gpu0 `69 / 70` W; idle: `0 %`, closed). Wc watch on CPU with no run name: covered by the two live CPU-only captures above at this head; the earlier separate watch capture predates the chart changes and is not shown stale.
PLATFORM: EVIDENCE: n/a (no `sys.platform` / `os.name` branch in this change).

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/context_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/context_section.py
@@ -38,10 +38,9 @@ STEP_SECTIONS: Tuple[str, ...] = (
     "step_memory",
     "model_diagnostics",
 )
-RESOURCE_SECTIONS: Tuple[str, ...] = ("gpu_gauge", "system", "process")
+RESOURCE_SECTIONS: Tuple[str, ...] = ("system", "process")
 RUN_SECTIONS: Tuple[str, ...] = (
     "model_combined",
-    "gpu_gauge",
     "system",
     "process",
     "step_memory",
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/pages.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/pages.py
@@ -32,12 +32,7 @@ from .step_memory_section import (
     build_step_memory_section,
     update_step_memory_section,
 )
-from .system_section import (
-    build_gpu_gauge_section,
-    build_system_section,
-    update_gpu_gauge_section,
-    update_system_section,
-)
+from .system_section import build_system_section, update_system_section
 
 
 def _cell(flex: str):
@@ -76,66 +71,47 @@ def define_pages(cls):
 
             hero_cards = None
             if "model_combined" in shown:
-                # Row 1: hero (step-time ribbon + verdict) | GPU gauge
+                # Row 1: hero (step-time ribbon + verdict). A watch session
+                # has no step loop, so no hero: the resource panes lead.
                 with (
                     ui.row()
                     .classes("w-full items-stretch")
                     .style("gap:16px; flex-wrap:wrap;")
                 ):
-                    with _cell("2.4"):
+                    with _cell("1"):
                         hero_cards = build_model_combined_section()
                         cls.subscribe_layout(
                             MODEL_COMBINED_LAYOUT,
                             hero_cards,
                             update_model_combined_section,
                         )
-                    with _cell("1"):
-                        gauge_cards = build_gpu_gauge_section()
 
-                # Row 2: System | Process
-                with (
-                    ui.row()
-                    .classes("w-full items-stretch")
-                    .style("gap:16px; flex-wrap:wrap;")
-                ):
-                    with _cell("2"):
-                        system_cards = build_system_section()
-                    with _cell("1.3"):
-                        cards = build_process_section()
-                        cls.subscribe_layout(
-                            PROCESS_LAYOUT, cards, update_process_section
-                        )
-            else:
-                # A watch session has no step loop, so no hero: the
-                # resource panes take the page. Row 1: System | GPU gauge,
-                # Row 2: Process.
-                with (
-                    ui.row()
-                    .classes("w-full items-stretch")
-                    .style("gap:16px; flex-wrap:wrap;")
-                ):
-                    with _cell("2"):
-                        system_cards = build_system_section()
-                    with _cell("1"):
-                        gauge_cards = build_gpu_gauge_section()
-                with (
-                    ui.row()
-                    .classes("w-full items-stretch")
-                    .style("gap:16px; flex-wrap:wrap;")
-                ):
-                    with _cell("1"):
-                        cards = build_process_section()
-                        cls.subscribe_layout(
-                            PROCESS_LAYOUT, cards, update_process_section
-                        )
+            # System | Process: the evidence pair (the node, its ranks),
+            # side by side at exactly half the width each. Heights are not
+            # stretched to match (items-start): the two blocks share one
+            # skeleton (a tiles row, fixed-height chart slots, one
+            # disclosure), so they are equal by construction while their
+            # disclosures are closed, and an opened disclosure grows only
+            # its own card. Process adopts the skeleton in its own PR; until
+            # then it is today's card. The System block carries the GPU
+            # headline itself (util tile + per-GPU rows), so there is no
+            # gauge.
+            with (
+                ui.row()
+                .classes("w-full items-start")
+                .style("gap:16px; flex-wrap:wrap;")
+            ):
+                with _cell("1 1 0"):
+                    system_cards = build_system_section()
+                with _cell("1 1 0"):
+                    cards = build_process_section()
+                    cls.subscribe_layout(
+                        PROCESS_LAYOUT, cards, update_process_section
+                    )
 
-            # One SYSTEM_LAYOUT subscriber drives the chart and the gauge
-            # (two subscribers on one layout/client would evict each other).
-            def _update_system(_c, d, _sc=system_cards, _gc=gauge_cards):
-                update_system_section(_sc, d)
-                update_gpu_gauge_section(_gc, d)
-
-            cls.subscribe_layout(SYSTEM_LAYOUT, system_cards, _update_system)
+            cls.subscribe_layout(
+                SYSTEM_LAYOUT, system_cards, update_system_section
+            )
 
             # Row 3: Step Memory | Diagnostics (training runs only)
             if "step_memory" in shown or "model_diagnostics" in shown:
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
@@ -249,10 +249,6 @@ def odd_ones_out(gpus: List[Dict[str, Any]]) -> set:
 # wire; carrying it would let this read "avg of 10 cores".)
 CPU_LABEL = "host cpu util · avg across cores"
 
-# A run longer than the window by this factor is charted whole rather than
-# as its last 100 samples; below it, the window IS the whole run.
-RUN_VIEW_FACTOR = 1.2
-
 # Absent value marker. "n/a" rather than a dash: the Process card on the
 # same page already reads "N/A", and one page should not spell absence two
 # ways.
@@ -561,10 +557,10 @@ def update_system_section(panel: Dict[str, Any], data: Dict[str, Any]) -> None:
     run_avg = run.get("avg") or []
     run_max = run.get("max") or []
     run_span = float(run.get("span_s") or 0.0)
-    cpu_whole = len(run_t) > 2 and run_span > span * RUN_VIEW_FACTOR
+    cpu_whole = len(run_t) > 2
     prun = series.get("gpu_power_run") or []
     prun_span = float(prun[0].get("span_s") or 0.0) if prun else 0.0
-    power_whole = bool(prun) and prun_span > span * RUN_VIEW_FACTOR
+    power_whole = bool(prun)
     # The two whole-run charts sit one above the other, so they share one
     # anchor and one span: a dip in power and a rise in CPU at the same x
     # are then the same moment. Their own spans differ by a window (the
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
@@ -84,10 +84,6 @@ def format_span(seconds: Optional[float]) -> str:
     return f"last {seconds / 60.0:.0f} min"
 
 
-# Per-GPU utilisation below this reads as idle in the rows' header.
-IDLE_UTIL_PCT = 20.0
-
-
 def format_gb_pair(used_bytes: Any, total_bytes: Any) -> Tuple[str, str]:
     """A level against its capacity: ('6.3', '/ 16.1 GB')."""
     used = theme.gb(used_bytes) if used_bytes is not None else None
@@ -101,15 +97,16 @@ def format_gb_pair(used_bytes: Any, total_bytes: Any) -> Tuple[str, str]:
     return (num, f"/ {total_s} GB")
 
 
-def disclosure_text(
-    gpus: List[Dict[str, Any]], *, over: bool, is_open: bool
-) -> str:
-    """Header of the per-GPU rows, in GPU words: what the user needs to know.
+def disclosure_text(gpus: List[Dict[str, Any]], *, is_open: bool) -> str:
+    """Header of the per-GPU rows: the utilisation range, and nothing else.
 
-    ``over`` is the trigger's verdict (the across-GPU spread crossed the
-    bar); the words say what that means on this node, never how it was
-    computed. The tail follows the rows' real state, which a click may have
-    changed: the rows open on their own only on a rising edge.
+    It states what the GPUs read, low to high, and leaves the reading to
+    the person or to the diagnosis engine. An earlier version said "1 of 4
+    GPUs busy, 3 idle" off a 20 % bar invented here, while the engine calls
+    anything under 30 % low (``GPUUtilizationBands``): one page, two
+    thresholds, and a word this layer had no standing to say. The trigger
+    that opens the rows is unchanged and stays off the screen. The tail
+    follows the rows' real state, which a click may have changed.
     """
     utils = []
     for g in gpus:
@@ -124,15 +121,8 @@ def disclosure_text(
     tail = " · click to close" if is_open else " · click to open"
     if n == 1:
         return "1 GPU" + tail
-    if over:
-        idle = sum(1 for u in utils if u < IDLE_UTIL_PCT)
-        if idle:
-            head = f"{n - idle} of {n} GPUs busy, {idle} idle"
-        else:
-            head = "uneven load across GPUs"
-    else:
-        head = f"all {n} GPUs alike"
-    return head + tail
+    lo, hi = min(utils), max(utils)
+    return f"{n} GPUs · util {lo:.0f} to {hi:.0f}%" + tail
 
 
 def node_scope_text(ctx: Dict[str, Any]) -> str:
@@ -824,6 +814,6 @@ def update_system_section(panel: Dict[str, Any], data: Dict[str, Any]) -> None:
         panel["rows"].value = True
     panel["_over"] = over
     panel["rows_hint"].text = disclosure_text(
-        gpus, over=over, is_open=bool(panel["rows"].value)
+        gpus, is_open=bool(panel["rows"].value)
     )
     panel["rows_html"].content = rows_html(gpus, pseries, spread=spread)
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
@@ -733,17 +733,26 @@ def update_system_section(panel: Dict[str, Any], data: Dict[str, Any]) -> None:
             refs = []
             if limit is not None:
                 refs.append(
-                    (float(limit), f"{float(limit):.0f} W limit", _LIMIT_RED)
+                    (
+                        float(limit),
+                        f"{float(limit):.0f} W limit",
+                        _LIMIT_RED,
+                        "insideEndTop",
+                    )
                 )
             floor_w = gp.get("floor")
             if floor_w is not None and (
                 limit is None or float(floor_w) < float(limit) * 0.9
             ):
+                # Opposite corner from the limit, and below its own line,
+                # where the chart is empty: the two labels would otherwise
+                # stack at the right edge and the longer one be cut off.
                 refs.append(
                     (
                         float(floor_w),
                         f"{float(floor_w):.0f} W lowest seen",
                         _FLOOR_GREY,
+                        "insideStartBottom",
                     )
                 )
             lines[0]["markLine"] = (
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
@@ -1,28 +1,340 @@
-"""System metrics — CPU/GPU utilization time-series + a GPU-util gauge (PR2).
+# Copyright 2026 OptAI UG (haftungsbeschraenkt)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0
 
-Two subscribers on the same System payload (the driver's multi-subscriber
-registry): the dual-line chart card and the gauge card. Payload unchanged.
+"""
+System block: the host and its GPUs, as four tiles, two charts and rows.
+
+Tiles (one plain number each): GPU utilisation = across-GPU average, a
+short-window median; GPU memory = the max-used GPU against its capacity;
+GPU temperature = the hottest GPU; host RAM = used against total. Levels
+always carry their denominator, rates carry the estimator in words.
+
+Charts: GPU power per GPU against the board limit (power is the one system
+metric that moves on every run), and host CPU as a small zero-anchored
+series (its shape drifts over long runs). Both label their gridlines and
+their window span, and neither narrates its mechanism.
+
+Rows: one line per GPU behind a disclosure that opens on its own when the
+across-GPU utilisation spread crosses ``SPREAD_EXPAND_PTS``. The average
+describes the node; the rows are what make it honest when one GPU of four
+is busy, so the two are one design, not two.
+
+No verdict words anywhere: judgement comes from the diagnosis engine or it
+does not appear.
 """
 
 from __future__ import annotations
 
+import math
 from datetime import datetime
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from nicegui import ui
 
 from . import theme
 
+# Across-GPU utilisation spread (max - min, percentage points, window p95)
+# above which the per-GPU rows open on their own. Measured separation on
+# the reference runs is 0 (all busy) vs 100 (one busy of four).
+SPREAD_EXPAND_PTS = 20.0
 
-def _to_ms(s: str) -> Optional[int]:
-    try:
-        return int(datetime.fromisoformat(s).timestamp() * 1000)
-    except Exception:
-        return None
+_GPU_COLORS = (
+    "#f97316",
+    "#3b82f6",
+    "#10b981",
+    "#a855f7",
+    "#0d9488",
+    "#0ea5e9",
+    "#eab308",
+    "#ec4899",
+)
+_LIMIT_RED = "#dc2626"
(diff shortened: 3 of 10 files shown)


abhinavsriva:
Implementation update, this supersedes the corresponding details in the PR description:

- The 20% whole-run switch now lives in `SystemDashboardComputer`, not `system_section.py`.
- Short runs still perform a lightweight span check, but skip expensive whole-run aggregation.
- CPU-only runs do not query GPU power history.
- Whole-run history is capped at 120 points.
- The visible power wording is now “average and lowest”.

abhinavsriva (approved):

PR #393, opened 2026-08-25
Replace row-based summary windows with time-based history retention

### LINKED ISSUE #391: Align final-summary sections to one step-derived time window
## Problem

`--summary-window-rows` applies the same row count to Step Time, Step Memory,
System, and Process data even though those tables have different sampling
rates. Equal row counts therefore describe different portions of a run and can
also make a bounded telemetry interval look like the full run duration.

## Production change

This issue intentionally makes a breaking configuration change:

- Remove `--summary-window-rows` and `TRACEML_SUMMARY_WINDOW_ROWS` without a
  deprecated alias.
- Add `--history-retention DURATION`, `history_retention`, and
  `TRACEML_HISTORY_RETENTION`; row counts are not converted to durations.
- Resolve one final-report window from common completed Step Time steps inside
  the selected duration.
- Query System and Process inside the shared timestamp bounds and Step Memory
  inside the shared step bounds.
- Prune Step Time and Step Memory through the minimum of their highest aligned
  steps and the highest aged step. Prune periodic telemetry through that
  step's timestamp.
- Drop later arrivals at or before an already deleted step or timestamp before
  insertion.
- Calculate top-level run duration only from lifecycle timestamps. Leave it
  null when lifecycle duration is unavailable.

Runs without step telemetry use a receive-time fallback for periodic history.
Runtime metadata is never pruned.

## Acceptance criteria

- Every final-summary section uses the same analysis start and end bounds.
- System and Process may contain different row counts from Step Time.
- `duration_s` never substitutes a bounded analysis interval for full runtime.
- SQLite history remains bounded and pruning uses indexed columns.
- The removed row-based options are documented as a breaking change with the
  new duration-based replacements.
- Regression coverage verifies the shared frontier, periodic timestamp
  deletion, runtime-metadata preservation, and rejection of known-old arrivals.

## Out of scope

- Changing live dashboard windows.
- Interpolating metrics onto step boundaries.
- Long-term downsampling, rollups, or a size-based retention policy.

abhinavsriva:
# Replace row-based summary windows with time-based history retention

Closes #391. Source branch: `dev/abhinav/issue391`.

## What changed

- Replaced `--summary-window-rows` with `--history-retention DURATION` across
  CLI, YAML, environment settings, manifests, and Ray configuration. Durations
  accept seconds, minutes, hours, or days; bare numbers mean seconds.
- Removed the 10,000-row defaults, 1.5x retention multiplier, row-limited
  report queries, and old compatibility fields.
- Added a 30-minute history window. Step Time and Step Memory advance through
  one shared aligned step frontier; periodic telemetry follows its timestamp.
- Later arrivals at or before a deleted step or timestamp are dropped before
  insertion.
- Added indexed, transactional pruning for raw System, GPU, Process, Step
  Time, Step Memory, and stdout/stderr history. Runtime metadata is preserved.
- Final reports now resolve one window from common completed optimizer steps:
  System and Process use the same timestamps, while Step Memory uses the same
  step range. Runs without Step Time use a time-only telemetry window.
- Added `analysis_window` with shared bounds, per-section sample counts, and
  observed timestamp bounds.
- Top-level `duration_s` now comes only from real training lifecycle
  timestamps; it stays `null` when they are unavailable.
- Bumped the final-summary schema to 1.8 and updated relevant documentation.
- Finalization now refreshes summaries created earlier by `traceml.summary()`.
  Atomic replacement preserves the previous JSON if regeneration fails.

## Why

Step and periodic telemetry have different sampling rates, so the same row
count represented different parts of a run. That misaligned section coverage
and could incorrectly present a short bounded telemetry window as the full run
duration.

Time-based retention gives every section one understandable interval while
allowing each source to retain its natural sample count.

## Breaking change

`--summary-window-rows` and `TRACEML_SUMMARY_WINDOW_ROWS` are removed without a
deprecated alias. Use `--history-retention`, `history_retention`, or
`TRACEML_HISTORY_RETENTION`. Row counts are not converted to durations because
the tables have different sampling rates.

## How I tested

- Regression coverage for pruning every raw-history table while preserving
  runtime metadata.
- A 500-step report with different periodic sample counts, aligned bounds,
  Step Memory alignment, and correct duration.
- Finalization tests for stale-summary replacement and failure preservation.
- End-to-end launcher smoke test.
- Full suite and repository checks pass.

- [x] Tests added or updated
- [x] Existing tests run
- [x] Manual smoke test

## Runtime impact

- [x] May affect telemetry storage and final reporting

Pruning runs at most once per minute and once at finalization. Reads and
deletes remain bounded by indexes. Live dashboard windows are unchanged.

## Docs

- [x] Updated

Updated CLI/API, Ray configuration, reporting schema, reading-output, and
architecture documentation. README and Quickstart are unchanged.

## Next work (separate PRs)

- Roll up/summarize older telemetry before deleting raw rows, providing
  longer-term history without retaining full-resolution data.
- Add a byte/size-based safety cap in addition to time retention, especially
  for high node counts.
- Benchmark and design the 1,000-10,000-node storage path, including possible
  lower-frequency or pre-aggregated telemetry transport.

This PR intentionally implements only the bounded raw-history foundation; it
does not add rollups, downsampling, or size-based retention.


Pendu:
Reviewed and tested this one closely, since it changes retention semantics and a public contract at the same time. The core is solid. I pushed one small fix and have one question that is yours to decide.

## Verified

- Full suite reproduces your number exactly: `1331 passed, 2 skipped`.
- Pruning is correct. `_HISTORY_TABLES` covers exactly the six raw-telemetry tables, `runtime_environment` is not in that tuple and is never deleted, so the rank-context contract still holds. Table names come from a constant tuple and are rejected before interpolation, the cutoff is parameterised, and the transaction ownership flag means a failing prune rolls back through the caller rather than half-applying.
- No dangling references to the deleted `reporting/config.py` anywhere in `src/` or `tests/`.
- Schema 1.8 is consistent: I generated a summary and compared it against SCHEMA.md, and the `analysis_window` keys, per-section observation keys, and coverage states all match what is documented. Every consumer of top-level `duration_s` guards for `None` before doing arithmetic, so making it nullable does not break the terminal card, the HTML report, or the SDK projection.
- `mkdocs build --strict` exits 0. black and codespell are clean. The two ruff findings are in `tests/`, which the ruff hook does not cover, and both already exist on `version_0.3.7`, so they are not from this PR.

I also ran a second independent review with a different model to cross-check the areas where this PR carries the most risk. It converged on the same single defect below and otherwise agreed that pruning, the window maths, and the schema are clean. Its one high-severity claim, that a long prune would block summary readers until the 5 second SQLite timeout, does not hold: the writer sets `journal_mode=WAL`, and with a writer holding an open transaction containing a 150k-row DELETE, a default-timeout reader still completed its SELECT in 0.005s against the pre-DELETE snapshot. Readers do not wait on writers in WAL.

## Pushed (`1954f2c`)

`examples/README.md:101` still used the removed flag, so the documented comparison command failed with `unrecognized arguments: --summary-window-rows`. It is the only hard break left in the repo. I dropped the flag rather than substituting a duration: the flag was there to keep summaries comparable across machines, and the new window resolves from common completed optimizer steps with the same default everywhere, so comparability no longer needs an explicit value.

## Question: the removal has no deprecation path

`--summary-window-rows` and `TRACEML_SUMMARY_WINDOW_ROWS` ship in released v0.3.6, so anyone upgrading with either in a script or a launcher config gets an argparse hard failure rather than a warning. Our own stability rule is that a public option gets a migration path and a deprecation notice before it disappears, and the PR description takes the opposite position deliberately, so I am not going to overrule it in a review.

Worth an explicit call: either accept the clean break for 0.3.7 and put it in the release notes as a breaking change, or keep `--summary-window-rows` for one release mapped onto `--history-retention` with a single stderr deprecation warning. I lean towards one release of the alias, only because the failure mode for a user is a broken command rather than a degraded result, but it is a judgement call and either is defensible.

## Two optional minors

- `RETENTION_TABLES` is now dead code. It is still defined in `sqlite_writers/system.py:47`, `process.py:55`, `step_time.py:56`, and `step_memory.py:42`, with no remaining consumers after the move to `_HISTORY_TABLES`.
- Five test fixtures still pin `"schema_version": 1.7` while production emits 1.8: `tests/reporting/test_view.py:85`, `tests/reporting/summary/test_step_time_nullable.py:194`, `tests/sdk/test_summary_client.py:123`, and `tests/step_time/test_contract_baseline.py:583-584`. They pass because nothing rejects 1.7, so this is only about the fixtures no longer representing what the code produces.

Neither blocks. Happy with this once you have made the deprecation call.

Pendu:
One more minor finding, from a second pass over the coverage logic. I missed it in the review above.

**`partial` coverage is reported when nothing inside the window was actually deleted.**

`_coverage_status` (`src/traceml_ai/reporting/analysis_window.py:241-248`) tests the retention ledger against the window's start only, with no upper bound:

```python
overlap = row[1] is not None and int(row[1]) >= window.start_step
# and, for time-based sections
overlap = row[0] is not None and window.start_ts_s is not None and float(row[0]) >= window.start_ts_s
```

So a deletion that falls entirely above the window still sets `overlap`. SCHEMA.md:76-77 defines the state as "retention deleted data that overlaps the requested interval", and a deletion above `end_step` / `end_ts_s` does not overlap it.

Reproduced directly against the real function:

```
window steps 13..19 ; max_deleted_step=20 (above window)  -> 'partial'
window steps 13..19 ; max_deleted_step=15 (inside window) -> 'partial'   (correct)
window ts 100..200  ; max_deleted_ts=999  (above window)  -> 'partial'
```

A realistic trigger is a rank that stops early. If rank B dies at step 10 and rank A keeps producing steps 11 onward alone, those steps are never common, so the window stays anchored at the last common step. When rank A's solo rows later age past the retention horizon and are pruned, `max_deleted_step` climbs above the window end and step_time reports `partial` for the rest of the session, even though every row inside the window survived. The ledger keeps a running maximum, so the label does not clear.

The direction is safe: this over-reports `partial` and cannot produce a false `complete`, since a deletion inside the window always satisfies the same test and a missing ledger row yields `unknown`. So it is diagnostic noise after a rank crash rather than a data problem, which is why I would not hold the PR for it.

Fix is a judgment call and yours: either bound the overlap test by the window end as well, or leave the conservative behaviour and reword SCHEMA.md so `partial` means "retention deleted data at or after the window start". The current code and the current wording disagree, so it is worth picking one.

Pendu:
Fixed the coverage bug in `17bff40`.

The obvious fix does not work. Bounding the existing test by the window end would introduce a false `complete`, which is worse than the noise it removes. Deletion is a prefix in receive time, but coverage is asked about step and sample-time bounds, and late arrivals mean those two orders differ. A maximum alone therefore cannot separate "deleted only above the window" from "deleted across and above it": both give `max_deleted > end`. A prune that removes early steps and one late-arriving high step lands exactly there, and an upper-bound test would have reported `complete` while real rows inside the window were gone.

So the ledger now records the whole deleted interval rather than only its top:

- `history_retention_state` gains `min_deleted_sample_ts_s` and `min_deleted_step`.
- `_prune_history_table` selects `MIN` alongside `MAX` and accumulates it across prunes with the same NULL-safe pattern already used for the maximums.
- `_coverage_status` now tests interval intersection against both window bounds, via a small `_intervals_overlap` helper.

The helper fails safe: no deletion means no overlap, but any bound it cannot establish is treated as overlapping, so a gap degrades to `partial` rather than claiming coverage the ledger cannot support. Older databases without the new columns make the SELECT raise, which the existing handler already turns into `unknown`, so the legacy path is unchanged.

Verified by execution, covering both directions:

```
deleted steps 20..20, window 13..19 (above)          -> complete
deleted steps  1..5,  window 13..19 (below)          -> complete
deleted steps 15..20, window 13..19 (spans into it)  -> partial
deleted steps 14..16, window 13..19 (inside)         -> partial
time-based, deleted ts 999, window 100..200          -> complete
time-based, deleted ts 150..999, window 100..200     -> partial
```

End to end against the real pruner, seeding steps 1..30 plus a late-arriving row for step 99 with an old receive time: one prune records the interval as steps 1..99, a window of steps 12..20 still correctly reports `partial`, and a second prune keeps the minimum at 1 while accumulating the row count. That is the case a max-only ledger would have gotten wrong.

Two tests added in `tests/reporting/summary/test_final.py`, one per direction, plus the existing retention fixture updated for the new columns. Full suite is `1333 passed, 2 skipped`, and ruff, black, isort and codespell are clean at the pinned versions.

This leaves the deprecation question from the earlier review as the only open item.

abhinavsriva:
@Pendu   Addressed the review feedback. The old row-based options are intentionally removed as a documented breaking change. Retention now uses one shared aligned step frontier, and the obsolete grace, coverage ledger, interval tracking, and dead constants were removed. Known-old arrivals are dropped before insertion. The PR description, issue, schema, docs, and tests are updated.

PR #400, opened 2026-08-26
fix: preserve unavailable process metrics as null

### LINKED ISSUE #394: Define normalized aggregator export records for training telemetry
## Goal

Define a small, versioned export contract that the aggregator can derive from telemetry it already receives. This contract is for external exporters only; it must not replace the current rank payloads, SQLite schemas, dashboard, terminal, or diagnosis inputs.

Export three record kinds:

- `step_timing`: step wall/data wait plus forward, backward, optimizer, and optional H2D timing;
- `step_memory`: per-step peak allocated/reserved device memory;
- `runtime_context`: rank identity, topology, distributed backend, and training strategy when available.

## What already exists / reusable

- Rank and sampler metadata: `telemetry/envelope.py` and `runtime/identity.py`.
- Timing fields: `StepTimeSampler` and `samplers/schema/step_time_schema.py`.
- Memory fields: `StepMemorySampler` and `samplers/schema/step_memory.py`.
- Workload context: `RuntimeEnvironmentSampler` and `runtime/environment.py`.
- The aggregator already normalizes and dispatches these payloads before SQLite ingestion.

## Exact changes

1. Add a transport-independent `TelemetryExportRecordV1` model in a leaf telemetry module. Common fields include schema version, record kind, timestamp, session/run id, global/local rank, world sizes, node rank, hostname, pid, and sampler source.
2. Add typed payloads for `step_timing`, `step_memory`, and `runtime_context`. Use milliseconds for durations and bytes for memory. Missing measurements remain null/absent, never zero-filled.
3. Add a pure aggregator-side mapper from normalized existing envelopes to export records. Map each incoming row independently; do not join timing and memory or query SQLite.
4. Keep diagnosis, severity, recommendations, renderer fields, and SQLite-specific ids out of the export contract.
5. Document field names, units, availability semantics, and example records. Add mapper and serialization round-trip tests.

## Code to delete, merge, deprecate, or move

- Delete nothing in this issue.
- Do not change rank-side schemas or the TCP/msgpack protocol.
- Keep event-name parsing and legacy-field normalization inside the mapper so exporters receive stable named fields.

## Dependencies

None. This is the first issue in the OTLP feature sequence.

## Acceptance criteria

- [ ] Existing rank telemetry maps deterministically to versioned export records.
- [ ] Step timing, step memory, runtime context, and distributed identity are represented with documented units.
- [ ] Missing signals remain unavailable rather than becoming zero.
- [ ] The model/mapper imports no exporter, SQLite, diagnosis, reporting, or UI code.
- [ ] Existing transport, SQLite, dashboard, terminal, diagnosis, and summary tests remain unchanged and green.

abhinavsriva:
## Changes

- Return `None` when CPU or RSS collection fails.
- Keep total RAM unknown when initialization fails.
- Preserve unknown CUDA availability and GPU count until CUDA can be safely probed.
- Update the Process sample schema to allow these nullable values.
- Add focused sampler tests.

## Scope

This PR changes upstream collection and wire semantics only. SQLite already preserves these values as `NULL`.

Legacy terminal, dashboard, and diagnosis handling of unavailable values is intentionally out of scope and will be addressed separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
           pytest \
             tests/arch \
             tests/config \
+            tests/samplers \
             tests/core \
             tests/step_time \
             tests/diagnostics \
--- a/src/traceml_ai/samplers/process_sampler.py
+++ b/src/traceml_ai/samplers/process_sampler.py
@@ -88,7 +88,7 @@ class ProcessSampler(BaseSampler):
         """
         Read total system RAM.
         """
-        self.ram_total = 0.0
+        self.ram_total: Optional[float] = None
         try:
             self.ram_total = float(psutil.virtual_memory().total)
         except Exception as e:
@@ -136,7 +136,7 @@ class ProcessSampler(BaseSampler):
         self.gpu_count: int = 0
         self.device_index: Optional[int] = None
 
-    def _sample_cpu(self) -> float:
+    def _sample_cpu(self) -> Optional[float]:
         """
         Return process CPU utilization as a percentage.
         """
@@ -148,9 +148,9 @@ class ProcessSampler(BaseSampler):
             )
         except Exception as e:
             self.logger.error(f"[TraceML] WARNING: CPU sample failed: {e}")
-            return 0.0
+            return None
 
-    def _sample_ram(self) -> float:
+    def _sample_ram(self) -> Optional[float]:
         """
         Return process resident memory (RSS) in bytes.
         """
@@ -160,7 +160,7 @@ class ProcessSampler(BaseSampler):
             )
         except Exception as e:
             self.logger.error(f"[TraceML] WARNING: RAM sample failed: {e}")
-            return 0.0
+            return None
 
     def _cuda_safe_to_touch(self) -> bool:
         """
@@ -235,12 +235,12 @@ class ProcessSampler(BaseSampler):
         try:
             gpu_metrics = self._sample_gpu()
 
-            gpu_available = (
-                bool(self.gpu_available)
-                if self.gpu_available is not None
-                else False
+            # ``None`` means CUDA has not been safely probed yet. Preserve
+            # that distinction instead of reporting a fabricated ``False``.
+            gpu_available = self.gpu_available
+            gpu_count = (
+                int(self.gpu_count) if self.gpu_available is not None else None
             )
-            gpu_count = int(self.gpu_count) if self.gpu_available else 0
 
             sample = ProcessSample(
                 sample_idx=self.sample_idx,
--- a/src/traceml_ai/samplers/process_sampler.py
+++ b/src/traceml_ai/samplers/process_sampler.py
@@ -140,12 +140,10 @@ class ProcessSampler(BaseSampler):
         """
         Return process CPU utilization as a percentage.
         """
+        if self.process is None:
+            return None
         try:
-            return (
-                float(self.process.cpu_percent(interval=None))
-                if self.process
-                else 0.0
-            )
+            return float(self.process.cpu_percent(interval=None))
         except Exception as e:
             self.logger.error(f"[TraceML] WARNING: CPU sample failed: {e}")
             return None
@@ -154,10 +152,10 @@ class ProcessSampler(BaseSampler):
         """
         Return process resident memory (RSS) in bytes.
         """
+        if self.process is None:
+            return None
         try:
-            return (
-                float(self.process.memory_info().rss) if self.process else 0.0
-            )
+            return float(self.process.memory_info().rss)
         except Exception as e:
             self.logger.error(f"[TraceML] WARNING: RAM sample failed: {e}")
             return None
--- a/src/traceml_ai/samplers/schema/process.py
+++ b/src/traceml_ai/samplers/schema/process.py
@@ -112,14 +112,14 @@ class ProcessSample:
     timestamp: float
     pid: int
 
-    cpu_percent: float
+    cpu_percent: Optional[float]
     cpu_logical_core_count: int
 
-    ram_used: float
-    ram_total: float
+    ram_used: Optional[float]
+    ram_total: Optional[float]
 
-    gpu_available: bool
-    gpu_count: int
+    gpu_available: Optional[bool]
+    gpu_count: Optional[int]
     gpu: Optional[ProcessGPUMetrics]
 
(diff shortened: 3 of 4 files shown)


Pendu:
Reviewed this one closely because it widens a wire contract. The change itself is correct and I could not break it: I tried to make the new nulls crash or corrupt a number anywhere downstream and failed in every direction I tried. One thing worth fixing before merge, one loose end in the same function, and two questions.

## Verified

- Full suite `1330 passed, 2 skipped` at head against `1329 passed, 2 skipped` at base, so exactly the one added test and no regressions.
- The new test pins the fix. Run against base it fails with `assert 0.0 is None`; at head it passes.
- The gpu truth table holds. Driving `_ensure_cuda_device` with a stubbed torch across all three states of `self.gpu_available`, base and head emit identical values for `False` and `True`, and only the unprobed `None` state changes, which is the intended fix. The `is not None` branch cannot leak a stale count because the probe assigns availability and count together and sets the count to 0 whenever availability is False.
- Nulls survive the whole path. I built a `ProcessSample` with all five fields `None` and pushed it through `to_wire()`, the real msgspec codec, `sqlite_writers/process.build_rows`, and into SQLite: the row comes back `{cpu_percent: None, ram_used_bytes: None, ram_total_bytes: None, gpu_available: None, gpu_count: None}`. No coercion back to zero.
- Every reader of `process_samples` is null-safe. `renderers/process/common.py` uses SQL `MAX`/`AVG`, which skip NULL; `reporting/sections/process/loader.py:126-128` maps each field with an explicit `is not None`; `renderers/step_memory/common.py:90` and `reporting/sections/step_memory/loader.py:64` skip a NULL `gpu_available` and fall through to `system_samples`, which still carries a real bool. `renderers/system/` reads `system_samples` only, so it is untouched.
- Backward compatibility across releases holds in the direction that matters. The released v0.3.5 and v0.3.6 aggregators already have 13 `else None` guards each in the process writer and zero `NOT NULL` constraints on these five columns, so an upgraded rank sending nulls to an older aggregator lands as NULL rather than failing.
- The change aligns the sampler with the contract already written in `reporting/SCHEMA.md:9-14`: null means never measured, a measured zero stays `0.0`.
- The step-memory "no GPU" path still works. Both `renderers/step_memory/common.py:90` and `reporting/sections/step_memory/loader.py:64` reduce `gpu_available` with `MAX()` and then branch on `is False`, so an all-NULL `process_samples` could in principle flip that message. It does not in a real run: both queries fall through to `system_samples`, `SystemSampler` is registered `rank_zero_only` and always writes a real bool, and a `--mode summary` run at head still prints `STEP MEMORY: NO GPU` with the torch-based-telemetry evidence line. The only window where it differs is before the first system sample lands, which affects the live panel text and not the final summary.
- `pre-commit run` at the pinned revs (isort 5.13.2, ruff 0.9.4, black 24.10.0, codespell 2.4.1) passes on all three files. `tests/arch` passes. A real `--mode summary` run at head still reports `CPU only (no GPU detected)` in the header, same as base.

## Worth fixing before merge: CI never runs this test

`tests/samplers/` is not referenced by any workflow. Every CI job enumerates explicit paths (`ci.yml:83-91`, `132-143`, `110`, `185-192`) and none runs a bare `pytest`, so the only test this PR adds never executes in CI. It passes locally, which makes it easy to miss.

One line fixes it, in the torch-free core job at `ci.yml:84`:

```
             tests/arch \
             tests/config \
+            tests/samplers \
             tests/core \
```

That job is the right home: the test imports only `ProcessSampler` and needs no torch.

## Same function, one branch left behind

`_sample_cpu` and `_sample_ram` have two failure paths and this PR migrated one. The `except` branch now returns `None`, but the `if self.process ... else 0.0` branch at `process_sampler.py:147` and `:159` still fabricates a zero. That branch is reached when `_init_process` could not attach psutil and set `self.process = None` at `:85`, which is the worse failure: nothing at all is known about the process, and it reports a confident 0.0% CPU and 0 bytes RSS for the entire run.

```
process = None (attach failed)  -> _sample_cpu() = 0.0   _sample_ram() = 0.0
process present, call raises    -> _sample_cpu() = None  _sample_ram() = None
```

Those lines are not in the diff, so this is not a regression. I am raising it because the PR body says "Return `None` when CPU or RSS collection fails" and this is the same condition three lines away, so it looks like the intended scope rather than a separate change.

## Two questions

**Torch-free installs now report null forever.** `_cuda_safe_to_touch` returns False unconditionally when `torch is None` (`:169-170`), so `_ensure_cuda_device` is never reached and `gpu_available` never leaves `None`. Base emitted `false`/`0` for every sample on that path; head emits `null`/`null` for the whole run. Nothing downstream breaks, because `detect_gpu_available` falls through to `system_samples`. But the new comment reads "CUDA has not been safely probed yet", and on a torch-free install it never will be, so "unknown" is permanent rather than transient. Is that the semantics you want, or should a torch-free install report a known `false`?

**How do you want to sequence the rest of the migration?** `system_sampler.py:115-116` still initialises `gpu_available = False` and `gpu_count = 0`, and `:85`/`:100` still zero-fill the core count and total RAM, so the two samplers now disagree about how an unavailable value is represented. `cpu_logical_core_count` is also still `int` in the same dataclass (`schema/process.py:116`) while its four neighbours became `Optional`. Given #394's criterion that missing signals stay unavailable rather than becoming zero, is System a follow-up PR, or deliberately left as is?

## Coverage note

Two of the five behaviours in the PR body are tested. A coverage run over the full suite shows `process_sampler.py` lines `234-260` never executed, which is all of `sample()` including the new `gpu_available` passthrough and the `gpu_count` ternary, and `94-95` never executed, which is the `_init_ram` failure path that sets `ram_total = None`. `ProcessSample` is not constructed by any test in the repo, and `tests/aggregator/test_tcp_wire_roundtrip.py` builds only `SystemSample` envelopes, so the wire semantics the PR headlines have no test crossing the wire. Not a blocker, and worth knowing that the three untested clauses are the ones a future regression would ship silently.

Happy with this once CI actually runs the test.

abhinavsriva:
Addressed the review feedback:
- Added tests/samplers to the existing torch-free core CI job.
- Updated the missing-process-handle path to return None instead of fabricated 0.0 CPU/RSS values.
- Added a focused regression test for that path.
System sampler consistency and downstream terminal/dashboard/diagnosis presentation remain intentionally out of scope for this upstream-only PR.

PR #401, opened 2026-08-26
Dashboard: Process block as four tiles, two per-rank charts and rank rows

Pendu:
# Dashboard: Process block as four tiles, two per-rank charts and rank rows

> **Review scope: the Process block only.** The System block above it is #389, already merged. Step Time, Step memory and Diagnostics in the screenshots are today's rendering, untouched here and rebuilt in later PRs. The block adopts the System block's skeleton (a tiles row, fixed-height chart slots, one disclosure) so the side-by-side pair is equal by construction with both disclosures closed.

## What

Every rank's trainer process, read on that rank's own clock.

- **Four tiles, each naming its estimator and, where one rank owns the number, that rank.** `cpu capacity` (worst rank, named, as a share of the host's capacity), `rss` (worst rank, named, used against total), `cuda reserved` (least-headroom rank, named, against its device), `cuda allocated` (median across ranks, which belongs to no single rank). A bottleneck hunt asks which rank is the problem, so three of the four tiles name one; the median across ranks is carried in the payload beside the worst for the card to compare against. Levels carry their denominator; rates name their estimator.
- **Two charts, one shared wall-clock axis.** Process CPU per rank, which sits directly under the System block's host CPU chart, and RSS per rank. Both switch from the raw sample window to a decimated whole-run view once the run outgrows the window, exactly as the System block does, and both label the span they actually hold.
- **Per-rank rows behind a disclosure**: rank, gpu, node, cpu capacity, RSS, a CPU trend sparkline, CUDA allocated, CUDA reserved, and the rank's age. They open on their own when the across-rank reserved imbalance crosses the diagnosis engine's own warn bar, on a rising edge only, so closing them is respected.
- **The block states its scope in the header**: `one process per rank · DataLoader workers not included`. Worker processes are invisible to this sampler by design and their cost lands in the System block, so host CPU high while every rank's process CPU is low means the work is outside the trainer. That reading is only available if the boundary is stated, and it is the pairing the two blocks exist to make possible.
- **No verdict words anywhere.** The rows header reports coverage and the imbalance number; whether that is bad is the engine's call.

## Why

Four things today's card gets wrong, all visible in the screenshots below.

1. **A dead rank froze the whole block.** The payload advanced on `min(latest seq across ranks)`, so one rank that stopped reporting pinned every other rank to that rank's last tick. On the rank-dead fixture the card stops at `last 12 samples` while three ranks keep reporting: the surface goes quiet exactly when a distributed job needs it. Ranks are now read independently and a stale rank is dimmed, counted in the header (`4 ranks · 1 stale, excluded`), and dropped from every aggregate so it cannot keep speaking through the tiles. A rank silent longer than the recent window keeps its row through a second one-row-per-rank read, so the block does not quietly forget it minutes later. When NO rank is reporting there is nobody to exclude them in favour of, so the last known numbers stand and the header says `none reporting` rather than claiming an exclusion that did not happen.
2. **CPU was unbounded.** `cpu_percent` from psutil sums cores, so the tile read `288 %` on the 2-node capture and `395 %` peak on the 23-minute one. It is now divided by the host's logical core count, which was already stored beside it and never carried into the payload.
3. **GPU memory read the newest raw sample of a sawtooth.** At 2 s the allocator's live bytes are undersampled: on the bert capture the window median is 1.66 GB, the window max 7.3 GB, and the newest sample 0.02 GB. The tile showed the 0.02. It is now the window median, and it is never called a peak, which is step memory's word.
4. **Imbalance was computed on allocated.** Across-rank allocated spread reaches 3 GB on perfectly healthy runs, because each rank's sampler catches a different step phase; that noise was the old `GPU IMBAL 1.71 GB` tile. Reserved is the stable quantity and the one the engine's own imbalance rule uses.

## Scope

`DATA PATH: process_section <- PROCESS_LAYOUT <- ProcessRenderer.get_dashboard_renderable <- ProcessDashboardComputer <- process_samples`

One table, one package. The section evaluates no threshold: which view the charts draw, which rank each tile speaks for, which ranks are stale, and whether the rows have earned opening are all decided in the computer. `structure_gate.py` reports no read-side violations and `tests/arch` passes.

The imbalance BAR is imported from `diagnostics/process/policy.py` rather than copied, so the threshold cannot drift. The imbalance METRIC is computed here, `(max-min)/max` over per-rank reserved medians, which mirrors the engine's `_imbalance_pct` but is a second implementation; the two could diverge and only the bar is shared. It is read at import, so a runtime-configured policy would not reach the dashboard.

Payload, all additive on the Process payload itself: `rollups.ranks[]` (identity, capacity, memory, age, stale), `rollups.cpu_capacity` / `rss` / `cuda` (each naming its rank), `rollups.reserved_imbalance_pct`, `rollups.rows_over`, `ranks_reporting` / `ranks_total` / `ranks_stale`, and `series.{cpu_capacity,rss}` plus their `_run` whole-run twins. `history` and `gpu_used_imbalance` keep their place in the dataclass for the CLI snapshot, which is untouched.

Outside the Process package, three changes:

- `nicegui_sections/theme.py`: the series helpers the System block already had (`apply_span_axis`, `shared_run_axis`, `format_span`, `format_gb_pair`, `sparkline_svg`, `cpu_axis_max`, and friends) moved here so both blocks use one implementation instead of two. `system_section.py` imports them back under the same names, so its call sites and tests are unchanged.
- Axis ticks now take their precision from the axis range. A chart fitted to a 20 MB drift was labelling a tick `1.402237892150879 GB`.
- `system_section.py` labels: the charts said `whole run`, which history retention (30 minutes by default since #393) makes false on any longer live run. Both blocks now state the span they actually hold (`last 22 min`), which is true under any retention setting. That also fixed a latent inconsistency: the two System charts share one axis but were quoting spans a rolling window apart.

## Estimator per item

| item | form | estimator |
|---|---|---|
| cpu capacity tile | number | per-rank window median, then the WORST of those, with its rank |
| cpu capacity chart | per-rank series | rolling mean over a run-scaled window (30 s to 5 min) |
| rss tile | number + denominator | worst live rank chosen on its window median, shown at its newest value |
| rss chart | per-rank series | rolling mean; axis fitted to the data, not zero-anchored |
| cuda allocated | number | per-rank window median (the sawtooth is undersampled) |
| cuda reserved | number + denominator | newest reported sample of the least-headroom rank |
| reserved imbalance | number | `(max-min)/max` over per-rank window medians |
| rank age | seconds | newest arrival clock anywhere minus this rank's; stale past three ticks |
| rows trigger | hidden | imbalance over the engine's warn bar, armed once every rank has held an allocation |

Three notes on those choices, each measured rather than argued.

**Liveness uses the aggregator's arrival clock, not the rank's wall clock.** Comparing `sample_ts_s` across machines makes one fast clock mark every other node dead. Measured skew on the 2-node capture is 11 ms under NTP, so it would not have bitten there, but arrival time is skew-free by construction.

**The rows trigger is a window median, not a streak of ticks.** A tick streak cannot debounce a replay: a finished run rendered once never accumulates one, so its rows would stay shut however lopsided the run was. The median debounces the staggered CUDA ramp instead, and the arming rule (every reporting rank must have held an allocation) keeps the first seconds of a run from reading as total imbalance. The 12-second FSDP capture, the one that opens the System block's rows on its start-up ramp, reads `reserved imbalance 1%` here and stays shut.

**The RSS axis is not zero-anchored.** Ranks sit around 1.5 GB and drift by tens of MB across a run. On a zero-anchored axis that movement is under one pixel of a 92 px chart, so the chart could not show the one thing it is for. The scenario matrix caught this; the axis now fits the data and the 3-hour capture reads `1.382 / 1.392 / 1.402 GB`.

## Verification

- `pytest tests/display tests/renderers tests/reporting tests/aggregator tests/arch tests/core/test_packaging_constraints.py tests/integrations/test_telemetry_conformance.py`: 669 passed, 1 skipped, on `version_0.3.7` at `5039e1b` (which includes #400).
- 29 new tests. 16 on the payload: a dead rank dims instead of stopping the block, stale ranks leave the aggregates, GPU tiles survive teardown rows, allocated is the window median, CPU is bounded by host capacity, imbalance reads reserved not allocated, the rows open only when armed and over the bar, the ramp alone never opens them, one series per chart and never both, the run history holds its 120-point budget, every degraded path keeps the series schema, a CPU-only host reports no GPU, a rank dead longer than the window keeps its row, the exclusion claim is only made when ranks still report, the nullable sampler columns from #400 read as absent rather than zero, and the payload drives the section end to end. 13 on the section: the scope line, the rows header without verdict words, a dimmed stale rank, absence rendered as `n/a` and never zero, per-rank trend colours matching the rows, no rank colour in the verdict red, the drift axis, tick precision from the range, rising-edge opening, and build plus update without a browser including the CPU-only state.
- Two defects were found by rendering rather than by tests, and both now have tests: the zero-anchored RSS axis, and the raw float on a fitted tick.
- Cost, medians of 40 renders per fixture, CPU time as a share of one core at the 2 s render interval, before against after:

| fixture | before | after | payload before | payload after |
|---|---|---|---|---|
| 4 ranks, 23 min | 0.03 % | 1.29 % | 62.3 KB | 49.4 KB |
| 2 nodes, 3 h | 0.08 % | 0.13 % | 62.0 KB | 24.1 KB |
| 1 rank, 96 min | 0.02 % | 0.05 % | 61.6 KB | 12.5 KB |
| 1 rank, 2.4 min | 0.01 % | 0.02 % | 22.5 KB | 6.2 KB |

The block does more than it did (per-rank medians, two whole-run histories, liveness, and a second read so a long-dead rank keeps its row) so it costs more. The worst case, 1.29 % of one core on the 4-rank capture, is above the System block's measured band of 0.37 to 0.89 %; it is 26 ms of CPU per 2 s tick on a 48-core host, and it is the number to argue with if the overhead budget is the concern. Two things keep it there: the recent-window read is bounded by the cadence the ranks are observed to sample at rather than reading back to the start, and the rolling histories, which do scan the retained table, refresh on their own 15 s clock rather than every tick, since a mean over minutes cannot change visibly between two 2 s ticks. The payload is smaller than before on every fixture measured, all of which are 1, 2 or 4 ranks; it grows with rank count, since both charts carry a line per rank. Nothing here runs per rank or per step in the training process.

- black / isort / ruff clean at 79 on every file this PR touches. The five ruff findings elsewhere in `tests/` predate this branch and are left alone.

## Known limitations, stated rather than discovered in review

- **The two blocks scope differently on multi-node captures.** System shows one node (`node 0 of 2`) because host telemetry is per machine; Process shows every rank across both nodes, since ranks are globally unique and each row carries its `N0`/`N1`. This is deliberate, but a reviewer will notice the difference.
- **The charts share a kind of axis, not a pixel grid.** Both blocks label wall-clock time and both decimate their own data, so their spans can differ by a few percent and the same pixel column is not exactly the same instant. Compare by the tick labels or the hover readout, which are exact.
- **The RSS chart shows drift within retained history, not necessarily the whole run.** On a healthy live run at the 30-minute default from #393, a leak developing over hours is only visible for its retained tail. The bound is not a plain clock cut: retention advances to a step frontier that requires every expected rank to have completed that step, so a straggler or a dead rank stalls pruning and history keeps growing past the window. The case where the history matters most is the case where it is kept, which is fortunate rather than designed. Longer retention or a decimated slow series is an aggregator change, not a display one.
- **Many ranks.** Every fixture we own is 1, 2 or 4 ranks. One line per rank and one row per rank will not read at 32 or more; the shape that will (a percentile band plus outlier ranks) needs a real capture at that scale before it is designed.

## Scenarios (screenshots)

One image per fixtured cell, cropped to the Process block, which is the only block this PR changes. The caption is the block's own rendered text, so a cell that reads wrong is visible without opening the image. The load-bearing ones are `Dd_S2_rank3_dead` and `Ed_2x1_rank1_dead`, where a rank stops and the block keeps going while naming it, and the two 2-node cells.

Every fixture cell is a REPLAY of an archived capture, most of them recorded before history retention existed, so their databases still hold the whole run and the labels read `last 172 min`, `last 93 min` and so on. A live run prunes as it goes, so the same charts there show whatever the retention window has kept. That is the reason the label states the span it holds instead of claiming the whole run.

The two live cells below were run through the real launcher on this machine and captured from their own sessions after they finished, so they show the whole-run view; the CPU-only state was also verified on the live dashboard while the run was in its window stage.

Each image is the Process block alone, the only block this PR changes; the rest of the page is today's rendering and out of scope here.

**B · B_xf_1gpu_long** · 96 min, 15k steps, the metrics-tour run; CPU drift visible  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 25.1% / worst rank · R0 · of host capacity / RSS / 1.4 / 16.6 GB / worst rank · R0 / CUDA RESERVED / 5.6 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 2.2 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 93 MIN · ROLLING 2 MIN / 25.1% / RSS · PER RANK · LAST 93 MIN · ROLLING 2 MIN / 1.4 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
B B_xf_1gpu_long

**B · B_cnn_1gpu** · 4 min cnn, 1000 steps  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 27.3% / worst rank · R0 · of host capacity / RSS / 1.7 / 16.6 GB / worst rank · R0 / CUDA RESERVED / 3.9 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 0.58 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 4 MIN · ROLLING 30 S / 27.3% / RSS · PER RANK · LAST 4 MIN · ROLLING 30 S / 1.7 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
B B_cnn_1gpu

**C · C_M1_bert** · bert-base, 220 steps, COMPUTE-BOUND, 1 rank on a 4xT4 node  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.6 / 200 GB / worst rank · R0 / CUDA RESERVED / 8.5 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 1.8 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 2 MIN / 2.1% / RSS · PER RANK · LAST 2 MIN / 1.6 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
C C_M1_bert

**C · C_sysB_1of4** · 19-Aug campaign run B: 1 busy GPU of 4, 800 steps; the aggregate-hides-idle case  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.3 / 200 GB / worst rank · R0 / CUDA RESERVED / 5.6 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 2.2 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 5 MIN · ROLLING 30 S / 2.1% / RSS · PER RANK · LAST 5 MIN · ROLLING 30 S / 1.3 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
C C_sysB_1of4

**D · D_S2_ddp4** · 4-rank DDP, healthy, 40 s  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.1 / 200 GB / worst rank · R1 / CUDA RESERVED / 1.3 / 15.6 GB / least-headroom rank · R2 / CUDA ALLOCATED / 0.86 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 41 S / 2.1% / RSS · PER RANK · LAST 41 S / 1.1 GB / per-rank rows / 4 ranks · reserved imbalance 16% · click to open / arrow_drop_down`  
D D_S2_ddp4

**D · D_F1_fsdp4** · 4-rank FSDP (strategy chip)  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.2 / 200 GB / worst rank · R0 / CUDA RESERVED / 0.29 / 15.6 GB / least-headroom rank · R3 / CUDA ALLOCATED / 0.04 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 12 S / 2.1% / RSS · PER RANK · LAST 12 S / 1.2 GB / per-rank rows / 4 ranks · reserved imbalance 1% · click to open / arrow_drop_down`  
D D_F1_fsdp4

**D · D_sysA_ddp4_3k** · 19-Aug campaign run A: 4-rank DDP, 3000 steps, 23 min  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.3 / 200 GB / worst rank · R2 / CUDA RESERVED / 5.6 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 2.2 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 22 MIN · ROLLING 30 S / 2.1% / RSS · PER RANK · LAST 22 MIN · ROLLING 30 S / 1.3 GB / per-rank rows / 4 ranks · reserved imbalance 0% · click to open / arrow_drop_down`  
D D_sysA_ddp4_3k

**D' · Dd_S2_rank3_dead** · S2 with rank 3's last 10 ticks removed  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 1.1 / 200 GB / worst rank · R1 / CUDA RESERVED / 1.3 / 15.6 GB / least-headroom rank · R2 / CUDA ALLOCATED / 0.86 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 41 S / 2.1% / RSS · PER RANK · LAST 41 S / 1.1 GB / per-rank rows / 4 ranks · 1 stale, excluded · reserved imbalance 16% · click to open / arrow_drop_down`  
D' Dd_S2_rank3_dead

**E · E_xf_2x1_long** · 2 nodes x 1 GPU, 178 min, 30k step rows; the seq-drift case  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 73.0% / worst rank · R1 · of host capacity / RSS / 1.5 / 16.6 GB / worst rank · R0 / CUDA RESERVED / 5.6 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 2.2 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 172 MIN · ROLLING 5 MIN / 73.0% / RSS · PER RANK · LAST 172 MIN · ROLLING 5 MIN / 1.5 GB / per-rank rows / 2 ranks · reserved imbalance 0% · click to open / arrow_drop_down`  
E E_xf_2x1_long

**E · E_cnn_2x1_fsdp** · 2 nodes x 1 GPU FSDP, 5 min  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 64.2% / worst rank · R1 · of host capacity / RSS / 1.7 / 16.6 GB / worst rank · R1 / CUDA RESERVED / 4.0 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 0.26 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 4 MIN · ROLLING 30 S / 64.2% / RSS · PER RANK · LAST 4 MIN · ROLLING 30 S / 1.7 GB / per-rank rows / 2 ranks · reserved imbalance 0% · click to open / arrow_drop_down`  
E E_cnn_2x1_fsdp

**E' · Ed_2x1_rank1_dead** · 2x1 with node 1's rank cut 10 ticks early  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 64.1% / worst rank · R0 · of host capacity / RSS / 1.7 / 16.6 GB / worst rank · R0 / CUDA RESERVED / 4.0 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 0.26 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 4 MIN · ROLLING 30 S / 64.1% / RSS · PER RANK · LAST 4 MIN · ROLLING 30 S / 1.7 GB / per-rank rows / 2 ranks · 1 stale, excluded · click to open / arrow_drop_down`  
E' Ed_2x1_rank1_dead

**W · W_W1_busy** · watch, 4 GPUs, 1 busy (98/0/0/0)  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 2.1% / worst rank · R0 · of host capacity / RSS / 0.66 / 200 GB / worst rank · R0 / CUDA RESERVED / 0.22 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 0.14 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 2 MIN / 2.1% / RSS · PER RANK · LAST 2 MIN / 0.66 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
W W_W1_busy

**W · W_W2_idle** · watch, 4 GPUs, all idle  
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 0.0% / worst rank · R0 · of host capacity / RSS / 0.63 / 200 GB / worst rank · R0 / CUDA RESERVED / 0.0 / 15.6 GB / least-headroom rank · R0 / CUDA ALLOCATED / 0.0 GB / median rank · live tensors / PROCESS CPU · CAPACITY PER RANK · LAST 90 S / 0.0% / RSS · PER RANK · LAST 90 S / 0.63 GB / per-rank rows / 1 rank · click to open / arrow_drop_down`  
W W_W2_idle

**A · live, CPU-only host, 1 rank** · run through the launcher on a laptop, 19 min; the GPU tiles hold their place and say so, and a 40 MB process reads as `0.04 GB` rather than `0.0`
`Process / one process per rank · DataLoader workers not included / CPU CAPACITY / 0.4% / worst rank · R0 · of host capacity / RSS / 0.04 / 17.2 GB / worst rank · R0 / CUDA RESERVED / n/a / no GPU / CUDA ALLOCATED / n/a / no GPU / PROCESS CPU · CAPACITY PER RANK · LAST 19 MIN · ROLLING 30 S`
A live cpu only

**A2 · live, CPU-only host, 2 ranks** · same workload with `--nproc-per-node 2`; the worst CPU rank and the worst RSS rank are different ranks, which is the point of naming them
`Process / ... / CPU CAPACITY / 0.4% / worst rank · R1 · of host capacity / RSS / 0.04 / 17.2 GB / worst rank · R0 / ... / per-rank rows / 2 ranks · click to open`
A2 live cpu two ranks

## Evidence

GATE: conformance ran=true, `tests/integrations/test_telemetry_conformance.py` 1 passed / 0 failed / 1 skipped on this branch.
TRANSITIONS: tested window view -> whole-run view (`test_one_series_per_chart_never_both` pins that a short run sends the window and a long one sends only the run view, with a database each; no fixture is replayed across the crossing itself), live rank -> stale rank (`test_a_dead_rank_dims_instead_of_stopping_the_block`, and the D' and E' replays render it), closed -> open rows on the rising edge with a reader's close respected (`test_rows_open_on_the_rising_edge_only` plus the section test), and GPU -> no GPU on the same panel (`test_a_cpu_only_host_reports_no_gpu`, `test_a_cpu_only_host_keeps_every_slot`).
RANKS: per-rank asymmetry tested? yes: every tile names the rank it speaks for, a stale rank is excluded from all of them while keeping its row, and the imbalance is computed across ranks on reserved. The 2-node replays show ranks from both nodes with `N0` / `N1` identity; the rank-dead replays show one dimmed row while the block keeps updating.
TRIPWIRE: made the rows trigger fire on purpose? yes, `test_rows_open_only_after_the_bar_holds` drives one rank to 12 GB against its peers' 5 GB and the rows open; the inverse cases stay shut, including `test_the_ramp_alone_never_opens_the_rows` (a rank that has not allocated yet) and the 12-second FSDP replay that reads `reserved imbalance 1%`.
SCENARIOS: A live recipe (run by hand) · A2 live recipe (run by hand) · B ok (B_xf_1gpu_long, B_cnn_1gpu) · C ok (C_M1_bert, C_sysB_1of4) · D ok (D_S2_ddp4, D_F1_fsdp4, D_sysA_ddp4_3k) · D' ok (Dd_S2_rank3_dead) · E ok (E_xf_2x1_long, E_cnn_2x1_fsdp) · E' ok (Ed_2x1_rank1_dead) · F no fixture · W ok (W_W1_busy, W_W2_idle) · Wc live recipe (run by hand)
PLATFORM: EVIDENCE: n/a (no `sys.platform` / `os.name` branch in this change).

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
@@ -84,11 +84,17 @@ def rows_hint(roll: Dict[str, Any], *, is_open: bool) -> str:
         return ""
     stale = int(roll.get("ranks_stale") or 0)
     parts = [f"{total} rank{'s' if total != 1 else ''}"]
-    if stale:
+    if stale and roll.get("excluding_stale"):
         parts.append(f"{stale} stale, excluded")
+    elif stale:
+        # Nothing is reporting, so nothing was excluded: the numbers above
+        # are the last ones these ranks sent.
+        parts.append("none reporting")
     imbalance = roll.get("reserved_imbalance_pct")
     if imbalance is not None:
-        parts.append(f"reserved imbalance {float(imbalance):.0f}%")
+        value = float(imbalance)
+        shown = "<1" if 0 < value < 1 else f"{value:.0f}"
+        parts.append(f"reserved imbalance {shown}%")
     parts.append("click to close" if is_open else "click to open")
     return " · ".join(parts)
 
@@ -100,7 +106,9 @@ def format_age(seconds: Any) -> str:
     value = float(seconds)
     if value < 90:
         return f"{value:.0f} s"
-    return f"{value / 60.0:.0f} min"
+    if value < 90 * 60:
+        return f"{value / 60.0:.0f} min"
+    return f"{value / 3600.0:.1f} h"
 
 
 def rows_html(
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
@@ -33,9 +33,9 @@ from .theme import (
     format_gb_pair,
     format_span,
     format_window,
-    num as _num,
-    sparkline_svg,
 )
+from .theme import num as _num
+from .theme import sparkline_svg
 
 # One colour per rank, shared by the charts and the rows' chips so a line
 # and a row are recognisably the same rank. Red is not among them: it reads
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
@@ -156,6 +156,26 @@ def rows_html(
     return f'<table class="tml-gpus">{head}{body}</table>'
 
 
+def drift_axis_bounds(values: List[Any]) -> Tuple[float, float, float]:
+    """A y range that fits the data, for a series whose signal is DRIFT.
+
+    RSS is a level a few GB high that moves by tens of MB across a run.
+    Zero-anchoring it (right for a share of capacity, like CPU) puts that
+    whole movement inside one pixel: measured on the 3-hour capture, the
+    ranks sit at 1.48-1.50 GB on an axis that would run to 5. The leak
+    this chart exists to show would be invisible.
+    """
+    numbers = [float(value) for value in values if value is not None]
+    if not numbers:
+        return (0.0, 1.0, 0.5)
+    low, high = min(numbers), max(numbers)
+    if high <= low:
+        high = low + max(abs(low) * 0.01, 0.01)
+    pad = (high - low) * 0.25
+    low, high = max(0.0, low - pad), high + pad
+    return (low, high, (high - low) / 2.0)
+
+
 def _series_axis(
     *groups: List[Dict[str, Any]],
 ) -> Optional[Tuple[float, float]]:
@@ -318,7 +338,14 @@ def _update_chart(
         for entry in entries
         for value in (entry.get("max") or entry.get(values_key) or [])
     ]
-    chart.options["yAxis"]["max"] = ymax_of(flat)
+    bounds = ymax_of(flat)
+    if isinstance(bounds, tuple):
+        low, high, tick = bounds
+        chart.options["yAxis"]["min"] = low
+        chart.options["yAxis"]["max"] = high
+        chart.options["yAxis"]["interval"] = tick
+    else:
+        chart.options["yAxis"]["max"] = bounds
     chart.update()
     window_s = float((run[0].get("window_s") if run else 0.0) or 0.0)
     words = format_window(window_s)
@@ -435,7 +462,7 @@ def update_process_section(
             window=series.get("rss") or [],
             aligned=aligned,
             scale=float(1024**3),
-            ymax_of=lambda values: theme.nice_ymax(values, floor=1.0),
+            ymax_of=drift_axis_bounds,
             tooltip=(
                 "Host memory held by each rank's trainer process. The "
                 "shape is the point: steady growth across the run is the "
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/process_section.py
@@ -1,176 +1,462 @@
-"""Process metrics — RAM% / GPU-mem% time-series + KPIs (PR2 revamp).
+# Copyright 2026 OptAI UG (haftungsbeschraenkt)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0
 
-Reuses the original client-side rollup math; renders with glass + ECharts.
-Renderer payload (history rows + series + gpu_used_imbalance) is unchanged.
+"""
+Process block: the trainer process on every rank, as four tiles, two charts
+and per-rank rows.
+
+Scope, stated on the block because it is the block's most useful fact: one
+process per rank, its own PID only. DataLoader workers are separate
+processes, so their CPU lands in the System block and never here. Reading
+the two blocks together is what the page is for: host CPU high while every
+rank's process CPU is low means the work is outside the trainer.
+
+Tiles are levels with their denominator; charts carry the two quantities
+whose information is their shape over time (CPU capacity, RSS); memory is a
+step function and stays a number. No verdict words: the diagnosis engine
+owns those.
 """
 
 from __future__ import annotations
 
-from datetime import datetime
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from nicegui import ui
 
 from . import theme
+from .theme import (
+    apply_span_axis,
+    cpu_axis_max,
+    format_gb_pair,
+    format_span,
+    format_window,
+    num as _num,
+    sparkline_svg,
+)
+
+# One colour per rank, shared by the charts and the rows' chips so a line
+# and a row are recognisably the same rank. Red is not among them: it reads
+# as a verdict.
+_RANK_COLORS = (
+    "#f97316",
+    "#3b82f6",
+    "#0d9488",
+    "#a855f7",
+    "#0ea5e9",
+    "#eab308",
+    "#ec4899",
+    "#10b981",
+)
+
+NA = "n/a"
+_MONO = "font-family:var(--mono);"
+
+# The block's scope, on screen rather than in a tooltip: without it the
+# CPU numbers read as the whole box's, and the pairing that makes them
+# diagnostic is invisible.
+SCOPE_NOTE = "one process per rank · DataLoader workers not included"
 
 
-def _to_ms(s: str) -> Optional[int]:
-    try:
-        return int(datetime.fromisoformat(s).timestamp() * 1000)
-    except Exception:
-        return None
+def rank_color(rank: int) -> str:
+    return _RANK_COLORS[int(rank) % len(_RANK_COLORS)]
 
 
-def _percentile(vals: List[float], p: float) -> float:
-    vals = sorted(v for v in vals if v is not None)
-    if not vals:
-        return 0.0
-    k = (len(vals) - 1) * p / 100.0
-    f = int(k)
-    c = min(int(k) + 1, len(vals) - 1)
-    return vals[f] if f == c else vals[f] * (c - k) + vals[c] * (k - f)
+def should_auto_open(*, prev_over: bool, over: bool) -> bool:
+    """Open on the rising edge only.
+
+    Twin of the System block's rule: a reader who closes the rows must not
+    be fought every tick while the condition persists.
+    """
+    return bool(over and not prev_over)
 
 
-def _compute_rollups(window: List[Dict[str, Any]]) -> Dict[str, Any]:
-    last = window[-1]
-    cpu_hist = [float(r.get("cpu_max", 0.0) or 0.0) for r in window]
-    ram_hist = [float(r.get("ram_used_max", 0.0) or 0.0) for r in window]
-    gpu_hist = [
-        float(r.get("gpu_used", 0.0) or 0.0)
-        for r in window
-        if r.get("gpu_used") is not None
-    ]
-    return {
-        "gpu_available": last.get("gpu_used") is not None,
-        "cpu": {
-            "now": cpu_hist[-1],
-            "p50": _percentile(cpu_hist, 50),
-            "p95": _percentile(cpu_hist, 95),
-        },
-        "ram": {
-            "now": ram_hist[-1],
-            "p95": _percentile(ram_hist, 95),
-            "total": float(last.get("ram_total", 0.0) or 0.0),
-        },
-        "gpu": {
-            "now": gpu_hist[-1] if gpu_hist else 0.0,
-            "p95": _percentile(gpu_hist, 95) if gpu_hist else 0.0,
-        },
+def rows_hint(roll: Dict[str, Any], *, is_open: bool) -> str:
+    """Header of the per-rank rows: coverage, imbalance, and nothing else.
+
+    It states what was observed. Whether that is bad is the engine's call,
+    so no word here classifies it.
+    """
+    total = int(roll.get("ranks_total") or 0)
+    if not total:
+        return ""
+    stale = int(roll.get("ranks_stale") or 0)
+    parts = [f"{total} rank{'s' if total != 1 else ''}"]
+    if stale:
+        parts.append(f"{stale} stale, excluded")
+    imbalance = roll.get("reserved_imbalance_pct")
+    if imbalance is not None:
+        parts.append(f"reserved imbalance {float(imbalance):.0f}%")
+    parts.append("click to close" if is_open else "click to open")
+    return " · ".join(parts)
+
+
+def format_age(seconds: Any) -> str:
+    """A per-rank age in the same words the strip uses."""
+    if seconds is None:
+        return NA
+    value = float(seconds)
+    if value < 90:
+        return f"{value:.0f} s"
+    return f"{value / 60.0:.0f} min"
+
+
+def rows_html(
+    ranks: List[Dict[str, Any]], series: List[Dict[str, Any]]
+) -> str:
+    """Per-rank table: identity, capacity, memory, and how fresh it is.
+
+    A stale rank is dimmed and kept. Dropping it would hide the one fact
+    worth having when a job stalls, which is WHICH rank stopped.
+    """
+    trend = {
+        int(entry.get("global_rank", -1)): (
+            entry.get("avg") or entry.get("v") or []
+        )
+        for entry in series or []
     }
+    head = (
+        "<tr><th>rank</th><th>gpu</th><th>node</th><th>cpu cap</th>"
+        "<th>rss</th><th>cpu trend</th><th>cuda allocated</th>"
+        "<th>cuda reserved</th><th>age</th></tr>"
+    )
+    body = ""
+    for rank in ranks:
+        idx = int(rank.get("global_rank", 0))
+        colour = rank_color(idx)
+        used, rest = format_gb_pair(
+            rank.get("ram_used"), rank.get("ram_total")
+        )
+        reserved, reserved_rest = format_gb_pair(
+            rank.get("gpu_reserved"), rank.get("gpu_total")
+        )
+        alloc, alloc_rest = format_gb_pair(rank.get("gpu_alloc"), None)
+        cap = rank.get("cpu_capacity")
+        stale = bool(rank.get("stale"))
+        row_open = '<tr class="tml-stale">' if stale else "<tr>"
+        gpu_index = rank.get("gpu_index")
+        node_rank = rank.get("node_rank")
+        gpu_cell = f"G{int(gpu_index)}" if gpu_index is not None else NA
+        node_cell = f"N{int(node_rank)}" if node_rank is not None else NA
+        cap_cell = f"{_num(cap, '{:.1f}')} %" if cap is not None else NA
+        body += (
+            row_open + f'<td><span style="color:{colour}">■</span> R{idx}</td>'
+            f"<td>{gpu_cell}</td>"
+            f"<td>{node_cell}</td>"
+            f"<td>{cap_cell}</td>"
+            f"<td>{used} {rest}</td>"
+            f"<td>{sparkline_svg(trend.get(idx, []), colour)}</td>"
+            f"<td>{alloc} {alloc_rest}</td>"
+            f"<td>{reserved} {reserved_rest}</td>"
+            f"<td>{format_age(rank.get('age_s'))}</td>"
+            "</tr>"
+        )
+    return f'<table class="tml-gpus">{head}{body}</table>'
+
+
+def _series_axis(
+    *groups: List[Dict[str, Any]],
+) -> Optional[Tuple[float, float]]:
+    """One anchor and span covering every series on the block.
+
+    Both charts are pinned to it so a vertical read across the pair, and
+    across to the System block's charts above, lands on the same moment.
+    """
+    starts, ends = [], []
+    for group in groups:
+        for entry in group or []:
+            stamps = entry.get("t") or []
+            if stamps:
+                starts.append(stamps[0])
+                ends.append(stamps[-1])
+    if not ends:
+        return None
+    newest = max(ends)
+    return (newest, max(newest - min(starts), 1.0))
 
 
 def build_process_section() -> Dict[str, Any]:
-    kpis: Dict[str, Any] = {}
+    """Build the block once; ``update_process_section`` fills it per tick."""
+    panel: Dict[str, Any] = {"tiles": {}, "subs": {}}
     card = ui.element("div").classes("glass reveal")
     card.style(
-        "padding:18px 20px; width:100%; height:100%; "
-        "display:flex; flex-direction:column; overflow:hidden;"
+        "padding:18px 20px; width:100%; display:flex; "
+        "flex-direction:column; overflow:hidden;"
     )
     with card:
         with (
             ui.row()
             .classes("w-full items-center")
-            .style("margin-bottom:8px; gap:12px;")
+            .style("margin-bottom:10px; gap:12px;")
         ):
             ui.label("Process").classes("ctitle")
-            for nm, col in [("RAM", theme.C_CPU), ("GPU mem", theme.C_GPU)]:
-                with ui.element("div").classes("legchip"):
-                    ui.element("div").classes("legdot").style(
-                        f"background:{col};"
-                    )
-                    ui.label(nm)
             ui.element("div").style("flex:1;")
-            win = ui.label("waiting for data").classes("cmeta")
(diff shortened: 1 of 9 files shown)


abhinavsriva (changes requested):
Plz change

abhinavsriva:
I agree with the intended functionality: per-rank visibility, dead-rank handling, normalized CPU, recent and retained history, and whole-run aggregation can all be useful.

My request for changes is about architecture, scope, and reviewability. Since we are doing a deeper changes, it might be better also to do proper refactoring so the code is modular and have clear boundaries and not build upon the previous tech debt. 

The current PR changes Process data access, computation, freshness, caching, payload structure, presentation, shared chart helpers, and parts of the System presentation together. Even when each individual decision is reasonable, combining all of them in one PR makes the data flow and behavior difficult to review and maintain.

## Requested approach

Please close or supersede the current PR and restart from `version_0.3.7`.

Add the functionality progressively through multiple small PRs. Please do not put the complete data-layer redesign, whole-run aggregation, freshness behavior, caching, Process UI, System changes, and theme refactoring into one PR.

Each PR should introduce one clear responsibility, preserve a readable data flow, and have focused tests for that responsibility.

## Target architecture

The Process data flow should be:

```text
SQLite
  -> ProcessRepository
  -> ProcessDashboardComputer
  -> ProcessDashboardPayload
  -> process_section.py
```

Runtime and shared policy should enter the compute layer explicitly:

```text
Runtime settings -> FreshnessPolicy -------+
                                             |
RunSeriesPolicy -> RunSeriesPlanner --------+-> ProcessDashboardComputer
```

### Suggested files and responsibilities

```text
src/traceml_ai/renderers/shared/
    run_series.py
    freshness.py

src/traceml_ai/renderers/process/
    repository.py
    dashboard_models.py
    dashboard_compute.py
    computer.py

src/traceml_ai/renderers/system/
    repository.py
    dashboard_models.py
    dashboard_compute.py
    computer.py

src/traceml_ai/aggregator/display_drivers/nicegui_sections/
    charting.py
    formatting.py
    process_section.py
    system_section.py
    theme.py
```

#### `renderers/shared/run_series.py`

This is shared compute infrastructure, not view code. It should own only reusable run-series mechanics:

- Recent-versus-retained series selection.
- Rolling-window selection.
- Cadence and stride planning.
- Point-budget enforcement.
- Optional refresh policy if caching is demonstrated to be necessary.

It should not contain SQLite table names, Process/System metric definitions, NiceGUI code, or arbitrary SQL expressions.

The shared module should produce a plan such as `RunSeriesPlan`. The Process and System repositories should execute their own table-specific SQL using that plan.

#### `renderers/shared/freshness.py`

This should define one explicit, sampler-aware telemetry freshness policy.

- Receive the configured sampler interval or a clearly defined observed cadence.
- Calculate when telemetry becomes stale.
- Be reusable by Process and System where entity-level freshness is needed.
- Do not hardcode a separate two-second sampling interval in multiple modules.

Telemetry freshness must remain separate from cached-payload fallback:

- **Telemetry freshness** determines whether a rank or node is still reporting.
- **Last-good TTL** determines how long cached output may be returned after a database or renderer failure.

These are different behaviors and should not share an ambiguous `stale` policy.

#### `process/repository.py`

This layer should own `process_samples` database access:

- Open SQLite connections.
- Fetch recent samples per rank.
- Fetch the latest sample per rank.
- Fetch run statistics.
- Execute run-history SQL using a supplied `RunSeriesPlan`.

It should not choose dashboard windows, freshness thresholds, cache durations, diagnostic thresholds, tile estimators, or disclosure behavior.

Please prefer explicit methods such as `fetch_cpu_capacity_run(plan)` and `fetch_rss_run(plan)` over a generic method that accepts an arbitrary `value_sql` string.

#### `process/dashboard_compute.py`

This layer should own Process metric meaning:

- Group samples by rank.
- Normalize CPU against logical core count.
- Apply the shared freshness policy.
- Decide which ranks contribute to aggregates.
- Calculate per-rank medians.
- Select worst CPU and RSS ranks.
- Select the least-headroom CUDA rank.
- Calculate reserved-memory imbalance.
- Request recent or retained series through the shared run-series interface.
- Build one typed `ProcessDashboardPayload`.

If automatic expansion is intended to follow diagnostics, use a public shared diagnostic signal or classifier. Importing one diagnostic threshold while independently reimplementing the metric can still drift from the actual diagnostic rule.

#### `process/dashboard_models.py`

Define explicit payload contracts such as:

- `RankSnapshot`
- `MetricRollup`
- `ChartTrace`
- `ChartSeries`
- `ProcessDashboardPayload`

The chart payload should explicitly state whether it contains `recent` or `retained` history. The view should not infer the mode from whichever dictionary key happens to be non-empty.

Avoid deeply nested `Dict[str, Any]` contracts where practical. They make it too easy for computation and fallback behavior to leak into the view.

#### `process_section.py`

This should be presentation only:

- Build NiceGUI components.
- Format values and units.
- Apply colors and chart options.
- Render the payload.
- Maintain UI-only state such as the rising edge of an expansion panel.

It should not execute SQL, aggregate metrics, calculate freshness, own diagnostic thresholds, or decide which ranks are excluded.

#### `charting.py`, `formatting.py`, and `theme.py`

- `charting.py`: reusable axis, time-series, line, and sparkline helpers.
- `formatting.py`: shared duration, number, memory, and missing-value formatting.
- `theme.py`: colors, fonts, CSS, spacing, and generic visual tokens.

Please do not turn `theme.py` into a catch-all for data transformation, time parsing, formatting, series alignment, SVG generation, and chart mutation.

## Required PR sequence

Please build this progressively from `version_0.3.7`:

### PR 1: Process architecture only

- Introduce the Process repository, typed models, compute boundary, and view boundary.
- Preserve existing `version_0.3.7` behavior.
- Do not add whole-run behavior or redesign the UI yet.
- Do not change System or `theme.py`.

### PR 2: Shared run-series and freshness infrastructure

- Add the small shared `run_series.py` and `freshness.py` modules.
- Add focused unit tests for window planning, point budgets, cadence, and freshness.
- Keep table-specific SQL in the Process repository.
- Wire the real sampler interval or an explicit observed-cadence input.

### PR 3: Process compute and payload features

- Add per-rank independent history.
- Add rank freshness and coverage.
- Add normalized CPU and Process-specific rollups.
- Add retained/whole-run CPU and RSS series.
- Add caching only if it has a demonstrated need and an explicit policy.
- Test the payload without involving NiceGUI.

### PR 4: Process dashboard presentation

- Add the four tiles, charts, and per-rank rows.
- Consume the completed typed payload without changing computation.
- Keep this PR focused on layout, formatting, accessibility, and presentation tests.

### PR 5: System migration and synchronization

- In a separate follow-up PR, migrate System to the same shared run-series and freshness infrastructure.
- Remove duplicated rolling-window and point-budget policies.
- Move System metric interpretation out of the view.
- Preserve System behavior during the migration.

### PR 6: Shared presentation cleanup, if still required

- Move genuinely shared chart and formatting helpers into clearly named modules.
- Keep `theme.py` limited to styling.
- Make this a behavior-preserving presentation refactor.

The PRs may be stacked if necessary, but they must remain independently understandable and reviewable. Each PR should make it clear what data enters, what computation is performed, what payload is produced, and what the view renders.

## Merge position

I do not recommend merging the current branch incrementally in its present form. It combines too many responsibilities and makes Process depend on architectural choices copied from System before those choices have been made reusable and consistent.

Please use `version_0.3.7` as the clean base, introduce the Process work in the small PRs above, and then bring System onto the same architecture in a separate synchronization PR.

The requested change is not to remove the intended features. It is to implement them with clear ownership, reusable policy, typed boundaries, and a reviewable progression instead of one large mixed-scope PR.

Pendu:
Superseded by #403, which lands this work from `version_0.3.7` as the sequence agreed in the review.

The branch stays in place as the reference for parts 3 and 4.

Issue #284, opened 2026-07-29
System telemetry: preserve measurement quality from collection to presentation


abhinavsriva:
## Summary

TraceML currently cannot distinguish a genuine zero measurement from a failed system-metric read. `SystemSampler` substitutes `0.0` when CPU or RAM collection fails and substitutes a fully zeroed GPU record when any NVML field read fails.

This issue introduces an end-to-end measurement-quality contract for System telemetry. Successful measurements remain numeric, including genuine `0.0` values. Failed, unavailable, or missing measurements use `null`, while a compact two-bit quality mask records why each value is absent.

This parent issue covers the complete System telemetry path. Detailed child tickets will be created separately.

## Why this is needed now

Fallback zeros look like successful measurements and can affect:

- Time-series charts and rollups.
- CPU, RAM, GPU-utilization, temperature, and power summaries.
- Bottleneck and resource-pressure diagnostics.
- Comparisons between nodes and GPUs.

Once a fallback zero has been persisted, it cannot be distinguished from a genuine historical zero. Correcting the contract before more consumers depend on the current behavior avoids accumulating ambiguous telemetry and another data migration later.

This work is about measurement correctness.

### CPU and RAM

If a psutil read fails, the sampler logs the exception and returns `0.0`. Downstream consumers cannot tell whether the machine was genuinely idle or the measurement failed.

### GPU

The sampler reads utilization, memory, temperature, power usage, and power limit inside one `try` block. If any read fails, all six fields for that GPU are replaced with `0.0`, including fields that were collected successfully.

### Downstream

The SQLite projection already permits nullable metric columns, but several consumers convert `null` back to zero. The new contract must therefore be implemented through sampling, transport, storage, computation, diagnostics, and presentation as one vertical slice.

## Scope

This issue applies only to:

- `SystemSampler`
- `SystemSample` and `GPUMetrics`
- System telemetry wire encoding
- `system_samples` and `system_gpu_samples`
- System aggregation and diagnostics
- System CLI, dashboard, and report presentation
- Backward compatibility and end-to-end tests

A shared quality enum and packing helpers may be reused by future samplers, but this issue does not migrate those samplers.

## Measurement-quality contract

Every fixed System metric has one of four states:

| Bits | Quality | Meaning |
|---|---|---|
| `00` | `OK` | Collection succeeded. The value is numeric and may be `0.0`. |
| `01` | `UNAVAILABLE` | The metric is unsupported or not applicable. |
[...]

PR #402, opened 2026-08-26
system samplers none removed

abhinavsriva:
Preserve unavailable System sampler measurements as `None` instead of reporting fabricated zero values.

Part of #394.

Partial implementation of #284.

## Changes

- Return `None` when system CPU or RAM collection fails.
- Keep total RAM unknown when initialization fails.
- Preserve GPU device-index alignment when a GPU read fails, while reporting its unavailable measurements as `None`.
- Update the System sample wire schema to support nullable CPU, RAM, and per-GPU measurements.
- Add focused collection-failure and wire-roundtrip tests.

## Scope

This PR changes upstream periodic System sampler and wire semantics only. SQLite already stores these values as `NULL`.

GPU discovery semantics (`gpu_available` and `gpu_count`), the system manifest, and terminal/dashboard/diagnosis presentation are intentionally unchanged. Those concerns can be handled separately without expanding this PR.

### DIFF
--- a/src/traceml_ai/samplers/schema/system.py
+++ b/src/traceml_ai/samplers/schema/system.py
@@ -15,7 +15,7 @@ Design principles
 """
 
 from dataclasses import dataclass
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 # ---------------------------------------------------------------------------
 # GPU metrics
@@ -39,14 +39,14 @@ class GPUMetrics:
     power_limit : watts
     """
 
-    util: float
-    mem_used: float
-    mem_total: float
-    temperature: float
-    power_usage: float
-    power_limit: float
+    util: Optional[float]
+    mem_used: Optional[float]
+    mem_total: Optional[float]
+    temperature: Optional[float]
+    power_usage: Optional[float]
+    power_limit: Optional[float]
 
-    def to_wire(self) -> List[float]:
+    def to_wire(self) -> List[Optional[float]]:
         """
         Convert this GPU metric to a compact wire representation.
 
@@ -62,7 +62,7 @@ class GPUMetrics:
 
         Returns
         -------
-        List[float]
+        List[Optional[float]]
             Compact, allocation-light representation suitable for
             JSON serialization or binary transport.
         """
@@ -76,13 +76,13 @@ class GPUMetrics:
         ]
 
     @staticmethod
-    def from_wire(data: List[float]) -> "GPUMetrics":
+    def from_wire(data: List[Optional[float]]) -> "GPUMetrics":
         """
         Reconstruct GPUMetrics from its wire representation.
 
         Parameters
         ----------
-        data : List[float]
+        data : List[Optional[float]]
             Wire-format list produced by `to_wire()`.
 
         Returns
@@ -123,9 +123,9 @@ class SystemSample:
 
     sample_idx: int
     timestamp: float
-    cpu_percent: float
-    ram_used: float
-    ram_total: float
+    cpu_percent: Optional[float]
+    ram_used: Optional[float]
+    ram_total: Optional[float]
     gpu_available: bool
     gpu_count: int
     gpus: List[GPUMetrics]
--- a/src/traceml_ai/samplers/system_sampler.py
+++ b/src/traceml_ai/samplers/system_sampler.py
@@ -18,7 +18,7 @@ The sampler is designed to be:
 from __future__ import annotations
 
 import time
-from typing import List
+from typing import List, Optional
 
 import psutil
 from pynvml import (
@@ -97,7 +97,7 @@ class SystemSampler(BaseSampler):
 
         Stores total system memory for reuse across samples.
         """
-        self.ram_total_memory = 0.0
+        self.ram_total_memory: Optional[float] = None
         try:
             self.ram_total_memory = float(psutil.virtual_memory().total)
         except Exception as e:
@@ -124,7 +124,7 @@ class SystemSampler(BaseSampler):
                 f"[TraceML] NVML initialization failed (GPU unavailable): {e}"
             )
 
-    def _sample_cpu(self) -> float:
+    def _sample_cpu(self) -> Optional[float]:
         """
         Sample current CPU utilization.
         """
@@ -132,9 +132,9 @@ class SystemSampler(BaseSampler):
             return float(psutil.cpu_percent(interval=None))
         except Exception as e:
             self.logger.error(f"[TraceML] CPU sampling failed: {e}")
-            return 0.0
+            return None
 
-    def _sample_ram(self) -> float:
+    def _sample_ram(self) -> Optional[float]:
         """
         Sample current RAM usage.
         """
@@ -142,14 +142,14 @@ class SystemSampler(BaseSampler):
             return float(psutil.virtual_memory().used)
         except Exception as e:
             self.logger.error(f"[TraceML] RAM sampling failed: {e}")
-            return 0.0
+            return None
 
     def _sample_gpus(self) -> List[GPUMetrics]:
         """
         Sample all available GPUs.
 
-        Failure to sample one GPU does not affect others, and a zeroed
-        placeholder is inserted to preserve index alignment.
+        Failure to sample one GPU does not affect others. An unavailable
+        placeholder preserves index alignment without fabricating zeros.
         """
         if not self.gpu_available:
             return []
@@ -183,14 +183,17 @@ class SystemSampler(BaseSampler):
                 self.logger.error(
                     f"[TraceML] GPU {gpu_id} sampling failed: {e}"
                 )
+                # Preserve GPU index alignment while marking every failed
+                # measurement unavailable. Zeros would look like valid idle
+                # hardware and are therefore incorrect telemetry.
                 gpus.append(
                     GPUMetrics(
-                        util=0.0,
-                        mem_used=0.0,
-                        mem_total=0.0,
-                        temperature=0.0,
-                        power_usage=0.0,
-                        power_limit=0.0,
+                        util=None,
+                        mem_used=None,
+                        mem_total=None,
+                        temperature=None,
+                        power_usage=None,
+                        power_limit=None,
                     )
                 )
 
--- a/tests/samplers/test_system_sampler.py
+++ b/tests/samplers/test_system_sampler.py
@@ -1,0 +1,62 @@
+# Copyright 2026 OptAI UG (haftungsbeschraenkt)
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import traceml_ai.samplers.system_sampler as system_module
+from traceml_ai.samplers.schema.system import GPUMetrics, SystemSample
+from traceml_ai.samplers.system_sampler import SystemSampler
+
+
+class _Logger:
+    def error(self, *args, **kwargs) -> None:
+        return None
+
+
+def _raise_unavailable(*args, **kwargs):
+    raise RuntimeError("unavailable")
+
+
+def test_collection_failures_are_unavailable_not_zero(monkeypatch) -> None:
+    sampler = SystemSampler.__new__(SystemSampler)
+    sampler.logger = _Logger()
+    sampler.gpu_available = True
+    sampler.gpu_count = 1
+
+    monkeypatch.setattr(
+        system_module.psutil,
+        "cpu_percent",
+        _raise_unavailable,
+    )
+    monkeypatch.setattr(
+        system_module.psutil,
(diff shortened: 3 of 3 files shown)


Pendu:
This is the System half of what #400 did for Process, and it lands cleanly. I went at the GPU placeholder path hardest, since that is the part with real structure to get wrong, and it holds up. Nothing blocking. The cluster view is the part to read: it is the only consumer that actually distinguishes the new nulls, which fixes a fabricated straggler and introduces one false statement. Then a coverage gap on the headline claim, two questions, and a short list of places the null is recorded but not yet consumed.

## Verified

- Full suite `1333 passed, 2 skipped` at head against `1331 passed, 2 skipped` at the merge base, so exactly the two added tests and no regressions.
- Index alignment actually works, which is the claim I most wanted to see proven. I built a two-GPU sample where device 0 fails and device 1 succeeds, then pushed it through `to_wire()`, the real msgspec codec, `sqlite_writers/system.build_rows` and into SQLite. Result: `gpu_idx 0` lands `util=None mem_used_bytes=None temperature_c=None` and `gpu_idx 1` keeps `12.0 / 5.0 / 40.0`. The failed device does not shift or swallow the healthy one.
- Nulls survive the whole path. `SystemSample.from_wire(sample.to_wire()) == sample` holds for an all-null sample including null GPU rows, and the row reaches SQLite as real NULL rather than being coerced back to zero.
- The writer was already ready for this. `sqlite_writers/system.py:349` guards `isinstance(g, list) and len(g) >= 6` before positional unpacking and then isinstance-guards each of the six elements, and the `system_gpu_samples` columns at `:166-171` are plain `REAL` with no `NOT NULL`. The arity of the wire list is unchanged at 6, so only the element type widened.
- Backward compatibility holds in the direction that matters. The released v0.3.6 system writer already has 14 `else None` guards, the same `len(g) >= 6` check, and no `NOT NULL` on the GPU metric columns, so an upgraded rank sending nulls to an older aggregator stores NULL instead of failing.
- The single-node terminal and dashboard paths render identically to base. `cli_compute.py:148-150` and `dashboard_compute.py` use the `or 0.0` idiom, which catches None and 0.0 alike. That also closes the one asymmetry worth checking in `renderers/system/renderer.py:70-71`, where the guard tests `ram_total` but the arithmetic uses `ram_used`: unreachable, because both are coerced before the renderer sees them. The cluster path is a different story and is the one section below that I would not merge without reading.
- Keeping `gpu_available` and `gpu_count` non-nullable is the right call and I checked it is load-bearing. Since #400, `process_samples.gpu_available` can be all-NULL, so `system_samples` is now the only reliable answer to "does this run have a GPU". With process NULL and system carrying a real bool, `detect_gpu_available` still returns False and True correctly in both directions, so the "No GPU detected" path is intact.
- CI covers the new file. `tests/samplers` is at `ci.yml:86` from #400, and that job is torch-free but `nvidia-ml-py` is a core runtime dependency at `pyproject.toml:46`, not an extra, so `pynvml` is present and all 4 tests in `tests/samplers/` pass there.
- `pre-commit run` at the pinned revs passes on all three files, `tests/arch` passes, and a real `--mode summary` run at head still prints `CPU only (no GPU detected)` with no exception.
- The branch is two commits behind `version_0.3.7`, so I checked the conclusions against the merged state rather than head alone. Those two commits (#410 process dashboard, #399 CUDA-allocated rename) touch `renderers/process/`, `reporting/terminal_card/`, docs and process tests only, and nothing under `renderers/system/`, `renderers/context/`, `sqlite_writers/system.py` or the system schema, so nothing here changes after a rebase.

## The cluster view changes numbers, mostly for the better and once for the worse

`renderers/system/cli_cluster.py` does not use the `or 0.0` idiom. It uses `_optional_float` and then filters `None` out, so unlike every other view it distinguishes the new nulls. I drove the real `_node_sample_from_row` and `_rollup` with one failed node and one healthy node at 80%, changing only whether a failed read is stored as `0.0` or `NULL`:

```
BASE (failed reads = 0.0)     gpu_util  median 40.0  worst 0.0  @ nodeA-failed
                              ram       median 40.0  worst 80.0 @ nodeB-healthy
                              cpu       median 40.0  worst 80.0 @ nodeB-healthy
HEAD (failed reads = NULL)    gpu_util  median 80.0  worst 80.0 @ nodeB-healthy
                              ram       median 80.0  worst 80.0 @ nodeB-healthy
                              cpu       median 40.0  worst 80.0 @ nodeB-healthy
```

Two of those are a straight improvement and worth calling out as a benefit of this PR rather than a risk. A failed NVML read used to enter the rollup as a real 0%, which both dragged the median down and made the failed node win `worst_node` for GPU utilisation. Base was inventing an idle straggler out of a read failure and then pointing at it. Head excludes it and correctly blames the healthy node.

The exception is CPU, and it is the same line in the same function: `cli_cluster.py:248` still does `float(row["cpu_percent"] or 0.0)` while `ram_percent` right below it uses `_optional_float` and yields `None`. So within one constructor, a node whose CPU sampling failed still contributes a fabricated 0% to the CPU median and can still be falsely named the worst node, while the same node is correctly excluded from RAM and GPU. That inconsistency was invisible on base because everything was zero-filled. It is visible now, and it looks unintended rather than deliberate.

The one I would actually fix: with GPUs present but every NVML read failing, the cluster panel now prints "GPU: Not available":

```
Machine HAS 2 GPUs (gpu_available=True in the DB); every NVML read fails
  BASE (0.0)   -> panel gpu_available = True   => GPU rows at 0%
  HEAD (NULL)  -> panel gpu_available = False  => "GPU: Not available"
```

`cli_cluster.py:176-178` derives the panel's own `gpu_available` from `any(sample.gpu_util_percent is not None ...)`, that is from whether metrics arrived, rather than from the authoritative `gpu_available` column. It is the only site that does this: `cli_compute.py:151`, `dashboard_compute.py:139` and `reporting/sections/system/loader.py:157` all read the column with the `or False` idiom, which is why the cluster panel is the only view that diverges. This PR deliberately keeps that column truthful, so the panel now contradicts the very field the PR went out of its way not to make nullable. Base was merely misleading, reporting real hardware at 0%; head makes a positively false statement about the hardware. It is presentation, which is your stated deferral, so I am not calling it blocking, but it is the one place where deferring costs a correct answer rather than a pretty one, and reading the real column instead is a small change.

## The alignment test does not test alignment

`test_collection_failures_are_unavailable_not_zero` sets `gpu_count = 1` and then asserts `_sample_gpus()[0].to_wire() == [None] * 6`. With a single device there is no alignment to preserve, so the assertion passes for any implementation that returns one placeholder, including one that returned the wrong index or dropped the entry entirely.

Index alignment only has meaning at two or more devices with a mix of failing and succeeding reads. That is the case I had to construct by hand to convince myself the change was correct, and it is the one a future refactor would break silently. Making the existing test use `gpu_count = 2` with only one device raising would pin the real property.

## Two questions

**One unsupported NVML query discards five good measurements.** The per-GPU try block at `system_sampler.py:161-169` reads util, memory, temperature, power, then power limit in sequence, and a single failure anywhere sends all six fields to None. `nvmlDeviceGetPowerManagementLimit` is the last call and is commonly unsupported on consumer and virtualised GPUs, so the likely real-world case is util, memory and temperature succeeding and being thrown away because the power limit is not queryable. Base had the same discard, so this is not a regression, but the reasoning in the new comment ("zeros would look like valid idle hardware and are therefore incorrect telemetry") applies just as well to nulling five fields that were measured. Do you want per-field granularity here, or is whole-device unavailability the semantics you intend? The bare `except Exception` also means `NVMLError_NotSupported` is not distinguished from a real failure, though `NVMLError` is already imported and used at `:122`.

**`cpu_logical_core_count` is now the last one left.** After #400 and this PR, the only measurement field still initialised to a fabricated zero in both samplers is the logical core count: `system_sampler.py:85` and `process_sampler.py:106,111`, still typed `int` in both schemas at `schema/system.py` and `schema/process.py:116`. Is that a third PR, or is core count considered discovery rather than measurement and therefore deliberately grouped with `gpu_available`/`gpu_count`? Worth noting that the same type-checker run already reports `int | None is not assignable to attribute cpu_logical_core_count with type int` at `system_sampler.py:88` on both branches, so there is a latent mismatch on that field independent of this PR: `psutil.cpu_count()` can return `None`.

## The null exists now, but nothing reads it yet

This is the part I would think about before the next PR rather than this one. Everything below is identical on base and head, so none of it is a regression, but this PR is what creates the distinction that would make each of them fixable.

- **Diagnosis cannot say "measured nothing".** `_default_primary` in `diagnostics/system/api.py:70` gates `NO_DATA` on `signals.samples <= 0`, and `samples` counts rows, not measured values. A run whose System sampling failed completely still produces rows, so it falls through to `NORMAL` with "CPU and RAM showed no system pressure." I ran all three cases against `_default_primary`: rows with all-NULL measurements gives `NORMAL`, rows with fabricated `0.0` gives `NORMAL`, and only zero rows gives `NO DATA`. So base and head agree exactly. The difference is that on base the claim was at least backed by literal zeros, whereas now the section asserts an absence of pressure from measurements it explicitly recorded as unknown.
- **Presentation converts the nulls straight back.** `renderers/system/cli_compute.py:148-150` and its dashboard equivalents apply `or 0.0`, so the storage layer now knows the difference between measured-zero and unmeasured and the view layer discards it. That is your stated display deferral and it is why nothing crashes, so I am only noting where the information stops.
- **The manifest still writes a fabricated zero for the same value.** `system_manifest.py:46` and `:182` annotate `ram_total_memory: float`, but `system_sampler.py:72` now passes a value that can be `None`, and `:111` coerces it with `float(ram_total_memory or 0.0)`. Nothing raises, and there is no type checking in CI, so this is not a failure, and the emitted manifest is byte-identical to base (`ram_total_bytes = 0.0` on both) since `:111` already had that coercion. Only the in-memory attribute type changed. Running a type checker over base and head does show the delta cleanly: base reports one error in these two files, head reports two, and the new one is `Argument 'float | None' is not assignable to parameter 'ram_total_memory' with type 'float'` at `system_sampler.py:72`. Still, the manifest records a fabricated `0.0` for the exact init failure this PR just taught the sample stream to report honestly. The manifest is in your stated out-of-scope list, so this is a flag rather than a request, though the annotation is now inaccurate either way.

## Coverage note

`schema/system.py` reaches 100% from the round-trip test. `system_sampler.py` sits at 55%, and the uncovered block includes `sample()` at `:211-224`, which is the method that actually assembles the nullable `SystemSample` and calls `to_wire()`. So the field wiring the PR changes is exercised only through the helpers, never through the method that emits a sample, the same gap #400 had. The GPU success path at `:163-171` is fairly uncoverable without hardware, so that part is expected.

For contrast on the asymmetry with #400: leaving `ProcessGPUMetrics` non-nullable there while making `GPUMetrics` nullable here reads inconsistent at first glance, but it is right. Process has one device and `_sample_gpu` returns None for the whole object, while System has N devices and needs an index-aligned placeholder. Different shapes, same principle.

Pendu:
Correction to my own framing above, in your favour on one point and against it on the other.

I described the cluster "GPU: Not available" case as needing every NVML read to fail, which reads like a driver disaster you would never see. The real trigger is much smaller, and it makes the two items I raised separately into one.

`_sample_gpus` wraps all six NVML calls in a single try block, so any one of them failing nulls the whole `GPUMetrics`. `nvmlDeviceGetPowerManagementLimit` is the last call and is unsupported on plenty of otherwise healthy cards. I stubbed a GPU where util, memory, temperature and power all succeed and only the power limit raises:

```
util/mem/temp/power ALL succeeded; only power_limit raised.
  GPU0 emitted -> [None, None, None, None, None, None]
  GPU1 emitted -> [None, None, None, None, None, None]
  cluster panel gpu_available = False
```

Because an unsupported query is a property of the hardware rather than of the moment, it fires identically on every node of a homogeneous cluster, which is exactly the condition `any(sample.gpu_util_percent is not None ...)` at `cli_cluster.py:176` needs to go false. So the path to a cluster panel saying "GPU: Not available" on a working GPU fleet is one unsupported power-limit query, not a fleet-wide failure.

That also means the granularity question and the panel finding are the same defect seen from two ends. Reading the real `gpu_available` column at `cli_cluster.py:176` fixes the symptom; catching per call, or at least splitting the power reads out of the block that carries util and memory, removes the cause and stops discarding four good measurements. Either is small, and doing the second makes the first much less likely to matter.

Nothing else in the review changes.

abhinavsriva:
Thanks. The per-field NVML handling, authoritative cluster gpu_available, null-aware presentation/diagnostics, and stronger mixed-GPU coverage are valid follow-ups, but they are outside this PR’s narrow sampler/wire nullability change and are already tracked by #284.  I AM merging it as-is.

Issue #406, opened 2026-08-28
Process compute and payload features


Pendu:
Part 3 of #403. The functional content of #401's data layer, rebuilt on the #404 boundaries and the
#405 infrastructure.

**Scope**
- [ ] Per-rank independent history (`RankSnapshot` in the typed models)
- [ ] Rank freshness and coverage via the shared freshness policy
- [ ] CPU normalized against the stored logical core count
- [ ] Process rollups: worst-rank CPU, worst-rank RSS chosen on window median, least-headroom
      CUDA rank, reserved-memory imbalance
- [ ] Retained/whole-run CPU and RSS series through the shared run-series interface; the payload
      states `recent` or `retained` explicitly, the view never infers it
- [ ] Explicit repository methods (for example `fetch_cpu_capacity_run(plan)`), no `value_sql`
      string parameters
- [ ] Caching only if a measured need exists, with an explicit policy
- [ ] Payload tested without NiceGUI

**Non-goals.** No UI changes; the existing card keeps rendering from the extended payload.

Issue #403, opened 2026-08-28
RFC: Process dashboard architecture, delivered as six PRs


Pendu:
The dashboard rebuild's Process work lands as a sequence of small PRs rather than one, following the review on #401. This issue is the agreed plan; each part below is its own issue and its own PR.

## Why

#401 carried the Process block's data access, computation, freshness, caching, payload shape and presentation in one diff, plus changes to the already-merged System block. Each decision in it was defensible, but together they are not reviewable as one unit, and Process ended up copying architectural choices out of System before those choices had been made reusable. The review asked for the same functionality, introduced in order, on a clean base.

## Approach

Build from `version_0.3.7`. Process becomes the reference implementation of the layering, then System is migrated onto it.

```
SQLite -> ProcessRepository -> ProcessDashboardComputer -> ProcessDashboardPayload -> process_section.py

Runtime settings -> FreshnessPolicy ----+
                                        +-> ProcessDashboardComputer
RunSeriesPolicy -> RunSeriesPlanner ----+
```

## The sequence

- [ ] #404 Part 1. Process repository, typed models, compute and view boundaries (behaviour-preserving)
- [ ] #405 Part 2. Shared run-series and freshness infrastructure under `renderers/shared/`
- [ ] #406 Part 3. Process compute and payload features
- [ ] #407 Part 4. Process dashboard presentation
- [ ] #408 Part 5. Migrate System onto the shared infrastructure
- [ ] #409 Part 6. Presentation-layer cleanup: `charting.py`, `formatting.py`, `theme.py`

Parts 1 to 4 deliver the Process block. Part 5 brings System onto the same architecture. Part 6 is a behaviour-preserving move of shared presentation helpers.

## Ground rules for the series

Each PR introduces one responsibility, states its blast radius in its body, and carries tests for that responsibility alone. A PR that changes an already-merged block states why. Behaviour-preserving parts pin the current behaviour in tests before moving code, and keep those tests green through the move.

## Not in scope

Multi-node live, new telemetry, and changes to the collection pipeline. The data collected is unchanged throughout; this series is about where the reading, the deciding and the drawing live.

PR #410, opened 2026-08-28
Process dashboard: repository, typed models, and compute/view boundaries

### LINKED ISSUE #404: Process dashboard: repository, typed models, and compute/view boundaries (behaviour-preserving)
Part 1 of #403, the architecture series from the review of #401.

**Goal.** Restructure the Process dashboard path into explicit layers with no behaviour change:

```
SQLite -> ProcessRepository -> ProcessDashboardComputer -> ProcessDashboardPayload -> process_section.py
```

**Scope**
- [ ] `renderers/process/repository.py`: owns all `process_samples` reads (moved, not rewritten)
- [ ] `renderers/process/dashboard_models.py`: typed payload contracts replacing nested dicts
- [ ] Compute layer owns the percentile/rollup/window arithmetic that currently runs inside the view
- [ ] `process_section.py` becomes presentation only: components, formatting, chart options
- [ ] Characterization tests pinning current behaviour BEFORE the move, kept green through it

**Non-goals.** No whole-run behaviour, no freshness policy, no UI redesign, no System or `theme.py` changes.

**Done means.** Existing dashboard renders identically; the view contains no SQL and no metric arithmetic; the repository makes no policy decision; full suite green.

Pendu:
Closes #404. Part 1 of #403.

## Review scope

The Process dashboard path only. This PR moves code between layers and changes no metric, no
query result and no rendering. The other panes in any screenshot are today's rendering and are
out of scope here.

`system_section.py` and `theme.py` are deliberately untouched, per #403 part 1.

## What

The Process dashboard path becomes four explicit layers:

```
SQLite -> ProcessRepository -> ProcessDashboardComputer -> ProcessDashboardPayload -> process_section.py
```

| file | before | after | what it owns now |
|---|---|---|---|
| `renderers/process/repository.py` | new | 252 | every `process_samples` read, and nothing else |
| `renderers/process/dashboard_models.py` | new | 132 | the typed contract between compute and the card |
| `renderers/process/dashboard_compute.py` | 203 | 269 | what the numbers mean: rollups, percentiles, the window |
| `renderers/process/common.py` | 293 | 33 | the terminal snapshot contract only |
| `nicegui_sections/process_section.py` | 176 | 153 | components, units, colours, chart options |

## Why

Three things sat on the wrong side of a boundary on `version_0.3.7`:

**The view was computing.** `process_section.py` ran `_percentile` and `_compute_rollups` on
every tick, so what a metric MEANS was decided inside the drawing code. Those moved to the
compute layer unchanged.

**Absence was inferred from key presence.** A history entry carried `gpu_used`, `gpu_total`,
`gpu_headroom` and `gpu_rank` only when a GPU had reported, so the card learned whether the run
had a GPU by asking whether a dictionary key existed. That is now `gpu: Optional[GpuSnapshot]`
on a typed entry: one named field, and a `None` that cannot be confused with a zero.

**The SQL had no home.** `ProcessMetricsDB` lived in `common.py` beside the payload dataclasses.
It moved to `repository.py` as `ProcessRepository` with the same seven methods and no logic
change: every method body is byte-identical to the original, and the ASTs match method for
method. Nothing about the reads was rewritten while they were relocated.

## Behaviour

Preserved, and measured rather than asserted. The same 13 fixture cells were rendered on
`version_0.3.7` and on this branch, and the Process card's text is identical on every one:

| cells | result |
|---|---|
| B (1 GPU), C (N GPU 1 process), D (N GPU N ranks), E (2 nodes), W (watch) | 11 of 11 identical |
| D' and E' (a rank dead mid-run) | 2 of 2 identical |

### The one exception, deliberate

A step whose `sample_ts_s` is missing is now dropped from the chart on its own, instead of
taking the rest of the run with it.

On `version_0.3.7` the card zipped a timestamp list that had already had its gaps filtered out
against an unfiltered window, behind a length check. Both branches of that check are wrong, and
both were measured on this branch before it changed:

- with few samples the check fails and the chart draws **nothing**: 0 points plotted where 2
  were plottable
- with enough samples the check passes and the two lists are zipped anyway, so every value lands
  on someone else's moment: on a 150-step history with 5 gaps, the first RAM value belonged to
  step 51 and was drawn at step 46

Points now carry their own timestamp. Reproducing a silent misalignment to keep the phrase
"no behaviour change" would have been the worse trade. Pinned by
`test_a_gap_in_the_timestamps_drops_only_that_sample`, whose docstring carries both numbers.

## How this was made safe

The Process dashboard path had no dedicated tests on `version_0.3.7`, so the first commit is 48
characterization tests written against the existing code and proven green on it before anything
moved: 14 on the read layer, 14 on compute, 20 on the card. A reviewer can check out that commit
and watch all 48 pass on untouched code.

The second commit moves the code. The read-layer tests survive with only an import changed. The
percentile and rollup tests move from the card file to the compute file, because the arithmetic
they cover moved; their expected values are unchanged. The card's own tests keep their on-screen
assertions byte for byte and only change how the payload reaching them is built.

`test_the_card_renders_the_same_from_a_real_database` closes the loop end to end, so the unit
tests on either side of the new boundary cannot both pass while the join between them is broken.

## Scope

```
STRUCTURE: process dashboard reads -> renderers/process/repository.py (owner of process_samples);
mirrored from the existing per-domain renderer packages; boundaries crossed: none
DATA PATH: process_section <- PROCESS_LAYOUT <- ProcessRenderer
           <- ProcessDashboardComputer <- ProcessRepository <- process_samples
SCOPE: declared renderers/process + nicegui_sections/process_section.py + their tests;
diff touches exactly those -> within
```

## Verification

```
GATE: pytest tests/ -> 1383 passed, 2 skipped; 52 cover this block (14 read, 26 compute, 12 card); the first commit adds 48 of them against untouched 0.3.7 code and passes there; black/isort/ruff clean at 79
TRANSITIONS: cold (no rows) -> populated -> capped at the retained maximum -> read failure served from the last good payload -> that payload expiring to empty; every degraded path returns an empty typed payload rather than a partial one
RANKS: 1 rank, 2 ranks, 4 ranks, and a rank dead mid-run (D', E'); the committed-seq rule that holds history at the slowest rank is pinned by test_committed_seq_is_the_slowest_rank
TRIPWIRE: the read layer is pinned by 14 tests including the least-headroom GPU choice and the partial-GPU-row rule; the compute layer by 26; a table rename in the writer breaks the fixture, because the schema comes from init_schema rather than a hand-written copy
SCENARIOS: 13 fixture cells rendered on version_0.3.7 and on this branch, Process card text identical on all 13 (B, C, D, E, W, D', E')
```

## Not in this PR

Per #403: no whole-run history, no freshness policy, no caching, no UI change, no System or
`theme.py` change. `RankSnapshot` and the explicit `recent`/`retained` flag land in #406, where
per-rank data and a retained mode first exist to describe; naming them here would have meant
adding types with nothing to hold.


abhinavsriva (approved):
I add a small check otherwise all okay.

Issue #408, opened 2026-08-28
System dashboard: characterization tests and repository extraction (part 5a-i)


Pendu:
Part 5a-i of #403. The System block's SQLite reads move to their own file, and current behaviour
is pinned first so the move is provably a move.

Split from the original 5a, which also covered the typed payload and the view's interpretation.
That scope was heading for roughly 2500 to 3000 lines against an 888-line view; #410 is the same
work for Process and landed at about 2100 against a 154-line view. The redesign is now #420, and
#419 consumes `renderers/shared` after both.

**Scope**
- [ ] Characterization tests first, same method as #404: pin what the card decides today,
      including the rules that look wrong, so any change to them fails loudly
- [ ] Extract `renderers/system/repository.py` from `common.py`: every SQLite read, mirroring
      `renderers/process/repository.py`

**Non-goals.** No typed payload and no interpretation moves (#420). No consumption of
`renderers/shared` (#419). No visual change. No change to what any number means.

**Constraints that shape the extraction**, proven by patching and re-running:
- `tests/display/test_system_section.py:647` imports `choose_window_s` from
  `renderers.system.common` by name; six tests fail if it moves, so the window ladder stays put
  and travels with #419 instead
- `tests/renderers/test_system_dashboard_gpu_payload.py:370-378` monkeypatches
  `SystemDashboardComputer._db.fetch_gpu_power_run_history` to prove a CPU-only run never issues
  the GPU query, so neither the attribute nor the method may be renamed

Pendu:
Shipped in #421, merged as `5093cbb` on `version_0.3.7`.

`SystemMetricsDB` became `SystemRepository` in its own module, taken across without edits, and
`common.py` went from 500 lines to 89. Nine characterization tests were written first, against
untouched code, so the move is provably a move.

Two things this issue listed were deliberately not resolved here and were carried into #420: the
two-GPU outlier tie, and the card and the computer disagreeing about "this GPU never reported".

Closing by hand: the `Closes` keyword does not fire for PRs merged into a release branch rather
than the default branch.

PR #411, opened 2026-08-28
Shared run-series and freshness infrastructure (renderers/shared)

### LINKED ISSUE #405: Shared run-series and freshness infrastructure (renderers/shared)
Part 2 of #403. Generalizes machinery that exists today as per-package copies (`choose_window_s` is
byte-identical in `renderers/system/` and the Process branch) into one shared home, so Process
and System consume one implementation instead of forking it.

**Scope**
- [ ] `renderers/shared/run_series.py`: recent-vs-retained selection, rolling-window selection,
      cadence and stride planning, point-budget enforcement; produces a `RunSeriesPlan`
- [ ] `renderers/shared/freshness.py`: one sampler-aware telemetry freshness policy, fed the
      configured sampler interval or an explicit observed cadence
- [ ] Telemetry freshness and last-good TTL are separate concepts with separate types
- [ ] Focused unit tests: window planning, point budgets, cadence, freshness thresholds

**Non-goals.** No SQLite table names, no Process/System metric definitions, no NiceGUI code, no
arbitrary SQL strings. Repositories execute their own table-specific SQL using the plan.

**Done means.** Both modules are importable and fully tested standalone; nothing consumes them
yet (wiring is #406 and #408); the sampler interval enters as an input, never a hardcoded constant.

Pendu:
Closes #405. Part 2 of #403.

## Review scope

Two new modules under `renderers/shared/` and their tests. Nothing consumes them yet: Process
wires them in #406 and System in #408. No existing file changes, so no existing behaviour can.

## What

| module | owns |
|---|---|
| `renderers/shared/run_series.py` | how a run-length series is sampled: the rolling window, recent vs retained, cadence, stride, the point budget |
| `renderers/shared/freshness.py` | when telemetry stops counting as current, and separately how long a cached payload may stand in for a failed read |

## Why these two, and why now

They are not new ideas. They are the machinery `renderers/system/` and the Process work had each
grown a copy of, generalised once before a third copy appears.

`choose_window_s` is **byte-identical** in `renderers/system/common.py` and in the Process
branch. So is `_return_stale`. `_opt_float` is 95% the same. The shared window here is verified
to return the same value as the System original across 14 run lengths from 0 to 24 hours, so
#408 has a proven-equal starting point when System is migrated onto it.

### The freshness split

Two different questions were both called "stale", with a hardcoded two-second assumption in
several modules:

- **telemetry freshness** asks whether a rank is still reporting, and is answered from the
  sampling cadence
- **cached-payload TTL** asks how long a previously good payload may answer for a read that
  failed, and says nothing at all about the health of the run

They are separate types here, and the separation is structural rather than documented:
`CachedPayloadTTL` has no `is_stale`, `FreshnessPolicy` has no `may_reuse`. A test proves a
one-hour cache TTL still cannot make a rank that has been silent for ten minutes look live.

The sampler interval enters as a value, from the configured setting or from the cadence actually
observed, and the observed one wins when both are available: what was configured is an
intention, what arrived is a measurement.

## Non-finite input

`to_ms` in #410 needed an `isfinite` guard, and the same gap was in both of these modules.

NaN clears every ordinary check, because `nan <= 0` is False. Measured before the guard existed:
`plan_run_series` with a corrupt cadence raised `ValueError: cannot convert float NaN to integer`
at the `int()` inside `preceding_rows`, and `is_stale(nan)` returned False, so a corrupt
timestamp made a dead rank read as live.

Non-finite values are now rejected at the boundary, which is what the rest of the codebase
already does (`renderers/step_time/renderer.py`, `diagnostics/system/policy.py`,
`step_time/sqlite.py` and seven other sites). `RunSeriesPlan` is guarded a second time because
it is public and a plan built by hand must not be able to take an `int()` down with it.

## Scope

```
STRUCTURE: run-series and freshness policy -> renderers/shared (owner of cross-domain read
policy); new unit, no sibling to mirror; boundaries crossed: none
SCOPE: declared renderers/shared + its tests; diff touches exactly those -> within
```

`run_series.py` holds no table name, no column name, no metric definition and no display code.
It plans; each repository executes its own SQL against the plan. The one SQL-shaped thing it
emits is the window frame, which is fully determined by the window and the cadence; leaving that
to each caller is what produced two copies of it in the first place.

## Verification

```
GATE: pytest tests/ -> 1441 passed, 2 skipped; 58 of those are these two modules; black/isort/ruff clean at 79; every pre-commit hook passes
TRANSITIONS: recent -> retained at 1.2x the window, with the hysteresis tested on both sides of the boundary so a chart cannot flip each tick; plan -> no plan below two samples; finite -> non-finite rejected at every entry point
RANKS: not rank-aware by design; this layer plans a series and judges an age, and is consumed per rank by #406 and per node by #408
TRIPWIRE: the point budget is asserted to hold across run lengths of 200 to 50000 samples, so a floor-vs-ceiling regression in the stride fails the suite; the window ladder is asserted monotonic and within its policy bounds; a test proves a long cache TTL cannot mask a stale rank
SCENARIOS: no rendering in this PR, nothing to screenshot; the shared window is proven equal to the System original across 14 run lengths
```

## Not in this PR

No consumer. Wiring Process is #406 and System is #408, so a regression in either lands in the
PR that causes it rather than here.

### DIFF
--- a/src/traceml_ai/renderers/shared/freshness.py
+++ b/src/traceml_ai/renderers/shared/freshness.py
@@ -1,0 +1,145 @@
+# Copyright 2026 OptAI UG (haftungsbeschraenkt)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Two different questions that were both being called "stale".
+
+**Telemetry freshness** asks whether a rank or a node is still reporting.
+It is answered from the sampling cadence: a rank that has missed several
+ticks has stopped, and the card must say so rather than keep drawing its
+last value as current.
+
+**Cached-payload TTL** asks how long a previously good payload may stand in
+for a read that just failed. It is answered from how long a reader will
+tolerate an unchanged view, and it says nothing whatsoever about the health
+of the run.
+
+They were one ambiguous notion of "stale" spread across several modules,
+each with its own hardcoded two-second assumption. They are separate types
+here so a caller has to say which one it means, and so the sampler interval
+enters as a value rather than a constant.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Optional
+
+# Ticks a rank may miss before the block stops calling it live. Three is
+# what the Process and System blocks converged on: one missed tick is
+# ordinary jitter, three is a rank that stopped.
+DEFAULT_STALE_TICKS = 3.0
+
+# However fast the sampler is configured, nothing is called stale before
+# this. Protects a sub-second cadence from flagging on scheduler noise.
+MIN_STALE_AFTER_S = 5.0
+
+
+@dataclass(frozen=True)
+class FreshnessPolicy:
+    """When telemetry from one entity stops counting as current.
+
+    Built from the cadence the samples actually arrive at, so a run
+    configured with a slow sampler is not permanently "stale" and a fast
+    one is not judged against a stale-by-default constant.
+    """
+
+    interval_s: float
+    stale_ticks: float = DEFAULT_STALE_TICKS
+    min_stale_after_s: float = MIN_STALE_AFTER_S
+
+    @classmethod
+    def from_interval(
+        cls,
+        interval_s: Optional[float],
+        *,
+        fallback_s: float = 2.0,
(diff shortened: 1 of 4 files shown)


abhinavsriva (approved):
Done. minor change fixed but overall clean and easier to follow. thanks!

Issue #407, opened 2026-08-28
Process dashboard presentation


Pendu:
Part 4 of #403. The #401 UI, consuming the completed typed payload without changing computation.

**Scope**
- [ ] Four tiles (CPU capacity worst rank, RSS worst rank, CUDA reserved least headroom, CUDA allocated)
- [ ] Two per-rank charts on one shared axis; per-rank rows with the engine-owned expansion signal
- [ ] Scope line in the header (one process per rank; DataLoader workers not included)
- [ ] New chart helpers land in `charting.py` / `formatting.py`, not `theme.py`
- [ ] Presentation tests: layout, formatting, degraded states; scenario-matrix render across the
      fixture cells

**Non-goals.** No computation changes; no System changes.

PR #412, opened 2026-08-28
Process dashboard: per-rank facts, freshness, normalized CPU rollups, and whole-run series

### LINKED ISSUE #406: Process compute and payload features

Pendu:
Closes #406. Part 3 of #403.

Part 2 (#411) is merged, so this branch is rebased straight onto `version_0.3.7` and is no
longer stacked. Review it against the release branch:

```
git diff upstream/version_0.3.7...dev/abhijeet/process-features
```

4 files, +1079/-12.

## Review scope

The Process compute layer and its payload. **No UI change**: `process_section.py` is untouched
and the card renders exactly what it rendered before this PR. That is asserted, not assumed, and
the assertion is described below.

## What

The data-layer content of the closed #401, rebuilt on part 1's boundaries and part 2's shared
policies.

| added | what it is |
|---|---|
| `RankSnapshot` | one rank's own state, read on its own clock |
| `RankCoverage` | total, live, stale and unknown rank counts |
| `RankChart` / `RankTrace` | per-rank series, with the mode stated |
| `cpu_capacity`, `rss_worst`, `gpu_reserved` | the per-rank rollups, as NEW fields |
| `reserved_imbalance_percent` | spread of reserved memory across ranks |

### Per-rank history, on each rank's own clock

Two reads rather than one. The windowed read carries each rank's recent history; the
latest-row read carries every rank that has ever reported. Without the second, a rank silent for
longer than the window has no rows in the first and drops out of the block entirely, so the
surface forgets a dead rank a few minutes after it dies, which is when its death starts to
matter.

### CPU is normalized

`psutil` reports process CPU summed across cores, so a healthy rank on an 8-core host reads
200%. Divided by its core count that is 25% of the host, which is bounded and comparable
between ranks on different machines.

### The rollups name a rank

The worst CPU rank is named, because a bottleneck finder has to answer "which rank" and a cohort
median cannot. The worst RSS rank is chosen on its window MEDIAN but displayed at its NEWEST
value: which rank is worst is a judgement about its typical state, while the number shown should
be what that rank last actually sent. The CUDA rollup reports the rank with the least headroom
rather than the largest user, because the process at risk is the one closest to filling its card.

Imbalance is computed on RESERVED memory, not allocated. The allocator's live bytes are a
sawtooth that this sampling cadence undersamples, so their across-rank spread is phase noise on a
healthy run; reserved is what the process actually holds.

### The series mode is stated

`RankChart.mode` is `recent` or `retained`, on the payload. The previous shape carried a window
series and a whole-run series in two differently named fields and left the view to work out which
to draw from whichever happened to be non-empty. That is a rule no reader can see.

### Caching, with its reason

`RUN_REFRESH_S` refreshes the whole-run reads on a slower clock than the tick. A rolling mean
over minutes cannot visibly change between two 2 s ticks, and recomputing these reads every tick
was the largest cost measured in this block. That is the demonstrated need; the policy is one
constant with the reason next to it.

## No SQL crosses the boundary

The whole-run reads are `fetch_cpu_capacity_run(plan)` and `fetch_rss_run(plan)`. Interpolation
happens in two private helpers that are only ever handed one of this module's own constants, so
no expression reaches the repository from outside the class. Asserted directly: no public method
on `ProcessRepository` takes a parameter whose name contains `sql`.

The window, cadence, stride and point budget are decided by part 2 and arrive as a
`RunSeriesPlan`; the repository executes, it does not plan.

## The card is unchanged, and that is tested

The new rollups are NEW fields. `cpu`, `ram` and `gpu` keep the meanings the card already reads.
Repurposing them would have changed the card silently, which is part 4's job to do openly.

This matters more than it sounds. Two of them nearly moved:

- the GPU tile went from **allocated** to **reserved**, `6.00 GB` to `7.00 GB`
- the CPU tile would have gone from raw `700%` to normalized `87.5%` on any run reporting a core
  count

Both were caught by `test_the_card_renders_the_same_from_a_real_database`, the end-to-end test
added in part 1, which runs database to compute to card and asserts the on-screen strings. A
screenshot could not have caught either: a tile reading `7.00 GB` looks entirely correct.

## A defect found while re-landing

The whole-run stride was planned on the AVERAGE rows per rank (`count // ranks`) and then applied
PER rank, so any rank with more rows than the mean overshot the point budget. Measured on a
four-rank run where one rank stopped early:

```
    stated cap: 120 points
    rank 0: 145 points   <- over
    rank 1: 145 points   <- over
    rank 2: 145 points   <- over
    rank 3:  45 points
```

Unequal row counts are the ordinary case, not an edge one: a rank that dies, joins late, or
samples slower produces them. The plan is now made on the busiest rank's count, and a test
asserts the budget holds for every rank rather than on average.

This is the same class as the floor-vs-ceiling stride fix on #389. That fix corrected the
divisor; the dividend was wrong too and stayed wrong. Worth naming so the next stride change
checks both.

## Adapting to the review fix on part 2

Part 2 landed with `is_stale() -> bool` replaced by `state_of() -> "fresh" | "stale" | "unknown"`.
This PR is that API's first consumer, so it had to decide what the third state means here.

Collapsing it back to a boolean would have undone the change at the first call site: an unknown
age would have become `stale=False` and a rank with no usable clock would have been counted as
live, which is the ambiguity the new signature exists to remove. So the state is carried through
to the payload. `RankSnapshot.freshness` holds it and `RankCoverage` counts it in its own bucket,
which keeps `live + stale + unknown == total` true by construction.

An unknown age still counts toward the aggregates. A missing timestamp is a clock problem, not
evidence that a rank stopped working, and dropping its reading would discard real data on that
basis.

Worth stating plainly: this state is not reachable from our own writer. `recv_ts_ns` is
`INTEGER NOT NULL` (`aggregator/sqlite_writers/process.py:146`), so the age is always computable
for a row we wrote. It is a guard, not a live path, and the test says so and asserts it against
constructed snapshots rather than pretending a database can produce one.

## Scope

```
STRUCTURE: per-rank facts and whole-run series -> renderers/process (owner of process_samples
and its meaning); shared policy consumed from renderers/shared; boundaries crossed: none
DATA PATH: process_section <- PROCESS_LAYOUT <- ProcessRenderer <- ProcessDashboardComputer
           <- ProcessRepository <- process_samples
           <- renderers/shared/{run_series,freshness} for window and freshness policy
SCOPE: declared renderers/process + its tests; diff against version_0.3.7 touches exactly those 4 files -> within
```

## Verification

```
GATE: pytest tests/ -> 1458 passed, 2 skipped; 57 on the Process renderer, 17 of them new here; isort, ruff and black clean at 79 under the repo's pinned pre-commit; every hook passes
TRANSITIONS: recent -> retained once the run outgrows its window; live -> stale per rank on the observed cadence; populated -> read failure served from the last-good payload -> that payload expiring; a rank present in the window read but absent from it later still appears via the latest-row read
RANKS: 1, 2, 3 and 4 ranks, plus a rank that stops mid-run; aggregates describe the live ranks while the stopped one keeps its row and its age; single-rank runs report no imbalance rather than zero; a rank with no usable arrival clock is counted apart from both
TRIPWIRE: the point budget is asserted per rank, not on average, so the 145-vs-120 defect above fails the suite; a test asserts the card's fields keep their meaning; no public repository method may take a SQL parameter; the payload tests import no NiceGUI
SCENARIOS: no rendering change in this PR, so no new screenshots; the card is proven unchanged by the end-to-end test rather than by a picture
```

## Not in this PR

The card. Part 4 (#407) switches the tiles to `cpu_capacity`, `rss_worst` and `gpu_reserved`,
splits the single ambiguous `GPU MEM` tile into the `allocated` and `reserved` pair that #399
settled the wording for, and draws the per-rank charts. Until then those fields are computed,
tested and unread, which is deliberate: a regression in the card then lands in the PR that
causes it.

### DIFF
--- a/src/traceml_ai/renderers/process/dashboard_compute.py
+++ b/src/traceml_ai/renderers/process/dashboard_compute.py
@@ -26,7 +26,19 @@ from __future__ import annotations
 
 import time
 from collections import deque
-from typing import Any, Deque, List, Optional, Sequence, Tuple
+from dataclasses import replace
+from typing import Any, Deque, Dict, List, Optional, Sequence, Tuple
+
+from traceml_ai.renderers.shared.freshness import (
+    CachedPayloadTTL,
+    FreshnessPolicy,
+)
+from traceml_ai.renderers.shared.run_series import (
+    DEFAULT_RUN_SERIES_POLICY,
+    RunSeriesPolicy,
+    finite,
+    plan_run_series,
+)
 
 from .dashboard_models import (
     ChartSeries,
@@ -35,12 +47,25 @@ from .dashboard_models import (
     MetricRollup,
     ProcessDashboardPayload,
     ProcessHistoryEntry,
+    RankChart,
+    RankCoverage,
+    RankSnapshot,
+    RankTrace,
 )
 from .repository import ProcessRepository
 
 # How many committed steps the card describes. Kept at the value the
 # section applied to its own history slice on 0.3.7.
 DASHBOARD_WINDOW = 100
+
+# Samples per rank in the recent-window view.
+RANK_WINDOW_N = 100
+
+# A rolling mean over minutes cannot visibly change between two ticks, so
+# the whole-run reads refresh on their own slower clock. Recomputing them
+# every tick was the single largest cost measured in this block, which is
+# the demonstrated need this cache exists for.
+RUN_REFRESH_S = 15.0
 
 
 def percentile(values: Sequence[Optional[float]], p: float) -> float:
@@ -58,6 +83,56 @@ def percentile(values: Sequence[Optional[float]], p: float) -> float:
     if low == high:
         return float(clean[low])
     return float(clean[low] * (high - k) + clean[high] * (k - low))
+
+
+def _opt_float(value: Any) -> Optional[float]:
+    """A usable number from a database cell, or ``None``."""
+    if value is None:
+        return None
+    try:
+        return finite(float(value))
+    except (TypeError, ValueError):
+        return None
+
+
+def _median(values: Sequence[float]) -> Optional[float]:
+    """The middle value, or ``None`` when there is nothing to take."""
+    clean = sorted(v for v in values if v is not None)
+    if not clean:
+        return None
+    middle = len(clean) // 2
+    if len(clean) % 2:
+        return float(clean[middle])
+    return float((clean[middle - 1] + clean[middle]) / 2.0)
+
+
+def _gpu_reported(row: Any) -> bool:
+    """Whether one row carries a usable GPU reading.
+
+    ``gpu_available`` alone is not enough: a rank reports it as true while
+    torch is still coming up, before any capacity is known.
+    """
+    try:
+        if not row["gpu_available"]:
+            return False
+        total = _opt_float(row["gpu_mem_total_bytes"])
+        return total is not None and total > 0
+    except (KeyError, IndexError, TypeError):
+        return False
+
+
+def _cpu_capacity_of(row: Any) -> Optional[float]:
+    """CPU as a share of the host's cores, not a sum across them.
+
+    ``psutil`` reports process CPU summed over cores, so a healthy
+    four-core rank reads 291%. Divided by the core count the number is
+    bounded and comparable between ranks on different machines.
+    """
+    used = _opt_float(row["cpu_percent"])
+    cores = _opt_float(row["cpu_logical_core_count"])
+    if used is None or cores is None or cores <= 0:
+        return None
+    return used / cores
 
 
 class ProcessDashboardComputer:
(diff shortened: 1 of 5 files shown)


abhinavsriva (approved):
Some small changes were made otherwise all okay

PR #413, opened 2026-08-29
Dev/abhinav/timing flow process

abhinavsriva:
Process freshness: wire configured sampler interval

The shared freshness policy already accepts the configured sampler interval, but the production Process renderer path did not pass it through.

This PR wires:
```
TraceMLSettings
→ ProcessRenderer
→ ProcessMetricsComputer
→ ProcessDashboardComputer
→ FreshnessPolicy
```

Observed cadence still takes priority once enough samples arrive. The configured interval is only the startup fallback, preventing slow custom samplers from being incorrectly marked stale before their next sample is due.
No freshness calculations, payloads, repository logic, or UI behavior were redesigned.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/cli.py
+++ b/src/traceml_ai/aggregator/display_drivers/cli.py
@@ -86,7 +86,10 @@ class CLIDisplayDriver(BaseDisplayDriver):
         # Watch profile
         self._renderers: List[CLIRenderer] = [
             SystemRenderer(db_path=self._settings.db_path),
-            ProcessRenderer(db_path=self._settings.db_path),
+            ProcessRenderer(
+                db_path=self._settings.db_path,
+                sampler_interval_s=self._settings.sampler_interval_sec,
+            ),
             StdoutStderrRenderer(db_path=self._settings.db_path),
         ]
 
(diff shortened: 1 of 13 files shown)


Pendu:
Read this one against the base and could not find a way to break it. Clean from my side. Two notes, neither blocking.

## Verified

- Full suite `1466 passed, 2 skipped` at head against `1465 passed, 2 skipped` at the merge base, so exactly the one added test and no regressions.
- The wiring reaches the policy. Constructing `ProcessMetricsComputer(db_path=..., sampler_interval_s=60.0)` gives `_dashboard._configured_interval_s = 60.0`, and `dashboard_compute.py:158,164` carries it into `FreshnessPolicy`. The startup fallback behaves as described, with observed cadence still taking over once samples arrive.

## The `cli.py` hunk has no effect today

`ProcessMetricsComputer` forwards `sampler_interval_s` only to the dashboard computer (`computer.py:50-54`); the CLI computer is constructed without it at `:46`. The CLI display driver only ever calls `compute_cli()`, so the value it now passes is never read on that path.

I checked whether that means the CLI surface still has the problem this PR describes, and it does not. `ProcessCLIComputer` has no freshness concept at all: grepping it for `freshness` or `sampler_interval` returns zero hits, and its only staleness notion is `stale_ttl_s`, which bounds how long a last-good snapshot may be reused rather than judging a rank stale. So there is nothing there for the interval to fix.

That makes the hunk harmless and consistent rather than functional. Worth saying out loud only so nobody later reads `cli.py` and concludes CLI freshness is configurable.

## The test stops one step short of the claim

`test_driver_passes_sampler_interval_to_process_renderer` asserts the nicegui driver hands `sampler_interval_s` to `ProcessRenderer`. The behaviour that actually matters is that the value reaches `FreshnessPolicy`, which the test does not reach: a regression that dropped the argument inside `ProcessMetricsComputer` or `ProcessDashboardComputer` would keep this test green.

`cli.py` received the identical edit with no test, though there is no CLI display-driver test file in `tests/display/` at all, so that is a gap in the suite rather than something this PR skipped.

PR #414, opened 2026-08-29
Dev/abhinav/issue394 3

abhinavsriva:
## Summary

Capture step-memory timestamps when the memory measurement occurs at step end, rather than later when the sampler drains its queue.

## Why

Step-memory events can remain queued before the sampler processes them. Using the queue-drain time makes the timestamp dependent on sampler scheduling and can misalign memory measurements with their training steps and other telemetry.

The event producer is the authoritative source of measurement time, so the timestamp is now captured there and preserved through sampling and serialization.

## Changes

- Add a required `timestamp` to `StepMemoryEvent`.
- Capture the timestamp when step memory is recorded.
- Preserve the event timestamp when creating `StepMemorySample`.
- Keep the timestamp required throughout the internal schema.
- Remove the fabricated `0.0` fallback when reading wire data without `ts`.
- Add focused tests covering timestamp capture, sampler preservation, and wire round-trip behavior.

## Design

`StepMemoryEvent` is an internal typed contract with a single producer, so the sampler accesses its fields directly. Missing external wire fields are rejected at the wire boundary instead of introducing nullable timestamps throughout downstream logic.

## Scope

This PR changes step-memory timestamp semantics only.

It does not modify:

- Telemetry windowing or aggregation
- SQLite retention
- Dashboard or terminal rendering
- Diagnosis behavior

### DIFF
--- a/src/traceml_ai/samplers/schema/step_memory.py
+++ b/src/traceml_ai/samplers/schema/step_memory.py
@@ -74,7 +74,7 @@ class StepMemorySample:
         """
         return StepMemorySample(
             sample_idx=data["seq"],
-            timestamp=float(data.get("ts", 0.0)),
+            timestamp=float(data["ts"]),
             model_id=data.get("model_id"),
             device=data.get("device"),
             step=data.get("step"),
--- a/src/traceml_ai/samplers/step_memory_sampler.py
+++ b/src/traceml_ai/samplers/step_memory_sampler.py
@@ -1,12 +1,9 @@
 from __future__ import annotations
 
-import time
-from typing import Any, Optional
-
 from traceml_ai.samplers.base_sampler import BaseSampler
 from traceml_ai.samplers.schema.step_memory import StepMemorySample
 from traceml_ai.samplers.utils import drain_queue_nowait
-from traceml_ai.utils.step_memory import step_memory_queue
+from traceml_ai.utils.step_memory import StepMemoryEvent, step_memory_queue
 
 
 class StepMemorySampler(BaseSampler):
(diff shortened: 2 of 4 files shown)


Pendu:
This is a real correctness fix and the risky parts hold up. Clean from my side, with one note on how the wire change should be read.

## Verified

- Full suite `1468 passed, 2 skipped` at head against `1465 passed, 2 skipped` at the merge base, so exactly the three added tests and no regressions.
- The fix does what it says. I drove `_event_to_sample` on both branches with a measurement at t=1000 and the drain 45 seconds later:

```
BASE  measured 1000.0, drained 1045.0 -> sample.timestamp = 1045.0   (drain time)
HEAD  measured 1000.0, drained 1045.0 -> sample.timestamp = 1000.0   (measurement time)
```

  The old timestamp absorbed the whole queue latency, which is exactly the misalignment described.

- The buffered event keeps its timestamp. `flush_step_memory_buffer` mutates only `evt.step` at `utils/step_memory.py:111` and queues the same object at `:113`, so the value captured at `:96` survives the deferred step fill.
- Adding a required field mid-dataclass is safe here. `StepMemoryEvent` is constructed with keywords at both of its sites, so inserting `timestamp` ahead of the two optional fields cannot silently shift a positional argument.
- Retention is unaffected. `step_memory_samples` sits in `_STEP_HISTORY_TABLES` (`sqlite_writer.py:58`), so it is pruned by step rather than by time. Earlier timestamps cannot age rows out sooner.
- Dropping the defensive `getattr` calls is justified. `step_memory_queue.put_nowait` appears exactly once in the tree, at `utils/step_memory.py:113`, so the typed contract really does have a single producer, and `sample()` still wraps `_drain_queue()` in try/except so a malformed event could not reach training.

## How to read the wire change

`from_wire` moving from `data.get("ts", 0.0)` to `data["ts"]` is right, and removing a fabricated zero is consistent with the rest of this series. One thing worth stating so the description is not read as stronger than it is: `StepMemorySample.from_wire` has no caller anywhere in `src/`.

The path that actually ingests step-memory payloads is `sqlite_writers/step_memory.py:235,243`, which reads the dict directly and already resolves a missing `ts` to `None` rather than `0.0`. That writer is untouched here.

So nothing in production begins rejecting payloads with this change; the aggregator continues to store NULL as it did before. What improved is an SDK and test facing helper that previously fabricated a zero. Good change either way, but "rejected at the wire boundary" reads like an ingest guarantee, and the ingest boundary is elsewhere and unchanged.

PR #415, opened 2026-08-30
native logger used instead of forced logging

abhinavsriva:
Why
TraceML was converting user failures into its own exit behavior, which could hide native Python tracebacks and interfere with torchrun error reporting.

What changed
- Re-raise user exceptions unchanged so Python displays the original traceback.
- Preserve native SystemExit and KeyboardInterrupt behavior.
- Use TorchElastic’s error recorder when launched through torchrun.
- Keep the existing error log as a temporary compatibility fallback.
- Ensure TraceML cleanup remains best-effort and cannot mask user failures.

Testing
- Verified native traceback and exit-code propagation.
- Verified SystemExit, KeyboardInterrupt, and TorchElastic integration.
- 14 passed.

### DIFF
--- a/src/traceml_ai/runtime/executor.py
+++ b/src/traceml_ai/runtime/executor.py
@@ -4,7 +4,11 @@
 # you may not use this file except in compliance with the License.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Execute a user script inside the TraceML runtime."""
+"""Run a user script with a best-effort TraceML runtime around it.
+
+The executor is intentionally transparent to user failures: Python and
+torchrun remain responsible for exception reporting and process exit status.
+"""
 
 import os
 import runpy
@@ -12,7 +16,7 @@ import sys
 import traceback
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 from traceml_ai.runtime.launch_context import (
     LaunchContext,
@@ -33,7 +37,6 @@ from traceml_ai.telemetry.retention import (
     parse_history_retention,
 )
 
-INTERRUPTED_EXIT_CODE = 130
 DEFAULT_LOGS_DIR = "./logs"
 DEFAULT_AGGREGATOR_HOST = "127.0.0.1"
 DEFAULT_AGGREGATOR_BIND_HOST = "127.0.0.1"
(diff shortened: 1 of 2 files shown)


Pendu:
Went at this one hardest of the batch, because it changes how user failures reach the shell. The core is right and I could not find a case where an exit code moved. I pushed one fix, and have two notes that are yours to rule on.

## Verified

- The headline claim holds end to end. I ran the real executor as a subprocess on both branches across five exit paths, changing nothing but the checkout:

```
                     BASE            HEAD
RuntimeError         exit 1,  no tb  exit 1,  traceback
SystemExit(7)        exit 7          exit 7
SystemExit(0)        exit 0          exit 0
KeyboardInterrupt    exit 130        exit 130, traceback
clean exit           exit 0          exit 0
```

  Every code is preserved, and the two cases that previously vanished into the log now print a real traceback. `KeyboardInterrupt` still yields 130 through Python's native handling rather than the removed constant, so dropping `INTERRUPTED_EXIT_CODE` costs nothing.

- Cleanup cannot swallow an ordinary failure. `stop_runtime` catches `Exception` and routes it to the runtime log (`executor.py:324-336`), so a shutdown error inside the new `finally` cannot replace the user's exception.
- The lazy torchelastic import is safe. With no `TORCHELASTIC_ERROR_FILE` the entrypoint runs directly, and a missing or older torch falls back through `ImportError` rather than blocking the run.
- Full suite `1469 passed, 2 skipped` with the fix below applied.

## Pushed (`ba7886a`)

`tests/runtime/test_executor.py` computed `ROOT = Path(__file__).resolve().parents[1]`, which is `tests/`, so `SRC` was `tests/src` and did not exist. The new subprocess test handed that non-existent path to the child as `PYTHONPATH`, so the child silently imported whatever `traceml_ai` the ambient environment resolved rather than this branch.

In a checkout where that resolves elsewhere, the child ran different code, swallowed the traceback, and the test failed on `assert "Traceback (most recent call last)" in result.stderr` with an empty stderr. The more interesting half is that when it passes, it was not necessarily testing this branch either.

`parents[2]` points `SRC` at the repo's own `src/`. That is the whole change, one line, and it takes the file from one failure to `14 passed`. The pre-existing `sys.path` use of those constants was unaffected either way, which is why this only surfaced now that a subprocess consumes them.

## Two notes

**A nonzero `SystemExit` no longer reaches `torchrun_error.log`.** Only `Exception` is logged now, and `SystemExit` is not one, which the new parametrised test pins deliberately. Running `sys.exit(2)` under both branches: base writes the log, head does not. Correct for exit-code fidelity, but a user-visible record disappears, and the file is described as the compatibility bridge until launcher-owned stderr persistence lands. Worth being explicit about whether that record is meant to survive until then.

**The masking guarantee covers `Exception`, not `BaseException`.** A `KeyboardInterrupt` raised inside `stop_runtime`, which is a real double-Ctrl-C shape, escapes the `except Exception` and replaces the in-flight user exception from the `finally`. Base had the identical `finally`, so this is not a regression and I am not asking for it here. Flagging only because the description states the guarantee without that edge.

For sequencing: #416 builds on this branch and carried the same test path, so it picks the fix up on its next rebase.

PR #416, opened 2026-08-30
Dev/abhinav/training error

abhinavsriva:
Summary
Make the supervised training process the authoritative source of the TraceML command result.

Why
After training started, TraceML could replace its exit code when:
- The aggregator failed during finalization.
- final_summary.json was missing.
- Manifest writing or cleanup failed.
This could incorrectly report successful training as failed or hide the actual training result.

What changed
- Added a small TrainingOutcome value for subprocess return-code handling.
- Preserved the training process exit code through launcher finalization.
- Converted directly observed POSIX signals to conventional CLI exit codes.
- Changed aggregator, summary, manifest, and cleanup failures into warnings.
- Added a final terminal line reporting the training outcome.
- Applied the same result handling to --disable-traceml.

Testing
- Verified successful and failed training exit codes.
- Verified aggregator and finalization failures do not replace the result.
- Verified directly observed signal handling.
- Verified the final outcome remains the last terminal line.

Scope
This PR does not change error-log files or stdout/stderr persistence.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/cli.py
+++ b/src/traceml_ai/aggregator/display_drivers/cli.py
@@ -86,7 +86,10 @@ class CLIDisplayDriver(BaseDisplayDriver):
         # Watch profile
         self._renderers: List[CLIRenderer] = [
             SystemRenderer(db_path=self._settings.db_path),
-            ProcessRenderer(db_path=self._settings.db_path),
+            ProcessRenderer(
+                db_path=self._settings.db_path,
+                sampler_interval_s=self._settings.sampler_interval_sec,
+            ),
             StdoutStderrRenderer(db_path=self._settings.db_path),
         ]
 
--- a/src/traceml_ai/aggregator/display_drivers/nicegui.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui.py
@@ -195,7 +195,10 @@ class NiceGUIDisplayDriver(BaseDisplayDriver):
         self._renderers: List[DashboardRenderer] = [
             ContextRenderer(db_path=self._settings.db_path),
             SystemRenderer(db_path=self._settings.db_path),
-            ProcessRenderer(db_path=self._settings.db_path),
+            ProcessRenderer(
+                db_path=self._settings.db_path,
+                sampler_interval_s=self._settings.sampler_interval_sec,
+            ),
             StepMemoryRenderer(db_path=self._settings.db_path),
         ]
 
--- a/src/traceml_ai/launcher/commands.py
+++ b/src/traceml_ai/launcher/commands.py
@@ -175,7 +175,16 @@ def _run_noncritical_launcher_step(
         )
 
 
-def _exit_with_training_outcome(outcome: TrainingOutcome) -> None:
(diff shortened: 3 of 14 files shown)


Pendu:
Reviewing this as a draft, so this is shape feedback rather than a merge verdict. The increment is sound. Two structural things are worth settling before it leaves draft, and one behaviour change deserves a decision.

## How this reads on GitHub right now

The branch is built on `dev/abhinav/native_failures` (#415) but targets `version_0.3.7`, so the diff shown here is +355/-126 and includes #415's executor work. Reviewed as its own increment (`d2745f5..62ac655`) it is 201 insertions across `launcher/commands.py`, `launcher/process.py` and `tests/runtime/test_launcher.py`, which is the part that is actually new.

Retargeting the base to `dev/abhinav/native_failures` would make the PR show only that, and it would stop the executor changes from being reviewed twice. Otherwise #415 needs to land first.

Related: the test-path fix I pushed to #415 (`ba7886a`) also covers the copy of `tests/runtime/test_executor.py` carried here, so a rebase clears the failing test rather than needing a second fix.

## Verified in the increment

- The signal arithmetic is right. `TrainingOutcome.cli_exit_code` (`launcher/process.py:58-61`) returns `128 - returncode` for a negative code, giving 130 for SIGINT, 137 for SIGKILL and 143 for SIGTERM, which is the conventional shell mapping. `signal_name` guards `ValueError` for unmapped numbers and is skipped on Windows, where a negative code is not a signal.
- `_run_noncritical_launcher_step` is a clean way to express the intent. Manifest updates, stderr capture finish, aggregator teardown and degraded-telemetry recording all become warnings, so none of them can replace a result the training process already produced.
- Full suite on this branch is `1473 passed` with the one executor-test failure that the #415 fix resolves.

## The missing-summary case now records nothing

This is the one I would settle before merge. When training succeeds but `final_summary.json` is absent, base wrote `update_run_manifest(manifest_path, status="failed")` and exited 1 (`commands.py:615-616` on the parent commit). Head prints a warning and falls through to the training outcome (`commands.py:660-664`).

Not failing the run is right, and it is the stated point of the PR: training succeeded, and TraceML failing to write a summary is not the user's failure. But the manifest now carries `status="completed"` with no record that the summary is missing, so the degradation is visible only in terminal output that may already be gone.

Twenty lines below, the aggregator-exited-early path does exactly what seems wanted here, recording `telemetry_status: "degraded"` alongside the detail. Reusing that shape for the missing-summary case would keep the run marked completed while leaving the artifact trail honest. The same reasoning applies to the aggregator nonzero-exit branch, which now warns without a manifest marker.

## Smaller

`final_status` is computed before the manifest lambda that closes over it, and every `_run_noncritical_launcher_step` call invokes its operation immediately, so there is no late-binding surprise in the closures inside the poll loop. Worth stating since that pattern is easy to get wrong when it is added later.

abhinavsriva:
@pendu  Valid concern. The missing-summary/aggregator failure should be recorded in the manifest, but structured telemetry_status and telemetry_reason are intentionally scoped to the next PR. This PR only makes the training result authoritative and emits warnings; the next PR will record telemetry degradation independently without changing the training exit code.

Issue #409, opened 2026-08-28
Presentation-layer cleanup: charting, formatting, theme


Pendu:
Part 6 of #403, behaviour-preserving.

**Scope**
- [ ] `nicegui_sections/charting.py`: axis, time-series, line, sparkline helpers moved from `theme.py`
- [ ] `nicegui_sections/formatting.py`: duration, number, memory, missing-value formatting
- [ ] `theme.py` reduced to colours, fonts, CSS, spacing, visual tokens
- [ ] Every moved helper keeps its signature; sections update imports only

**Done means.** `theme.py` contains no data transformation, time parsing, series alignment, SVG
generation, or chart mutation; suite green; pages render identically.

Pendu:
Landed in #436, squashed onto `version_0.3.7` as `1aba8b2`.

`theme.py` ends at 182 lines and three functions, all of which emit markup or register fonts. Chart construction moved to `charting.py`, the bytes-to-gigabytes helper to `formatting.py` as `gb_si`. 282 lines of builders nothing reaches were removed: the gauge chart the System block dropped months ago, a dual-line builder, a single-line builder, a `nice_ymax` helper and two private functions only they called.

Evidence for the behaviour-preserving claim, strongest first: the 22-cell scenario matrix renders identically, 0 of 66 text fields differ; the builders were frozen by characterization tests written against the untouched module before anything moved; and `theme.gb` vs its new home was probed across 14 inputs after the move narrowed its `except Exception`, which turned out to be a real delta at `10**400`.

One correction worth recording. A first pass also deleted `_area` on the grounds that nothing outside the module referenced it, which is the wrong question for a private helper: `span_line_options` calls it and is live. A transitive reachability walk replaced the grep.

Two divergences surfaced here and deliberately left, since resolving either would move a rendered number inside a file move: the two byte-to-gigabyte divisors (#434) and the two `format_gb_pair` implementations (#435, since fixed).

PR #417, opened 2026-08-30
Process dashboard presentation: four tiles, per-rank charts and rows

### LINKED ISSUE #407: Process dashboard presentation

Pendu:
Closes #407. Part 4 of #403.

Part 3 (#412) is merged, so this branch is rebased straight onto `version_0.3.7` and is no longer
stacked. Review it against the release branch:

```
git diff upstream/version_0.3.7...dev/abhijeet/process-presentation
```

10 files, +1915/-320.

## Review scope

The Process card. This is the first part of the series that changes what is on screen, so the
screenshots below are the reviewable artifact as much as the diff is. The other panes in them are
today's rendering and are rebuilt in later parts.

## What

The #401 UI, reading the typed payload part 3 built instead of a dictionary of loosely named keys.

| before | after | why |
|---|---|---|
| `CPU` (raw process %) | `cpu capacity` (share of the host) | 700% on an 8-core host is not comparable to 700% on a 64-core one. 87.5% of capacity is. |
| `RAM` | `rss`, named for what it is | the tile always held RSS; it now says so |
| `GPU MEM` (one tile) | `cuda allocated` **and** `cuda reserved` | one tile could not say which it held. #399 settled the wording; this is the split it was settling it for |
| `GPU IMBAL` (a tile) | reserved spread, in the rows header | a spread is context for the per-rank view, not a headline number |
| one shared chart | two per-rank charts | a single line across ranks hides which rank is drifting |
| no per-rank view | per-rank table, engine-triggered | when a job stalls, WHICH rank stopped is the whole question |

### One rule for the whole section (review round 2)

Every tile now summarises the last **60 seconds**, and says so in the section title. It was the
last 100 samples, which is a different span at every sampling rate: over three minutes at a two
second cadence, under one at half a second. A card cannot claim a window it does not have, and two
runs sampling differently could not be compared.

| tile | the rule | reads |
|---|---|---|
| cpu capacity | each rank's median, then the highest | `25% of host`, `highest median · R0` |
| rss | each rank's median, then the highest | `1.4 / 16.6 GB`, `highest median · R0` |
| cuda reserved | each rank's PEAK, then the least headroom | `14.0 / 15.6 GB`, `least headroom · R0` |
| cuda allocated | the SAME rank the reserved tile chose | `12.2 / 15.6 GB`, `least-headroom rank · R0` |

Three of those changed what is shown, and each fixes something.

**RSS was showing the newest sample.** The last rows of a finished run land after the process
released its memory, so the tile read a collapsed number beside a chart still drawn at the run's
real level. The median removes that.

**Reserved was selecting on the newest reading.** The tile asks which rank comes closest to
filling its card, which is a high-water mark, so it now selects on each rank's peak across the
window.

**Allocated described a different rank from reserved.** It was the cohort median while its
neighbour was the least-headroom rank, so the two GPU tiles described two devices. They now
describe one, and can be read together.

### A total line on both charts

The per-rank lines answer "which rank", and the total answers "how much of this host is the job
using". Neither can be read off the other.

Ranks do not share a sample clock, so the total is not an index-wise sum: each rank's last known
value is carried forward to every timestamp in the union and summed there, which is what a step
function does between samples. A rank that has not reported yet contributes nothing rather than a
zero, so the total does not dip while ranks are still starting. A single-rank run gets no total,
because a line identical to the only other line is noise.

The total is the sum, so it sits above every rank line. The axis is fitted to the lines that are
DRAWN rather than to the rank traces alone, and a test pins that: fitting to a subset is what
clipped the RSS chart earlier in this PR.

### Two charts, two different y ranges, on purpose

CPU capacity is a share, so its axis is anchored at zero: the distance from zero is the reading.
RSS is a level a few GB high that moves by tens of MB across a run. Zero-anchoring RSS puts that
entire movement inside one pixel, and the movement is the leak the chart exists to show. The two
charts share one time axis so a vertical read across the pair lands on the same moment.

### New modules, and what did not go in `theme.py`

`charting.py` holds axis ranges, the time-axis pinning, per-rank colours and sparklines.
`formatting.py` holds how one already-decided value is written down.

The split is not cosmetic. A palette change is a brand decision and an axis-range change is a
decision about what a metric's information IS, and those change for different reasons. Eight of
the ten helpers this card needs were added to `theme.py` on the closed #401, which is the
`theme.py` +199 you flagged there. None of them go back there. Migrating the chart builders
`theme.py` already holds is part 6 (#409), deliberately not this PR.

`theme.py` does gain one line, and it is a CSS rule: `.tml-gpus tr.tml-stale td`, mirroring the
`.tml-mark` rule directly above it. Colours and CSS are what that module is for, so the rule
belongs there; no logic follows it in. Why it was missing is in the bugs section below.

## Two deliberate exceptions to the stated non-goal

#407 says "no computation changes". This PR makes two, both because a thing the card must say is
a fact about the telemetry rather than a drawing decision, and the view is not allowed to decide
either one.

**The rows auto-open trigger.** The rows open themselves when the reserved spread is worth
looking at. That rule has an arming condition: ranks reach their first CUDA allocation seconds to
minutes apart, so an unarmed trigger reads an ordinary startup ramp as total imbalance and throws
the rows open on every healthy run's first ticks. It fires only once every reporting rank has
HELD an allocation across its window. The alternative was to compare the spread against a number
inside `process_section.py`, which would put a severity call in the one layer that may not make
one. It is `rows_open` on the payload and the view only obeys it; a test asserts the view holds
no threshold of its own, so a 99% spread with the engine silent leaves the rows shut.

**The per-rank allocated rollup.** The `cuda allocated` tile has to name a rank and read that
rank's bytes, which is selection and arithmetic. `_alloc_rollup` does it beside the reserved
rollup whose rank it follows. Why the tile needed a new source at all is in the next section.

## What rendering the matrix found

The 22 fixture cells were read, not just rendered, each card against the payload its own database
produces. That pass found more than the test suite did, and none of it could have failed a unit
test.

**The `cuda allocated` tile showed the wrong quantity.** Its label then read "median rank, live
tensors", but it was wired to the aggregated step history, whose value is the least-headroom
rank's instantaneous device usage. On the 4-rank DDP cell that read 2.8 GB against a true median
of 2.1 GB, a 33% overstatement, above rows that each said 2.1. On single-rank cells it was worse:
a confident number about 99x below the rank's actual allocated bytes. And after teardown the
history has no GPU snapshot at all, so on six cells the tile read "n/a" directly above rows
listing each rank's allocated bytes.

It now reads a per-rank `_alloc_rollup`, which after the review round follows the rank the
reserved tile chose. That is the second compute addition in this PR, for the same reason as the
first: selecting a rank and reading its bytes is arithmetic, and the view may not do arithmetic.

**A stale rank was never actually dimmed.** The row carried `class="tml-stale"` and the
stylesheet defined no rule for it, so the dead rank drew identically to a live one and the single
cue that says WHICH rank stopped was invisible. Nothing in the markup or the payload could catch
that, so the check now lives where it can: a test extracts every class the card emits and asserts
each has a rule in the stylesheet.

**A 105 second window announced itself as "last 2 min".** `format_span` switched to whole minutes
at ninety seconds, overstating the window by 14% in the label a reader uses to place the chart in
time. It now stays in seconds until a full two minutes.

## Five more defects, found the same way, fixed in #412

The same pass surfaced five bugs in part 3: both charts empty on every short run, the GPU column
blank on a 4-GPU run, a finished GPU run rendering as CPU-only, `has_data` answering from the
wrong collection, and the per-rank rollups reporting a `p95` nobody computed. They are
compute-layer defects, so they were fixed in #412 where the code lives rather than carried here.
This PR sits on top of them.

Worth naming the pattern rather than the five instances: every one of them read a value from the
newest committed step instead of the newest REPORTING one, and teardown always empties the
former. A single guard would likely be better than five fixes, but that is a design conversation
for the compute layer, not something to fold into a presentation PR.

## Scope

```
STRUCTURE: chart shaping -> display_drivers/nicegui_sections/charting.py (new, formatting-layer
peer of theme.py); value formatting -> nicegui_sections/formatting.py (new, same peer);
mirrored from theme.py, which is the existing helper module in that unit; boundaries crossed: none
DATA PATH: process_section <- PROCESS_LAYOUT <- ProcessRenderer <- ProcessDashboardComputer
           <- ProcessRepository <- process_samples
SCOPE: declared the Process card + its two new helper modules; the diff's only reach outside
nicegui_sections/ is two compute additions the card's own labels require, the rows trigger and
the per-rank allocated rollup, plus one CSS rule in theme.py; 9 files, all three named above with
reasons -> within, with three stated additions
```

## Verification

```
GATE: pytest tests/ -> 1523 passed, 2 skipped; 175 on the display layer, of which 27 cover this card and 29 the two new helper modules; isort, ruff and black clean at 79 under the repo's pinned pre-commit
TRANSITIONS: recent -> retained chart, and the label says which; rows shut -> opened by the engine -> closed by the reader and NOT reopened while the condition holds; GPU present -> absent; a rank fresh -> stale -> dimmed but kept
RANKS: 1, 2, 3 and 4 ranks, a rank that stopped, and a rank with no usable clock; each rank is its own line on both charts and its own row
TRIPWIRE: a test asserts the view compares nothing to a threshold, so moving the trigger back into the card fails the suite; the end-to-end test states both changed tile meanings as explicit numbers, so a silent revert to the old meanings fails; no rank colour may be red, which on this page reads as a verdict
TWINS: the recent window changed UNIT (a 100-sample count became a 60-second duration), so the search is the concept and its synonyms, not the renamed symbol: `RECENT_WINDOW_S` / `window_s` / `window_n` / `_window_span` / `span` -> 4 consuming sites: repository.fetch_recent_rank_window (the read), dashboard_compute._rank_snapshots (the caller), dashboard_compute._one_chart via mode_for (the recent-to-retained transition), and dashboard_models.RankChart.window_s (the label). The third was MISSED on the first pass: grepping only the retired symbol `RANK_WINDOW_N` found two sites, while `_window_span(payload.history)` computed the same idea under another name and still keyed the chart transition to a 100-step count, so the tiles measured seconds while the charts measured samples with the suite green. Abhinav caught it in 84e883c. `_window_span` is now deleted and one duration drives the read and the transition, asserted by test_chart_mode_depends_on_duration_not_sample_count
SCENARIOS: 22 fixture cells rendered and read, not just rendered, on the merged version_0.3.7: B ok (B_xf_1gpu_long, B_cnn_1gpu) · C ok (C_M1_bert, C_sysB_1of4) · D ok (D_S2_ddp4, D_F1_fsdp4, D_sysA_ddp4_3k) · D' ok (Dd_S2_rank3_dead) · E ok (E_xf_2x1_long, E_cnn_2x1_fsdp) · E' ok (Ed_2x1_rank1_dead) · W ok (W_W1_busy, W_W2_idle) · V ok (V_S2_compute_bound, V_S3_input_straggler, V_S4_straggler, V_S5_input_bound, V_S6_h2d_bound, V_S8_warmup, V_M1_single_compute, V_M2_two_rank, V_F1_residual_fsdp); A, A2 and Wc are live-CPU recipes run by hand; F has no fixture
```

## Not in this PR

Part 5 (#408) migrates System onto the shared run-series and freshness infrastructure. Part 6
(#409) moves the chart builders `theme.py` still holds into `charting.py` and the value
formatting into `formatting.py`, leaving `theme.py` as colours, fonts and CSS.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/charting.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/charting.py
@@ -1,0 +1,200 @@
+# Copyright 2026 OptAI UG (haftungsbeschraenkt)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shaping a chart: axes, ranges, per-rank series, sparklines.
+
+``theme`` owns the look (palette, fonts, the base ECharts option dicts).
+This module owns the arithmetic of fitting a chart to its data: what the
+y range should be for a given kind of signal, how a time axis is pinned and
+labelled, and how one series per rank is built.
+
+The split matters because the two change for different reasons. A palette
+change is a brand decision; an axis-range change is a decision about what a
+metric's information IS. Keeping them in one file meant every axis fix
+touched the module that defines the brand.
+
+No function here reads a payload or decides severity. They take numbers and
+return option fragments.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Optional, Sequence, Tuple
+
+# One colour per rank, shared by the charts and the rows' chips so a line
+# and a row are recognisably the same rank. Red is not among them: on this
+# page red reads as a verdict, and a rank identifier is not a verdict.
+RANK_COLORS: Tuple[str, ...] = (
+    "#f97316",
+    "#3b82f6",
+    "#0d9488",
+    "#a855f7",
+    "#0ea5e9",
+    "#eab308",
+    "#ec4899",
+    "#10b981",
+)
+
+
+def rank_color(rank: int) -> str:
+    """The colour that identifies one rank, stable across every surface."""
+    return RANK_COLORS[int(rank) % len(RANK_COLORS)]
+
+
+def capacity_axis_max(values: Sequence[Any]) -> float:
+    """Zero-anchored ceiling whose half is a whole percent (0 / 5 / 10).
+
+    Right for a share of a capacity: the distance from zero is the reading,
+    so zero must be on the axis.
+    """
+    numbers = [float(v) for v in values if v is not None]
+    peak = (max(numbers) * 1.2) if numbers else 0.0
+    for top in (4.0, 10.0, 20.0, 30.0, 40.0, 60.0, 80.0, 100.0):
+        if peak <= top:
+            return top
+    return 100.0
+
+
+def drift_axis_bounds(
+    values: Sequence[Any],
+) -> Tuple[float, float, float]:
+    """A y range fitted to the data, for a series whose signal is DRIFT.
+
+    RSS is a level a few GB high that moves by tens of MB across a run.
+    Zero-anchoring it, which is right for a share of capacity, puts that
+    whole movement inside one pixel: on the three-hour capture the ranks
+    sit at 1.48 to 1.50 GB on an axis that would run to 5. The leak this
+    chart exists to show would be invisible.
+    """
+    numbers = [float(value) for value in values if value is not None]
+    if not numbers:
+        return (0.0, 1.0, 0.5)
+    low, high = min(numbers), max(numbers)
+    if high <= low:
+        high = low + max(abs(low) * 0.01, 0.01)
+    pad = (high - low) * 0.25
+    low, high = max(0.0, low - pad), high + pad
+    return (low, high, (high - low) / 2.0)
+
+
+def value_axis_formatter(span: float, unit: str) -> str:
+    """Tick formatter whose precision comes from the axis RANGE.
+
+    Magnitude alone is not enough: an axis fitted to a 20 MB drift around
+    1.4 GB would label every tick "1.4 GB" and say nothing.
+    """
+    if span >= 20:
+        decimals = 0
+    elif span >= 2:
+        decimals = 1
+    elif span >= 0.2:
+        decimals = 2
+    else:
+        decimals = 3
+    return f"v=>v.toFixed({decimals})+'{unit}'"
+
+
+def apply_span_axis(
+    options: Dict[str, Any],
+    span: float,
+    newest_epoch: Optional[float] = None,
+) -> None:
+    """Pin a chart to its span and label it in wall-clock time.
+
+    The x values are seconds before the newest sample, which keeps the
+    series arithmetic simple, but a reader debugging a slowdown needs the
+    clock: it is what their logs are keyed on. The formatters convert on
+    the fly from the newest sample's epoch, and the hover label carries
+    both readings ("19:10 · 45 min ago") so the axis and the tooltip never
+    speak two different vocabularies.
+    """
+    span = max(float(span), 1.0)
+    axis = options["xAxis"]
+    axis["min"] = -span
+    axis["max"] = 0
+    axis["interval"] = span / 3.0
+    if newest_epoch is None:
+        axis["axisLabel"]["show"] = False
+        return
+    axis["axisLabel"]["show"] = True
+    clock = (
+        "const d=new Date((%f+%%s)*1000);const q=n=>('0'+n).slice(-2);"
+        "const c=q(d.getHours())+':'+q(d.getMinutes())%s;"
+        % (float(newest_epoch), "+':'+q(d.getSeconds())" if span < 600 else "")
+    )
+    axis["axisLabel"][":formatter"] = "v=>{%s return c;}" % (clock % "v")
+    pointer = options.get("tooltip", {}).get("axisPointer", {}).get("label")
+    if pointer is not None:
+        pointer[":formatter"] = (
+            "p=>{%s const s=Math.round(-p.value);"
+            "return c+(s<1?' · now':(s<120?' · '+s+' s ago':"
+            "' · '+Math.floor(s/60)+' min ago'));}" % (clock % "p.value")
+        )
+
+
+def sparkline_svg(
+    values: Sequence[Optional[float]],
+    color: str,
+    *,
+    width: int = 64,
+    height: int = 14,
+) -> str:
+    """Inline SVG polyline; gaps are dropped, an empty trace is no SVG."""
+    points = [(i, float(v)) for i, v in enumerate(values) if v is not None]
+    if not points:
+        return ""
+    low = min(value for _i, value in points)
+    high = max(value for _i, value in points)
+    span = (high - low) or 1.0
+    step = width / max(len(values) - 1, 1)
+    inner = height - 4
+    coords = " ".join(
+        f"{i * step:.1f},{1 + inner - (value - low) / span * inner:.1f}"
+        for i, value in points
+    )
+    return (
+        f'<svg viewBox="0 0 {width} {height}" '
+        f'style="width:{width}px;height:{height}px;vertical-align:middle">'
+        f'<polyline points="{coords}" fill="none" stroke="{color}" '
+        'stroke-width="1.4"/></svg>'
+    )
+
+
+def shared_span(*traces: Sequence[Any]) -> Optional[Tuple[float, float]]:
+    """One anchor and span covering every trace given.
+
+    Both Process charts are pinned to it so a vertical read across the pair
+    lands on the same moment. Each trace is expected to expose a
+    ``timestamps`` sequence.
+    """
+    starts: List[float] = []
+    ends: List[float] = []
+    for group in traces:
+        for trace in group or ():
+            stamps = [
+                float(value)
+                for value in getattr(trace, "timestamps", ()) or ()
+                if value is not None
+            ]
+            if stamps:
+                starts.append(stamps[0])
+                ends.append(stamps[-1])
+    if not ends:
+        return None
+    newest = max(ends)
+    return (newest, max(newest - min(starts), 1.0))
+
+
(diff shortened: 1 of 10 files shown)


abhinavsriva (changes requested):
Plz look at the the review.

abhinavsriva:
Can we simplify this section around one consistent rule?

Section title: Process resources · recent 60s

Use samples from the most recent **60 seconds**. (this is better than last 100 sample, also consistent ?) :

- CPU: calculate the median for each rank, then show the rank with the highest median.
  - Tag: Highest median · R0
  - Value: 25% of host
  
- RSS: calculate the median for each rank, then show the rank with the highest median.
  - Tag: Highest median · R0
  - Value: 1.4 / 16.6 GB
  
- CUDA reserved: find each rank’s peak reserved memory, then select the rank with the least headroom.
  - Tag: Least headroom · R0
  - Value: 14.0 / 15.6 GB
  
- CUDA allocated: show allocated memory from the same rank and sample selected for CUDA reserved.
  - Tag: Least-headroom rank · R0
  - Value: 12.2 / 15.6 GB
  
For the graphs:
- CPU usage (cores): show Total plus one line per rank.
- RSS usage: show Total plus one line per rank.

This gives the section one clear contract: tiles summarize the most relevant rank over the recent 60-second (or X-second) window, while graphs show total usage and the per-rank breakdown. 

Please use a time window rather than the last 100 samples, because sample count represents different durations at different sampling rates.

I think similar last 100 samples logic is used in multiple places and we need to slowly change that to seconds where it make sense, at least in system and process, starting with process here in dashboard.

Pendu:
Pushed. All of it is in, and the window change turned out to be the right thread to pull.

**The window is a duration now.** `RECENT_WINDOW_S = 60.0`, and the read is a timestamp floor rather than `ROW_NUMBER <= 100`. Your reasoning holds up in the fixtures: at a 0.5 s cadence a hundred samples is 50 seconds and at 2 s it is over three minutes, so the old label was true of neither. A test builds a 0.5 s run and asserts the span comes back under 60 s where a sample count would have taken the lot.

**One rule per tile**, as specified:

| tile | rule | reads |
|---|---|---|
| cpu capacity | each rank's median, then the highest | `2.1% of host`, `highest median · R0` |
| rss | each rank's median, then the highest | `2.1 / 187 GB`, `highest median · R1` |
| cuda reserved | each rank's peak, then the least headroom | `1.2 / 14.6 GB`, `least headroom · R2` |
| cuda allocated | the same rank reserved chose | `0.80 / 14.6 GB`, `least-headroom rank · R2` |

Three of those changed what is shown rather than just its label, and each fixed something real:

- RSS was showing the newest sample. On a finished run the last rows land after the process releases its memory, so the tile read a collapsed number beside a chart still drawn at the run's real level. The median removes that.
- Reserved was selecting on the newest reading. The question is which rank came closest to filling its card, which is a high-water mark, so it selects on the window peak.
- Allocated was the cohort median while its neighbour was the least-headroom rank, so the two GPU tiles described two different devices. They describe one now.

**Both charts carry a Total.** Ranks do not share a sample clock, so it is not an index-wise sum: each rank's last known value is carried forward to every timestamp in the union and summed there. A rank that has not reported yet contributes nothing rather than zero, so the total does not dip while ranks are starting, and a single-rank run gets no total because a line identical to the only other line is noise.

Two things worth flagging.

**Adding the Total nearly reintroduced a bug this PR had already fixed.** It is the sum, so it sits above every rank line, and the y axis was being fitted to the per-rank traces only. It would have been clipped, which is exactly what happened earlier here when the axis was fitted to the rolling peaks instead of the drawn values. The axis now fits the lines that are drawn and a test fails if anyone refits it to a subset.

**Moving the tile to a median made the card contradict itself, so the rows moved too.** The tile said `2.1` for R1 while R1's own row said `1.0`, because the rows were still showing each rank's newest sample. A reader who goes to the rows to check the tile has to find the same number there, so the RSS column now shows the same basis. Caught by rendering the card and reading it, not by the suite.

On that last point, and it is the one I would most like your read on: none of these rule changes failed a single existing test. Nothing pinned the old behaviour, so nothing objected to the new. Six tests now discriminate: a rank whose peak and newest disagree, a teardown sample, a median rank that is not the least-headroom rank, a 0.5 s cadence, and two for the total.

`pytest tests/` gives 1523 passed, 2 skipped. The 22 fixture cells were re-rendered and read.

One I left alone deliberately. The section title says `recent 60s` while a long run's chart says `last 22 min · rolling 30 s`, because the title states the tiles' window and each chart states the span it actually holds. On any run past a couple of minutes the chart switches to the whole-run view by design, so the two diverge permanently rather than briefly. Nothing is false and each chart is self-describing, but if you would rather the window sat on the tile row instead of the section title, say so and I will move it.

abhinavsriva:
@Pendu Added 2 changes: CUDA reserved and allocated should come from same row so included a new dataclass for it and then used same reserved rank row for allocated and charintg was still using 100 samples changed that to use 60 sec logic.

Issue #419, opened 2026-08-30
System: consume the shared run-series and freshness modules (part 5b)


Pendu:
Part 5b of #403, and the second half of the original #408. Depends on 5a (#408).

System consumes `renderers/shared/run_series.py` and `freshness.py`, and the duplicated
rolling-window and point-budget code is deleted. Unlike 5a this DOES change behaviour, in the
places listed below. Each is stated here so the review is about whether the change is right,
rather than about whether it was noticed.

One framing correction worth making up front: adopting `FreshnessPolicy` is **not** a
de-duplication. System has no counterpart to it in any form; it never compares `sample_ts_s`
against a clock, so it currently cannot say that a node has stopped reporting. That is a new
capability, and if it is unwanted here it should be dropped from this PR rather than smuggled in
as cleanup.

**What changes**

1. **Rolling frame: ROWS becomes RANGE.** `common.py:276-279` hardcodes
   `ROWS BETWEEN n PRECEDING`; the shared plan emits `RANGE BETWEEN window_s PRECEDING` on
   SQLite 3.28+. They agree only when sampling is perfectly regular.

   Measured on a 200-sample run at a 2 s cadence with one 60 s gap, 30 s window, 12 preceding
   rows: at the first sample after the gap the row frame reports a rolling mean of **83.85**
   where the time frame reports **10.00**. The row frame reaches twelve rows back across the
   hole and averages in pre-gap values; the time frame sees only what is actually within 30 s.

   Adopted deliberately. A row frame silently spans more wall clock than it claims whenever
   samples are missing, which is precisely when someone is reading the chart to find out what
   happened. Note the max over that run is 90.00 either way, and on a regular fixture the whole
   difference is 41.6833 against 41.4188, which is why no existing assertion catches it.
2. **Non-finite rejection.** `finite()` rejects NaN and the infinities at the boundary. Today an
   `inf` power limit is accepted as a real board limit.
3. **Retained-mode hysteresis.** The `min_span_s` threshold becomes `mode_for` with its 1.2
   factor. Numerically identical except when the recent window spans 0 s, where the two disagree.
4. **NaN window span.** `choose_window_s` returns 300 s on a NaN span where `window_for` returns
   30 s.
5. **The mode is stated.** The payload carries `SeriesMode` instead of the view inferring it from
   which series came back empty. Today the CPU chart infers with `len(t) > 2` and the power chart
   with a non-empty test, so the two charts on one card can disagree about which view they show.

[...]

Issue #418, opened 2026-08-30
System renderer crashes on a NULL sample timestamp


Pendu:
`renderers/system/common.py` raises an uncaught `TypeError` when a system sample carries a NULL
`sample_ts_s`.

`sample_ts_s` is nullable (`aggregator/sqlite_writers/system.py:142`) and the whole-run readers do
not filter it. In `fetch_cpu_run_history` the `try/except sqlite3.Error` ends at `common.py:290`,
but the `float(r[0])` list comprehension that consumes the rows is at `:292`, outside it. A NULL
timestamp therefore escapes as a `TypeError` rather than degrading to the empty series the
surrounding code expects.

The GPU path has the same shape: `float(ts)` at `:377` sits outside the `try` that ends at `:364`,
and `reported_sql` (`:323-329`) does not require `sample_ts_s IS NOT NULL`.

Found while surveying the block for #408. It is present in `version_0.3.7` today and is
independent of that migration, so it is filed on its own rather than folded into a refactor.

**Fix sketch:** require `sample_ts_s IS NOT NULL` in both readers' WHERE clauses, or move the
row-consuming comprehension inside the guarded block. A test with one NULL-timestamp row pins it.

Pendu:
Shipped in #428, merged as `762aea1` on `version_0.3.7`.

The fix is one predicate, `sample_ts_s IS NOT NULL`, applied to all five reads that treat a row as
a sample, rather than moving the crashing comprehension inside the `try`.

Worth recording that this issue described the smaller half. The reported `TypeError` needs more
unclocked rows than the read's dropped prefix, which took 30 in reproduction and is invisible below
that threshold. The common case is one row and never crashed: the compute layer read
`float(ts or 0.0)`, so a missing clock became epoch 1970, the recent window appeared to span 54
years, the run could never outgrow it, and the whole-run chart silently disappeared with the card
falling back to the short view.

That half was found by searching for the concept rather than the reported crash, since the site
that causes it does not raise.

Closing by hand: the `Closes` keyword does not fire for PRs merged into a release branch rather
than the default branch.

PR #422, opened 2026-08-31
System dashboard: typed payload, and the card's rules move to compute

### LINKED ISSUE #420: System dashboard: typed payload and view interpretation (part 5a-ii)
Part 5a-ii of #403, and the second half of the structural System work. Depends on #408 (5a-i,
the repository extraction). Part 5b (#419) consumes `renderers/shared` and comes after both.

Split out of the original 5a because that scope was heading for roughly 2500 to 3000 lines
against an 888-line view. For comparison, #410 is the same work for Process and landed at
about 2100 lines against a 154-line view. #408 is now the pure move; this issue is the redesign.

**Scope**
- [ ] Add `renderers/system/dashboard_models.py`: a typed payload where absence is a typed `None`
      rather than a missing dict key, and the series mode is stated rather than inferred from
      which field came back empty
- [ ] `system_section.py` reads the typed payload instead of nested dicts
- [ ] Move metric interpretation out of the view into the compute layer

**The interpretation to move**, from a survey of all 888 lines:
- `odd_ones_out` (`system_section.py:198-220`): threshold derivation plus entity selection, which
  drives a visible mark on GPU rows
- `disclosure_text` (`:107-108`): a second utilisation-spread definition that can disagree with
  the `gpu_delta.p95` governing auto-open
- `gpus_unreported` (extracted in 5a-i): the card's absence rule
- the whole-run versus window decision (`:828-829`), re-made in the view with two DIFFERENT rules,
  `len(cpu_run_t) > 2` for CPU and a non-empty test for power, on top of the computer's own gate

**Two rules to decide, not merely relocate.** Both are pinned by characterization tests from
5a-i, so moving them without deciding will fail the suite rather than pass quietly.
1. `odd_ones_out` ties to the higher group, so on a two-GPU host the busier card is always the
   highlighted one, even when the card that fell behind is the interesting one.
2. The card and the computer disagree about "this GPU never reported": the card asks
   `mem_total is None AND power is None`, the computer asks `mem_total OR power_limit_w` present.
   A GPU reporting one field and not the other is reported to one and unreported to the other.

**Known test constraints**, proven by patching and re-running:
- `tests/renderers/test_system_dashboard_gpu_payload.py:219-241` asserts the `cpu_run` dict by
  EQUALITY on two degraded paths, so adding any field to the payload breaks them. Those
  assertions have to be updated as a stated decision, not incidentally
- the GPU power whole-run entry's `min` key is load-bearing: the chart draws it as a floor line
  and folds it into the y-axis bounds

**Non-goals.** No consumption of `renderers/shared` (#419). No visual change.

Pendu:
Closes #420. Part 5a-ii of #403. Follows #421, which merged; this branch is rebased onto it.

The System payload becomes types, the card reads them, and the rules the card was deciding for
itself move to the compute layer.

No visual change, and that is proven rather than asserted: the 22-cell scenario matrix was
rendered on `version_0.3.7` at `5093cbb` and on this branch, and every cell's strip text, System
card text and panel list is identical. Zero differing fields across 66 comparisons.

## Review scope

The System block only. No screenshots are attached: the claim this PR makes about rendering is
that nothing changed, and a pair of identical pictures is weaker evidence than the cell-by-cell
text diff in the SCENARIOS line below.

## What moves, and why each one is interpretation rather than formatting

| rule | was | now |
|---|---|---|
| the outlier split | `odd_ones_out` in the view, deriving a midpoint threshold and selecting GPUs by it | `_odd_gpus` in compute, `SystemRollups.odd_gpus` |
| the utilisation spread shown on the disclosure | an inline expression in the view | `_util_range` in compute, `SystemRollups.util_range` |
| whether the spread has earned opening the rows | `SPREAD_EXPAND_PTS` in the view | the same constant in compute, `SystemRollups.rows_over` |
| whether a GPU ever reported | the view asked `mem_total is None and power is None` | the computer's `reported` flag, read through `SystemRollups.gpus_unreported` |
| whether a chart is in the whole-run view | the view, with two different rules for its two charts | `_has_whole_run` in compute, `cpu_run_whole` and `power_run_whole` |

A threshold compared against a measurement is a severity call, and selecting entities by a derived
threshold is a diagnosis. Neither belongs in a module whose job is to format.

## The two rules the issue asked to decide, not relocate

**A two-GPU host still highlights the busier card.** The split marks the smaller group and ties go
to the higher one, so with two GPUs the groups always tie and the faster card is always the marked
one, including when the card that fell behind is the interesting one. Preserved deliberately and
pinned by `test_a_two_gpu_host_marks_the_faster_one`. Changing it changes what a user sees on the
most common multi-GPU box, which is a decision worth making on its own evidence rather than inside
a refactor.

**The computer's definition of "never reported" wins.** The card's rule was stricter, so the two
disagreed about a GPU carrying a power limit and no memory total. The computer decides what a
metric means, so its answer is the one that survives and the card now reads the flag. The visible
consequence is that a device that spoke and had nothing to say no longer makes the card announce
that the whole GPU sample is missing.

## One disagreement the move made visible, and did not resolve

The disclosure line and the auto-open gate are two different definitions of "spread". The text
shows `util_range`, the range across per-GPU medians. The gate uses `gpu_delta.p95`, the 95th
percentile of per-tick max minus min. They can disagree, and cell `D_F1_fsdp4` in the matrix is a
live instance: the rows read `util 98 to 100%` and are open, because the per-tick spread crossed
20 points even though the medians are two apart.

That was true before this PR, hidden inside one inline expression and one gate the view could not
see together. It is now two named fields on the payload, which is the state a decision can be made
from. Deciding it belongs with #419.

## Three things found reviewing this branch against our own architecture notes

**The whole-run rule had a dormant second copy.** `CpuRunSeries.is_populated` restated
`len(t) > 2` on the payload and claimed in its docstring to be why the card stopped deciding. It
had no readers; the card reads `cpu_run_whole`. Two definitions of one question is the exact defect
this PR removes from the card, so it is removed from the payload too.

**`SystemSeries.gpu_power` was annotated as a flat float series.** It carries one entry per device,
each `{"gpu_idx", "values"}`. Nothing broke because nothing checked, which is the point: a type
that lies is worse than no type, and this one shipped inside the PR whose purpose is that the
contract stops the card guessing.

**`to_dict` had no consumer.** Eleven methods reproducing the old mapping key for key, called by
one test helper and nothing else. It was written as a migration bridge while the card still read
dicts, and the commit that migrated the card is what ended that. The test now reads the types, which also satisfies the constraint the issue flagged: the two degraded
paths that asserted the `cpu_run` mapping by equality now assert `CpuRunSeries()`, so adding a
field is a deliberate change to those tests rather than an accidental break.

`from_dict` stays. The compute layer still assembles mappings internally and adapting them in one
place is what keeps that private.

## What is deliberately still a mapping

`RunContext.system_node` and the two per-GPU series, `gpu_power` and `gpu_power_run`. Each is a
shape the card reads by key today. Typing them is a change to what the compute layer emits, not to
how the card reads it, so it belongs with #419 rather than being smuggled in here.

```
STRUCTURE: System payload contract -> renderers/system/dashboard_models.py (new, owner of the
compute-to-card types); mirrored from renderers/process/dashboard_models.py; boundaries crossed:
none
DATA PATH: system_section <- SYSTEM_LAYOUT <- SystemRenderer <- SystemDashboardComputer
           <- SystemRepository <- system_samples + system_gpu_samples
SCOPE: declared renderers/system plus system_section.py and their tests -> diff touches
dashboard_models.py, dashboard_compute.py, renderer.py, system_section.py and four test files ->
within
TWINS: searched the CONCEPT, not the symbol - spread / util_range / gpu_delta / rows_over /
disclosure / the literal 20.0 -> 3 view sites (system_section.py:118, :256, :796), all now reading
the payload; the constant and both spread definitions live in dashboard_compute.py; the Process
card's own spread family is a different card and is untouched
```

## Verification

```
GATE: pytest tests/ -> 1556 passed, 2 skipped, against the base's 1554 passed, 2 skipped at 5093cbb; System-scoped 112 against the base's 110; black, isort and ruff clean at 79 on every file in the diff
TRANSITIONS: the recent-to-whole-run switch is the only one this PR touches, and it moves from two rules in the view to one in compute; both charts now answer it identically, asserted by test_both_charts_answer_the_whole_run_question_the_same_way and end to end by test_a_short_run_puts_neither_chart_in_the_whole_run_view
RANKS: unchanged; node scoping and the per-GPU rows read the same rows for the same node as the base, and the two-node cells E and E' render identically
TRIPWIRE: the card holds no threshold and derives no entity selection; odd_gpus, util_range, rows_over, reported and the whole-run flags all arrive decided, and the characterization suite fails if any of them moves back
SCENARIOS: A live recipe (run by hand) · A2 live recipe (run by hand) · B ok (B_xf_1gpu_long, B_cnn_1gpu) · C ok (C_M1_bert, C_sysB_1of4) · D ok (D_S2_ddp4, D_F1_fsdp4, D_sysA_ddp4_3k) · D' ok (Dd_S2_rank3_dead) · E ok (E_xf_2x1_long, E_cnn_2x1_fsdp) · E' ok (Ed_2x1_rank1_dead) · F no fixture · W ok (W_W1_busy, W_W2_idle) · Wc live recipe (run by hand) · V ok (V_S3_input_straggler, V_S4_straggler, V_S2_compute_bound, V_F1_residual_fsdp, V_M2_two_rank, V_S5_input_bound, V_S6_h2d_bound, V_M1_single_compute, V_S8_warmup); every cell diffed against the same render on 5093cbb, 0 of 66 text fields differ
```

## Not in this PR

#419 consumes the shared run-series and freshness modules, types the two remaining mappings, and
is where the spread disagreement above gets decided. #418 is a NULL-timestamp crash present in
`version_0.3.7` today and independent of this work.

### DIFF
--- a/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
+++ b/src/traceml_ai/aggregator/display_drivers/nicegui_sections/system_section.py
@@ -19,9 +19,17 @@ from __future__ import annotations
 
 import math
 from datetime import datetime
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 from nicegui import ui
+
+from traceml_ai.renderers.system.dashboard_models import (
+    GpuRow,
+    RunContext,
+    SystemDashboardPayload,
+    SystemRollups,
+    SystemSeries,
+)
 
 from . import theme
 
@@ -86,7 +94,7 @@ def format_gb_pair(used_bytes: Any, total_bytes: Any) -> Tuple[str, str]:
 
 def disclosure_text(
     gpus: List[Dict[str, Any]],
-    roll: Dict[str, Any],
+    roll: SystemRollups,
     *,
     is_open: bool,
 ) -> str:
@@ -107,21 +115,21 @@ def disclosure_text(
     tail = " · click to close" if is_open else " · click to open"
     if n == 1:
         return "1 GPU" + tail
-    span = roll.get("util_range")
+    span = roll.util_range
     if not span:
         return f"{n} GPUs" + tail
     low, high = span
     return f"{n} GPUs · util {low:.0f} to {high:.0f}%" + tail
 
 
-def node_scope_text(ctx: Dict[str, Any]) -> str:
+def node_scope_text(ctx: Optional[RunContext]) -> str:
     """'node 0 of 2' when the payload dropped other machines, else ''.
 
     Reads only the System payload's own ``system_node`` (the hosts seen in
     this window); the strip's node count lives on the CONTEXT payload and
     is not this block's to read.
     """
-    node = ctx.get("system_node") if isinstance(ctx, dict) else None
+    node = ctx.system_node if ctx is not None else None
     if not isinstance(node, dict):
         return ""
     in_window = int(node.get("nodes_in_window") or 1)
@@ -201,7 +209,7 @@ def sparkline_svg(
     )
 
 
-def gpus_unreported(gpus: List[Dict[str, Any]]) -> bool:
+def gpus_unreported(gpus: Sequence[GpuRow]) -> bool:
     """Whether every GPU present sent nothing but the zero fallback.
 
     The sampler emits an all-zero row when it cannot read a device, which
@@ -215,10 +223,10 @@ def gpus_unreported(gpus: List[Dict[str, Any]]) -> bool:
     carried a power limit and no memory total. The computer decides what a
     metric means; this asks it.
     """
-    return bool(gpus) and not any(gpu.get("reported") for gpu in gpus)
+    return bool(gpus) and not any(gpu.reported for gpu in gpus)
 
 
-def odd_ones_out(roll: Dict[str, Any]) -> set:
+def odd_ones_out(roll: SystemRollups) -> set:
     """GPU indices the computer marked as the smaller utilisation group.
 
     Read, not derived. The rule splits at the midpoint between the lowest
@@ -227,7 +235,7 @@ def odd_ones_out(roll: Dict[str, Any]) -> set:
     diagnosis rule, so it moved to the compute layer unchanged; this reads
     the answer.
     """
-    return {int(i) for i in (roll.get("odd_gpus") or ())}
+    return {int(i) for i in roll.odd_gpus}
 
 
 # Host CPU is psutil.cpu_percent(): the mean across logical cores, where 100%
@@ -245,9 +253,9 @@ def _num(value: Any, fmt: str = "{:.0f}") -> str:
 
 
 def rows_html(
-    gpus: List[Dict[str, Any]],
+    gpus: Sequence[GpuRow],
     power_series: List[Dict[str, Any]],
-    roll: Dict[str, Any],
+    roll: SystemRollups,
 ) -> str:
     """Build GPU rows and highlight the smaller side of a wide util split.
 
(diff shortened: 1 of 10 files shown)


abhinavsriva (approved):
Some minor typing issue

PR #423, opened 2026-08-31
Adds telemetry startup policy and health reporting without changing training outcomes.

abhinavsriva:
## Summary

Makes TraceML's existing launcher behavior when the aggregator is unavailable explicit and configurable.

Before this PR, `traceml run` and `traceml watch` already refused to start training when the aggregator was unavailable. This PR preserves that strict default while adding an explicit `warn` escape hatch for production environments where training should continue without telemetry.

The SDK behavior remains unchanged: `traceml.init()` warns and becomes a no-op by default when the aggregator is unavailable, while users can opt into strict behavior.

After training starts, the training process exit code remains authoritative. A later telemetry failure does not replace the training outcome.

## Behavior

| Frontend | Policy | Behavior when aggregator is unavailable |
|---|---|---|
| `traceml run` / `traceml watch` | Default: `raise` | Do not start training — unchanged from the existing launcher behavior |
| `traceml run` / `traceml watch` with `warn` | Explicit escape hatch | Continue through the supervised launcher with telemetry disabled |
| `traceml.init()` | Default: `warn` | Warn and continue as a no-op — unchanged |
| `traceml.init(on_missing_aggregator="raise")` | Explicit strict mode | Raise before instrumentation starts |

Policy precedence:

```text
explicit flag/argument > TRACEML_ON_MISSING_AGGREGATOR > frontend default
```

Examples:

```bash
# Strict launcher behavior: telemetry is required
traceml run train.py

# Best-effort telemetry: training remains authoritative
traceml run train.py --on-missing-aggregator=warn

# Deployment-wide best-effort policy
TRACEML_ON_MISSING_AGGREGATOR=warn traceml run train.py
```

```python
# SDK default: warn and continue
traceml.init()

# SDK strict mode
traceml.init(on_missing_aggregator="raise")
```

This policy is intentionally not added to `traceml.yaml`. It controls whether an expensive training job should start when TraceML is unavailable, so it remains an explicit launch or deployment decision.

## Changes

- Adds `--on-missing-aggregator {raise,warn}` to `run` and `watch`.
- Moves policy resolution into shared runtime settings.
- Preserves the launcher's existing strict default.
- Preserves the SDK's existing warn-by-default behavior.
- Stops partially started aggregators before applying the selected policy.
- Continues warn-mode training through the normal supervised executor with `TRACEML_DISABLED=1`.
- Adds separate manifest telemetry-health fields:
  - `telemetry_status`
  - `telemetry_reason`
  - `aggregator_exit_code`
- Ignores stale or malformed finalization warning/error evidence from reused run directories.
- Adds a concise telemetry footer before the final training footer.
- Keeps the training process exit code authoritative after training starts.
- Updates public documentation and focused tests.

## Design

Responsibilities remain separated:

- `runtime/settings.py`: shared policy resolution
- `launcher/manifest.py`: manifest fields and finalization evidence
- `launcher/commands.py`: launcher orchestration and final status reporting
- `sdk/initial.py`: SDK integration
- `launcher/cli.py`: CLI surface

No output capture, retry system, distributed manifest ownership, or logging framework changes are included. Those remain in later PRs.

## Verification

- All required GitHub checks: **passed**
- Focused launcher, SDK, and process tests: **98 passed**
- Real launcher/final-summary smoke test: **passed**
- Strict documentation build: **passed**
- Compile and diff checks: **passed**


Pendu:
Reviewed this closely because it touches how TraceML behaves when its own telemetry is unavailable, which is the one place our instrumentation can cost a user their training run. The design is sound and the escape hatch is the right addition. I pushed one fix, and there are four things worth your call.

## First, a correction to how the PR reads

The behaviour table presents `raise` as the new default for `run` and `watch`, which reads like this PR starts refusing to train. It does not. Base already refused, unconditionally and with no way out: `commands.py:604` on the parent commit did `raise SystemExit(1)` whenever the aggregator was not ready, for aggregator-owning and worker nodes alike.

So the actual change is that the existing refusal becomes explicit and, for the first time, overridable. Verified on a single node with an unbindable aggregator port:

```
--on-missing-aggregator warn    exit 0   TRAINING RAN
default (raise)                 exit 1   not started, message names the flag
```

That moves toward "never block training for instrumentation reasons", not away from it. Worth saying plainly in the description, because as written a reader could reasonably think the strictness is new.

## Verified

- Full suite reproduces your number exactly: `1569 passed, 2 skipped`.
- The partial-start cleanup holds. `commands.py:684-687` terminates the half-started aggregator and captures its exit code before any policy is applied, so the raise path does not leave the process behind.
- The CLI surface handles bad input well. `_resolve_cli_missing_aggregator_policy` converts a `ValueError` into a clean `SystemExit` with the message, so a mistyped flag gives an error rather than a traceback.
- Precedence works as documented: explicit argument, then `TRACEML_ON_MISSING_AGGREGATOR`, then the frontend default.

## Pushed (`ed12220`)

An exported-but-empty `TRACEML_ON_MISSING_AGGREGATOR` crashed `traceml.init()`, including through the documented HuggingFace entry point, even with a healthy aggregator.

The cause is a fallback lost in the refactor. Base read the value as `(...) or "warn"` (`sdk/initial.py:353` on the parent), so `None` and `""` both fell through to the default. The shared resolver checks `if resolved is None` (`runtime/settings.py:44`), so an empty string skips the default and reaches validation:

```
BASE   hf.init() OK
HEAD   hf.init() RAISED ValueError: on_missing_aggregator must be 'raise' or 'warn', got ''.
```

`export TRACEML_ON_MISSING_AGGREGATOR=` with no value is ordinary in CI and shell profiles, so this is reachable without anyone doing anything wrong. The fix treats blank and whitespace as unset, which restores the old behaviour and leaves a misspelled value still failing loudly. It comes with a parametrised test that fails four out of four cases without the change.

## Four for you to decide

**1. An invalid value still raises inside user training code.** The fix above covers blank, but `TRACEML_ON_MISSING_AGGREGATOR=warnn` still raises `ValueError` out of `traceml.init()`, and therefore out of `integrations/huggingface.py:48` and `integrations/lightning.py:121`. This is new: `sdk/initial.py:543` resolves the policy eagerly, before the connect attempt, where base resolved it at lines 601 and 620 inside the `except` handlers, so a bad value only mattered when the aggregator was already missing. A misconfigured environment variable now stops a healthy training run at import-time setup. The launcher already models the answer, converting the same `ValueError` into a clean exit; the SDK could warn and fall back to its documented `warn` default instead of raising. I did not push that because moving the resolution was a deliberate part of your refactor.

**2. Worker nodes never record a final telemetry status.** `commands.py:787` initialises `telemetry_status` and `telemetry_reason` to `None`, and both are only assigned inside `if owns_aggregator:` at `:789`. On any node that does not own the aggregator, the manifest therefore ends without a terminal telemetry state, so the new health fields describe rank 0 only. For a multi-node run that is the majority of the manifests.

**3. `summary_exists` does not use the staleness check you wrote.** `commands.py:806-808` decides it with a bare `(session_root / "final_summary.json").is_file()`. Meanwhile `_is_current_finalization_artifact` (`manifest.py:61-96`) does a real recency comparison against `aggregator_started_at`, and is called only from `read_current_finalization_reason` at `:105` and `:110`. So the artifact users care about most is the one exempt from the stale-artifact handling the PR describes.

I want to be accurate about the impact: I reran into a reused directory and the warn path still reported `telemetry_status: unavailable` correctly, because the earlier precedence rule catches it. The case that would bite is an aggregator that starts, is judged available, then never writes a summary, in a directory that already holds one from a previous run. Then the previous run's file satisfies `summary_exists` and suppresses `summary_missing`.

**4. A null aggregator exit code counts as healthy.** In `_derive_final_telemetry` (`commands.py:204-206`) the failure test is `aggregator_exit_code not in (0, None)`, so `None` is treated like a clean zero. `aggregator_exit_code` is read from `agg_proc.returncode` after `terminate_process_group`, which can leave it unset if the process was not reaped. An aggregator that had to be force-killed can therefore land in `complete` rather than `failed`. Whether `None` should mean "no evidence of failure" or "unknown, do not claim complete" is your call, but right now it silently means the former.

## Smaller

- The raise-path message lost diagnostics base had. Base printed the endpoint and the exit code: `aggregator was not reachable at {host}:{port} (exit={rc})`. Head prints neither (`commands.py:700-705`), so a user debugging a wrong `--aggregator-host` no longer sees which address failed. It also names `--on-missing-aggregator=warn` but not the environment variable, which is the form a cluster operator would more likely set.
- The success-path telemetry footer is new always-on stderr output and is not mentioned in the docs changes.

## Unrelated to this PR

`tests/runtime/test_torch_free.py::test_watch_runs_end_to_end_without_torch` failed twice for me while the machine was under heavy parallel load, then passed five out of five in isolation on both this branch and the parent. So it is load-sensitive rather than broken, and not something this PR introduced, but it will produce confusing red on a busy CI runner.

Pendu:
A second pass turned up one correction to what I wrote above, one item that moves from question to confirmed, and one I had missed.

## Correction: the hard-fail is not SDK-only

I said the CLI handles a bad policy value well because it converts the `ValueError` into a clean `SystemExit`. That is true but it understates the change, because base never read the variable at all. `TRACEML_ON_MISSING_AGGREGATOR` and `on_missing_aggregator` appear zero times in both `launcher/commands.py` and `launcher/cli.py` on the parent commit, so on base a bogus or empty value was simply ignored by the launcher.

At head, `validate_launch_args` and `launch_process` resolve the policy up front, so a stray value in a shell profile or CI environment now aborts `traceml run` before anything starts, on an otherwise perfectly healthy setup. The pushed fix covers the empty case; a misspelled one still stops both frontends. So point 1 in my earlier comment applies to the launcher as well as the SDK, and the "healthy run blocked by a configuration string" concern is the same on both.

## Now confirmed rather than a question: a force-killed aggregator reports "complete"

I raised the `not in (0, None)` treatment of `None` as a question. It reproduces:

```
after terminate_process_group: child.returncode = None
later poll:                    -9        (the process was killed and is dead)

_derive_final_telemetry(aggregator_exit_code=None, ...) -> ('complete', None, None)
```

`terminate_process_group` escalates to SIGKILL but does not wait afterwards (`launcher/process.py:262-281`), so `Popen.returncode` stays `None` for a process it just killed. `commands.py:784` reads that into `aggregator_exit_code`, and the precedence rule treats `None` as "no evidence of failure". An aggregator that hung in finalization and had to be killed therefore ends the run labelled `complete`. A `wait()` after the kill would give the real `-9` and route it to `failed`.

## Missed: the reason string reports the symptom, not the cause

In the default mode (summary with history enabled) an aggregator that dies mid-run guarantees the summary is absent, and `summary_required and not summary_exists` (`commands.py:208-209`) is tested before `aggregator_exited_early` (`:210-211`):

```
crashed aggregator + missing summary -> ('failed',   'summary_missing',          1)
same, but summary present            -> ('degraded', 'aggregator_exited_early',  1)
```

So the case where the cause is known and recorded mid-run is exactly the case where the final manifest overwrites it with the downstream symptom. The exit code survives, but `telemetry_reason` sends a reader looking for a missing file rather than a dead aggregator. Ordering `aggregator_exited_early` first, or carrying both, would keep the cause.

## Extends the worker-node point

Interrupted runs have the same shape as the non-owner nodes I mentioned. The signal handler writes only `update_run_manifest(manifest_path, status="interrupted")` (`launcher/process.py:332`) with no telemetry fields, so a Ctrl-C run ends with a terminal `status` beside whatever `telemetry_status` was last set, typically `running`. Same fix in both places: give every terminating path a terminal telemetry value.

None of this changes the overall read. The design is right and the escape hatch is the valuable part; these are the edges where the new field can state something the run does not support.

abhinavsriva:
Thanks @Pendu for the review. I updated the PR description, fixed aggregator shutdown handling, prevented old summary files from affecting new runs, improved telemetry status reporting for crashes and interrupted runs, clarified that node 0 reports the final telemetry status, restored useful error details, documented the final telemetry message, and fixed the Windows test.

PR #431, opened 2026-09-01
distributed log done

abhinavsriva:
## Summar

This PR ensures that multiple TraceML node launchers using the same shared run directory do not overwrite the same root manifests or legacy stderr artifact.

## Problem

In a multi-node run, every launcher could previously write:

- `manifest.json`
- `code_manifest.json`
- `crash_stderr.log`
- root manifest lifecycle and telemetry updates

Because the documented multi-node topology uses a shared filesystem, these writers could overwrite one another.

## Changes

- Made node rank `0` the sole writer of:
  - `manifest.json`
  - `code_manifest.json`
  - root manifest lifecycle updates
  - root manifest telemetry updates
- Non-owner nodes now operate without creating or updating a root manifest.
- Added a small helper for deterministic node-owned artifact directories:

```text
<run>/nodes/node_<node-rank>/
```

- Moved the existing optional stderr-tail artifact from:

```text
<run>/crash_stderr.log
```

to:

```text
<run>/nodes/node_<node-rank>/crash_stderr.log
```

- Updated artifact discovery to accept the node-scoped stderr path.
- Updated CLI help and user documentation.
- Added focused coverage for root ownership, non-owner behavior, node-path isolation, shutdown wiring, artifact discovery, and CLI help.

## Artifact ownership

| Artifact | Writer | Location |
|---|---|---|
| `manifest.json` | Node rank 0 only | Run root |
| `code_manifest.json` | Node rank 0 only | Run root |
| `crash_stderr.log` | Each node launcher | `nodes/node_<node-rank>/` |

## Design

The implementation deliberately reuses the existing launcher and manifest modules.

It introduces:

- one explicit root-writer decision;
- one node-path helper;
- optional manifest handling for non-owner nodes.

It does not introduce:

- locks or coordination services;
- per-node manifests;
- a new artifact package or abstraction layer;
- a distributed metadata protocol;
- new configuration options;
- output parsing or traceback classification.

Atomic root-manifest writes remain unchanged.

## Scope

This PR establishes artifact ownership only.

The following are intentionally handled by later RFC PRs:

- complete stdout/stderr draining — PR 5;
- canonical node-scoped training stdout/stderr files — PR 6;
- removal of the transitional `crash_stderr.log` artifact — PR 6;
- aggregator process stderr persistence — PR 7;
- launcher/internal logger ownership and legacy error-log cleanup — PR 8.

(Current is PR4 for reference) 

This PR does not change what training output is captured. It only makes the existing optional stderr-tail path safe for multi-node execution.

## Output model after the full RFC series

The planned canonical workload files remain node-scoped:

```text
<run>/nodes/node_<node-rank>/training.stdout.log
<run>/nodes/node_<node-rank>/training.stderr.log
```

Per-rank raw-output redirection, Ray-owned worker logs, remote upload, rotation, compression, and centralized log collection are outside this PR.

## Validation

- Full test suite: `1611 passed, 2 skipped`
- Focused launcher suite: `95 passed`
- Runtime suite: `222 passed`
- `git diff --check`: clean
- No unresolved conflict markers
- Branch contains exactly one PR 4 commit on top of merged PR 3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Opt-in `--capture-stderr` output now uses
+  `nodes/node_<node-rank>/crash_stderr.log`, preventing multi-node launchers
+  from overwriting a shared root file.
 - **Breaking:** `--summary-window-rows` and
   `TRACEML_SUMMARY_WINDOW_ROWS` were removed. Use
   `--history-retention`, `history_retention`, or
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
-- Opt-in `--capture-stderr` output now uses
-  `nodes/node_<node-rank>/crash_stderr.log`, preventing multi-node launchers
-  from overwriting a shared root file.
+- **Breaking:** Opt-in `--capture-stderr` output moved for both single-node and
+  multi-node runs from `logs/<run-name>/crash_stderr.log` to
+  `logs/<run-name>/nodes/node_<node-rank>/crash_stderr.log`, preventing
+  multi-node launchers from overwriting a shared root file.
 - **Breaking:** `--summary-window-rows` and
   `TRACEML_SUMMARY_WINDOW_ROWS` were removed. Use
   `--history-retention`, `history_retention`, or
--- a/docs/user_guide/faq.md
+++ b/docs/user_guide/faq.md
@@ -359,8 +359,9 @@ traceml run train.py --mode=summary --capture-stderr
 
 You can also set `TRACEML_CAPTURE_STDERR=1`. TraceML continues to print the
 child process's stderr to the terminal and stores only its last 64 KiB in
-`logs/<run-name>/crash_stderr.log` when the child exits. The file remains local
-and stderr capture is disabled by default.
+`logs/<run-name>/nodes/node_<node-rank>/crash_stderr.log` when the child exits.
+The node-scoped path prevents launchers from overwriting one another on shared
+filesystems. The file remains local and stderr capture is disabled by default.
 
 ---
 
--- a/src/traceml_ai/launcher/cli.py
+++ b/src/traceml_ai/launcher/cli.py
@@ -231,7 +231,8 @@ def _add_launch_args(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         help=(
             "Tee child stderr to the console and retain its last 64 KiB in "
-            "logs/<run-name>/crash_stderr.log. Default: disabled."
+            "logs/<run-name>/nodes/node_<node-rank>/crash_stderr.log. "
+            "Default: disabled."
         ),
     )
 
--- a/src/traceml_ai/launcher/commands.py
+++ b/src/traceml_ai/launcher/commands.py
@@ -28,6 +28,7 @@ from traceml_ai.launcher.launch_config import (
 from traceml_ai.launcher.manifest import (
     collect_existing_artifacts,
     is_current_summary_artifact,
+    node_artifact_dir,
     read_current_finalization_reason,
     update_run_manifest,
     utc_now_iso,
@@ -155,12 +156,12 @@ def _stderr_capture_enabled(
 
 
 def _finish_stderr_capture(
-    capture: Optional[StderrTailCapture], session_root: Path
+    capture: Optional[StderrTailCapture], output_path: Path
 ) -> None:
     """Persist an optional stderr tail without affecting the training result."""
     if capture is None:
         return
-    capture.finish(session_root / "crash_stderr.log")
+    capture.finish(output_path)
 
 
 def _run_noncritical_launcher_step(
@@ -527,6 +528,8 @@ def launch_process(script_path: str, args: argparse.Namespace) -> None:
     _require_dashboard_dependencies(str(cfg["mode"]))
 
     owns_aggregator = aggregator_cfg.is_owner(node_rank=torchrun_cfg.node_rank)
+    # Root run metadata has one writer even when every node has a launcher.
+    is_root_writer = torchrun_cfg.node_rank == 0
 
     env = os.environ.copy()
     env["PYTHONUNBUFFERED"] = "1"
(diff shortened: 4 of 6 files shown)


Pendu:
Went after the ownership claim hard, since a shared filesystem is the one place this can go wrong quietly. It holds up. Nothing blocking. One thing I would change before merge, and three worth a look.

## Verified

- Full suite `1611 passed, 2 skipped` at head against `1609 passed, 2 skipped` at the merge base, so exactly the two added tests and no regressions. Focused launcher suite reproduces `95 passed`.
- The ownership gating is complete on the launcher side. All seven root-manifest write sites are guarded (`commands.py:701,740,762,780,842,878` plus `process.py:338`), and `write_run_manifest` / `write_code_manifest` sit inside `if is_root_writer:` at `commands.py:592`. Tracing every file operation under the run tree on a simulated `node_rank=1` launcher, the only writes are the node's own directory and its `crash_stderr.log`. No manifest, no code manifest, no attempted-and-failed write.
- The bug being fixed is real, not theoretical. Running the same two-node sequence against the merge base leaves a single shared `crash_stderr.log` whose content after both nodes is only the second node's, with `manifest.json` rewritten underneath it.
- Ctrl-C on a worker node is safe. `manifest_path` is `None` there, and firing the real handler with `manifest_path=None` gives a clean `SystemExit(130)` with no manifest touched. On an owner node the same path still records `status=interrupted`.
- Concurrent node-directory creation is safe, which matters because the premise is a shared filesystem. `process.py:168` uses `parents=True, exist_ok=True`; 24 concurrent creations across two ranks produced zero errors.
- Old run directories are unaffected. `collect_existing_artifacts` takes an explicit `crash_stderr_path` and records it only when the file exists, so nothing rescans a run root. A pre-existing run keeps the path its own manifest recorded at write time.
- The helper's odd inputs are unreachable. `node_artifact_dir` would happily build `node_-1`, but `--node-rank` is parsed through `_non_negative_int` and rejected when `>= nnodes` (`launch_config.py:75-80`), so a negative or out-of-range rank never reaches it.

## Worth changing: the relocation is not marked breaking, and it is not multi-node only

`node_artifact_dir` has no single-node branch, so `node_rank` 0 sends the artifact to `nodes/node_0/crash_stderr.log` on an ordinary single-node run. Confirmed with a real `--capture-stderr` run: no root-level file is created, and the manifest records the node-scoped path.

The CHANGELOG entry describes the change as "preventing multi-node launchers from overwriting a shared root file", which is accurate about the motivation but reads as though single-node users are unaffected. They are not: anyone whose CI or tooling reads `logs/<run-name>/crash_stderr.log` silently gets nothing, with no error to explain it.

The repo's own convention already covers this. `CHANGELOG.md:13` and `:24` both use a `**Breaking:**` prefix for removed options and moved modules, and a relocated artifact path is the same class of change. Marking this entry the same way, and saying plainly that the single-node path moved too, would be enough. Keeping the root path when `nnodes == 1` is the alternative if you would rather not break it at all.

## Three for your judgement

**Worker-node stderr is written but indexed nowhere.** Each node writes its own `nodes/node_<rank>/crash_stderr.log`, but only node 0 has a manifest, and it records only its own path (`commands.py:848-852`). So on a multi-node run the artifact that exists precisely to diagnose a worker crash is the one no manifest points at. A user has to know the directory convention to find it. That may be the intended split for this PR, with PR 6 making node-scoped output canonical, but it is worth being deliberate about rather than arriving there by default.

**One root write is still ungated, and it is node-specific.** `samplers/system_manifest.py:204-215` writes `<run>/system_manifest.json` from the training process. It is not ungated exactly: the System sampler is registered `rank_zero_only`, so `sampler_registry.py:62` limits it to local rank 0, which means one writer per node rather than one per run. It is also guarded by `if manifest_path.exists(): return` and written atomically, so it cannot corrupt. But on an N-node job that is still N processes racing for one root file, the check-then-write is not atomic across nodes, and the payload is per-node: `hostname` at `:63`, `logical_core_count` at `:89`, `gpu.count` at `:133`. On a heterogeneous cluster whichever node wins the race defines the recorded hardware for the whole run. Outside this PR's launcher scope, but it is the same collision class this PR exists to close, so it belongs on the list for the series.

**There are now two predicates for one idea.** `is_root_writer = node_rank == 0` (`commands.py:532`) sits beside `owns_aggregator = aggregator_cfg.is_owner(node_rank)` (`:530`). They agree today only because `owner_node_rank` defaults to 0 and has no CLI or environment setter anywhere in `src/`. If it ever gains one they diverge silently, and `commands.py:617` computes `telemetry_status` from `owns_aggregator` inside the `is_root_writer` block, so node 0 would write a null while node N held the computed status with no manifest to put it in. Worth either deriving one from the other or asserting they match.

I did notice the lifecycle updates moved from `owns_aggregator` to `manifest_path is not None` (`commands.py:761,780`), which is the more correct binding of the two. That direction applied to the telemetry fields would close the gap above.

## Test coverage

The new tests cover the help text, the helper, node-scoped discovery, and the non-owner no-manifest path. Three things they do not reach: the real signal handler, which is mocked in every launcher test, so nothing exercises owner-interrupted versus non-owner-skip end to end; actual file placement on disk, since the assertions check the path handed to a mocked `finish()` rather than the file; and old-run backward compatibility, which I verified by hand and which stays correct only as long as nothing starts rescanning a run root.

Unrelated to this PR: `tests/runtime/test_final_summary_smoke.py` failed once for me with `EADDRINUSE` on port 29500 under parallel load and passed in isolation, so it is a master-port flake rather than anything you changed.

Pendu:
Two additions to the above, one of which I under-weighted the first time.

## The system manifest is the same defect, not an aside

I filed `system_manifest.json` as an out-of-scope observation. That was too soft. This PR fixes one defect class, several nodes racing to write one file in the run root, and it fixes three instances of it: `manifest.json`, `code_manifest.json` and `crash_stderr.log`. `system_manifest.json` is the fourth instance of that same class, left untreated.

It is gated to local rank 0 by the sampler registry, so it is one writer per node rather than per process, but on an N-node job that is still N processes racing for a single root path. The guard is `if manifest_path.exists(): return` followed by a separate atomic write, so the check and the write are not atomic with respect to each other across nodes. The payload is per-node (`hostname`, `logical_core_count`, `gpu.count`), so on a heterogeneous cluster whichever node wins defines the hardware recorded for the whole run, and nothing indicates which node that was.

Your own words on #400 are the standard I should have applied here: "two failure paths and this PR migrated one." Same shape. The fix does not have to land in this PR, but the sweep is worth doing as part of the class rather than discovering the fourth instance later from a confusing manifest.

## Two predicates now decide one thing

Worth treating as more than a latent risk. Root-artifact ownership is now decided in two independent places: `is_root_writer = torchrun_cfg.node_rank == 0` at `commands.py:532`, and `owns_aggregator = aggregator_cfg.is_owner(node_rank)` at `:530`, which reads `owner_node_rank`. One question, two answers, kept consistent only by the accident that `owner_node_rank` defaults to 0 and has no CLI or environment setter.

The failure mode if they ever separate is quiet rather than loud: `commands.py:617` computes `telemetry_status` from `owns_aggregator` but sits inside the `is_root_writer` block, so the node holding the manifest would write a null while the node that computed the status has nowhere to put it. Deriving one from the other, or asserting they agree at startup, costs a line and removes the question permanently.

## Credit where it is due

The transitional seam is handled exactly right. `crash_stderr.log` is named as transitional, and PR 6 is named as the change that deletes it, in the same body. Naming a temporary seam together with the PR that retires it is the thing that usually gets skipped and then never happens, so it is worth saying that this one does it properly.

PR #439, opened 2026-09-02
Dev/abhinav/train proc stdout err

abhinavsriva:
## Summary

Implements **PR 5 of RFC-0001** by adding a reusable, byte-oriented subprocess-output drainer. This prepares the launcher to persist complete training stdout and stderr without changing current launcher or CLI behavior yet.

## What changed

- Added `ProcessOutputDrainer` to the existing launcher process module.
- Drains stdout and stderr concurrently to prevent pipe deadlocks.
- Preserves exact raw bytes in separate output files.
- Supports optional terminal mirroring for later display-mode integration.
- Retains only the final 64 KiB of stderr for failure excerpts.
- Continues draining and falls back to the terminal if file persistence fails.
- Returns confirmed output paths and one consolidated warning without affecting the training result.
- Provides bounded, idempotent shutdown and closes owned streams.
- Added focused coverage for:
  - simultaneous stdout/stderr flooding;
  - invalid UTF-8 and partial final lines;
  - exact file and mirror output;
  - bounded stderr retention;
  - file-open and file-write failures;
  - mirror failure;
  - terminal fallback;
  - bounded and idempotent shutdown.

## Relationship to PR 4

PR 4 established distributed-safe artifact ownership and node-scoped directories.

This PR does not change that ownership model. It adds the reusable output primitive that will later write:

- `nodes/node_<node_rank>/training.stdout.log`
- `nodes/node_<node_rank>/training.stderr.log`

## Scope

This PR only introduces and tests the reusable process-output primitive. Existing launcher behavior remains unchanged, and no production caller uses the new drainer yet.

## Out of scope

- Starting training with stdout/stderr pipes.
- Adding `--save-training-output`.
- Changing CLI, summary, or dashboard presentation.
- Removing the existing `StreamCapture`, `StdoutStderrSampler`, or `StderrTailCapture`.
- Removing `crash_stderr.log`.
- Capturing aggregator stderr.
- Consolidating TraceML internal error logs.
- Per-rank raw-output routing, rotation, compression, or upload.

## Follow-up PRs

- **PR 6:** Integrate the drainer into the launcher, save node-scoped training output, keep Rich CLI output quiet, show a bounded failure excerpt, and disable the old in-process stdout/stderr capture path for new runs.
- **PR 7:** Reuse the drainer to persist raw aggregator stderr.
- **PR 8:** Consolidate TraceML internal error-log ownership.

## Validation

- Full test suite: `1631 passed, 2 skipped`
- `git diff --check`: clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Added a byte-oriented subprocess-output drainer for future launcher-owned
+  stdout/stderr persistence; existing launcher behavior remains unchanged.
 - Root `system_manifest.json` writes now come only from global rank 0, avoiding
   nondeterministic node ownership on shared filesystems.
 - **Breaking:** Opt-in `--capture-stderr` output moved for both single-node and
--- a/src/traceml_ai/launcher/process.py
+++ b/src/traceml_ai/launcher/process.py
@@ -29,6 +29,8 @@ _CREATE_NEW_PROCESS_GROUP = getattr(subprocess, "CREATE_NEW_PROCESS_GROUP", 0)
 DEFAULT_TCP_READY_TIMEOUT_SEC = 15.0
 DEFAULT_SHUTDOWN_TIMEOUT_SEC = 5.0
 DEFAULT_STDERR_TAIL_BYTES = 64 * 1024
+_OUTPUT_READ_BYTES = 64 * 1024
+_OUTPUT_STOP_GRACE_SEC = 0.1
 INTERRUPTED_EXIT_CODE = 130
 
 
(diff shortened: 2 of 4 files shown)


Pendu:
Spent the time on the concurrency and the byte claims, since a drainer that deadlocks or truncates is worse than no drainer. Both hold. Nothing blocking. Three minors, all of which land on PR 6 rather than here, plus one question.

## Verified

- Full suite `1631 passed, 2 skipped` at head against `1624 passed, 2 skipped` at the merge base, so exactly the seven added tests and no regressions.
- The deadlock claim holds under a real load. A child flooding both pipes with 2 MiB each finishes with `rc=0`, both files exactly 2097152 bytes, no hang and no warning. That is the failure this primitive exists to prevent, and it is prevented.
- Byte fidelity is exact. A payload containing invalid UTF-8, a NUL, CRLF, lone surrogate bytes and no trailing newline round-trips byte-for-byte to the file.
- The 64 KiB stderr bound is exact at every boundary I tried, and it bounds only the tail. At 65535, 65536, 65537 and 5 MiB the retained tail is precisely the last 65536 bytes while the file on disk keeps the complete stream.
- Degradation is real, not nominal. A file that cannot be opened, a write that fails mid-stream, and a failing mirror each keep the drain running, produce one consolidated warning rather than a per-line storm, and never raise into the caller.
- The empty case is clean: a drainer constructed with no streams at all returns `ProcessOutputResult(stdout_path=None, stderr_path=None, stderr_tail=b'', warning=None)`.
- The no-caller claim is accurate; `ProcessOutputDrainer` has no reference in `src/` outside its own definition.
- Capturing the stderr tail before the sink write is a good ordering choice: the failure excerpt survives even when the file write is what failed.

## Three minors, all pointed at PR 6

**`finish()` is not idempotent under concurrent callers.** The guard at `process.py:244-245` reads `self._result` outside the lock and the assignment at the end is also unlocked, so concurrent callers all pass the check and each builds its own result. Eight concurrent calls returned five distinct result objects; on the timeout path, six concurrent calls closed the same stream seven times, once per caller plus the drain worker. Sequential use is fine and returns the same object, and the returned content agreed in every trial I ran, so this is small today with no caller. It is worth two lines because the class already holds `self._lock` and the description says idempotent without qualification. A second pass I ran did once observe two callers disagreeing on whether the timeout warning was present, but I could not reproduce that in three further attempts, so treat it as unconfirmed rather than as a known divergence.

**The docstring promises more closing than the code does.** `process.py:84` says "The drainer closes supplied streams; callers must not read them afterward." Source streams and the drainer's own sink are closed, but `stdout_mirror`, `stderr_mirror`, `stdout_fallback` and `stderr_fallback` never are. Verified: a `BytesIO` passed as mirror is still readable after `finish()` returns. Not closing a caller-owned terminal stream is the right behaviour, so the code looks correct and the sentence looks wrong. It matters because PR 6 is the first caller and may hand in a file object as fallback while believing the drainer owns it.

**A failed stream can leave a partial file that nothing points at.** `_completed_paths` is populated only when the loop reaches EOF without a sink failure, and `finish` further excludes timed-out streams, so on a mid-stream error the result reports `stdout_path=None` while the bytes already written stay on disk. Reproduced with a source that raises after two reads: `result.stdout_path=None` with `b'good1good2'` in the file. The caller cannot tell "no file" from "partial file", which is exactly the case crash diagnostics care about. PR 6 needs a policy here, whether that is unlinking, reporting the path with a partial flag, or documenting it.

## One question

Three output-capture paths now coexist in the tree: this drainer, `StderrTailCapture` in the same module, and `StdoutStderrSampler`. The body defers consolidation to PRs 6 and 8 and names them, which is the right way to carry a temporary seam. The question is only whether the drainer's shape is settled enough to survive that merge, because it is 224 lines with no caller and the first real caller will be the one to find out. If PR 6 is close, landing them adjacently would shorten the window where the tree carries two answers to the same question.

## Not a finding, worth noting

The added `tests/samplers/test_system_manifest.py` pins global-rank-zero ownership of the root system manifest. That sits outside this PR's stated scope, which is the output primitive, but it pins real behaviour and I would keep it rather than split it out.

Issue #440, opened 2026-09-02
System cluster view averages an unreported GPU into the node mean


Pendu:
`SystemCLIClusterBuilder._node_sample_from_row` aggregates per-GPU telemetry across a node's
devices with no reported-ness test. Both sibling computers in the package have one:
`cli_compute.py:116` and `dashboard_compute.py:355` both skip a row that
`renderers/system/common.py:gpu_reported` says never reported.

The cluster builder is the third path that answers the same question, and it is the one that
still averages a failed device in.

## What it costs

The current sampler writes NULL when NVML cannot read a device, and `_optional_float` drops a NULL
correctly, so traces written today are unaffected. The gap is the older form, which
`common.py:54-56` documents: a placeholder row of all zeros for the same failure state.

On such a trace, a node whose devices are all placeholders contributes a real `0.0` to
`gpu_utils`, and `max(gpu_temps)` over `[0.0]` reports `0.0 °C` for that node. The node then wins
the "worst" comparison for utilisation, because `_rollup` is called with `higher_is_worse=False`
for `gpu_util` and zero is the lowest any node can report, and the cluster median moves with it.

That is the same defect #429 fixed on the dashboard and the single-node terminal card, alive in
the multi-node view.

## The fix

Import `gpu_reported` from `common.py` and skip an unreported row in the loop at
`cli_cluster.py:225`, matching the two siblings. The module now has three tests; a fourth covering
a placeholder-row node would pin it.

## Why it was not folded into the absence work

That change was about a metric nobody read, and this is about a device that never reported: a
different question, with a different trigger and a different guard. Fixing it here would also have
changed aggregation for four metric families on a path whose only reachable input is a legacy
trace, which is not the shape of a focused change.

giovannicozzolongo:
Hi, I can take this. Could you assign it to me?

Issue #429, opened 2026-09-01
System: an unreported GPU is averaged in as a real zero, so the util tile under-reads


Pendu:
An unreported GPU is averaged into the utilisation series as a real zero, so a fully busy host
reads as partly idle and the card opens itself claiming an imbalance that does not exist.

![what the card shows](https://traceopt-share-543486084785.s3.eu-central-1.amazonaws.com/pr-shots/gpu-series-unreported/unreported.png)

## Measured

Four GPUs, all pinned at 100%, over a 60-sample run. One of them stops reporting utilisation:

| | healthy | gpu1 `util` NULL | gpu1 NVML fallback row |
|---|---|---|---|
| GPU UTIL tile (mean) | 100.0 | **75.0** | **75.0** |
| spread driving auto-open (max - min) | 0.0 | **100.0** | **100.0** |
| rows auto-opened | no | **yes** | **yes** |
| disclosure text beside them | 100 to 100% | 100 to 100% | 100 to 100% |
| GPU TEMP tile (max) | 45 | 45 | 45 |
| payload's `reported` flag for gpu1 | - | true | **false** |

The last row is the sharp part. In the NVML case the computer already knows the device did not
report, the flag is on the payload and is correct, and the series does not consult it.

## Where

`renderers/system/dashboard_compute.py`, the recent-window series loop:

```python
utils = [float(g["util"] or 0.0) for g in rows]
...
gpu_avg[i] = sum(utils) / float(len(utils))
gpu_delta[i] = max(utils) - min(utils)
```

Two separate ways in, both real:

1. **`util` arrives NULL.** `or 0.0` turns a missing reading into a measured zero. The device is
   otherwise reporting, so `_gpu_reported` is true for it; this is a per-METRIC absence on a
   present device, which the per-device flag does not describe.
2. **NVML fails and the sampler writes its all-zero fallback row.** Here the value genuinely is
   `0.0`, so `or 0.0` never fires, and only the `reported` flag distinguishes it from an idle GPU.

## Why only two numbers are wrong

`gpu_avg` is a mean and `gpu_delta` is `max - min`, so a spurious zero moves both. `temp_max` and
`gpu_mem_worst` take a `max`, and a zero never wins a maximum, so those tiles are correct by
accident rather than by design. `util_range`, which produces the visible "100 to 100%" text, is
computed on the rollup path that already excludes unreported devices.

That mismatch is why the failure presents as a self-contradicting card: the rows open to show an
[...]

Issue #426, opened 2026-09-01
System: a failed read renders "no data" instead of the last good payload marked stale


Pendu:
Found while surveying the System block for part 5b (#424) and deliberately left out of it, because
the obvious half of the fix makes the card worse on its own. All three parts below have to land
together.

## The defect

`SystemDashboardComputer.compute()` has exactly one error boundary. It catches, calls
`_return_stale`, and serves the last good payload within its TTL with a status string saying so.
That is the right design.

`SystemRepository` then swallows underneath it. Four `except sqlite3.Error` blocks return `None`
or `[]` (`repository.py:181`, `:246`, `:299`, plus `:228` which is a different case, see below).
A database failure therefore arrives at the compute layer as a **valid-looking empty result**, the
boundary never runs, and the card confidently draws "no data" for a host that is fine.

This is the pattern already removed from the Process repository in `bde1ca6`, "error returns None
or [] so stale thing is shown", which also added `test_run_stats_sql_errors_propagate` and
`test_run_history_sql_errors_propagate`. `renderers/process/repository.py` now has zero error
handling. The two blocks disagree today.

![what the card shows in each case](https://traceopt-share-543486084785.s3.eu-central-1.amazonaws.com/pr-shots/system-5b-decisions/issue2_readfail.png)

## Why propagating alone is NOT the fix

`_return_stale` writes the status onto the payload (`dashboard_compute.py:724`,
`rollups=replace(cached.rollups, status=msg)`) and `SystemRollups.status` exists on the model.
**`system_section.py` never reads it**: `grep -c status` returns 0 across its 884 lines.

So propagating today would trade two empty charts for a frozen whole card with no stale marker
anywhere, and the reader would have no way to tell a held-over payload from a current one. That is
worse than the current behaviour, not better.

## Not all four blocks are the same thing

`repository.py:228`, inside `fetch_cpu_run`, carries the comment "Window functions need SQLite >=
3.25; without them the whole run view is simply unavailable and the window view stands." That is
deliberate graceful degradation, not a swallow. Deleting it with the others turns a missing chart
into a dead card on an old-SQLite host, which is a real target given Slurm and
`requires-python >= 3.10`.

`renderers/shared/run_series.py:59` already sets the precedent for the right form:
`HAS_RANGE_FRAME = sqlite3.sqlite_version_info >= (3, 28, 0)`, an explicit capability check rather
than a caught exception.

## The three parts, which must ship together

[...]

PR #441, opened 2026-09-02
System: an all-absent window is not a measurement of zero

### LINKED ISSUE #437: System: an all-absent window renders as measured zeros, two of them alarming
When a whole window is unmeasured, every System rollup field carries `0.0`, and two of those
zeros read as alarming rather than as missing.

## What it looks like

A window-wide NVML outage, or the first seconds of a run before any `cpu_percent` lands:

| field | value | how it reads |
|---|---|---|
| `cpu.now` / `p50` / `p95` | 0.0 | an idle host |
| `gpu_util`, `gpu_delta` | 0.0 | idle, evenly balanced |
| `gpu_mem.now` | 0.0 | no memory in use |
| `gpu_mem.headroom` | 0.0 | **memory full** |
| `ram.headroom` | `ram_total - 0.0` | **all RAM free** |
| `temp.status` | `"OK"` | a verdict derived from no reading |

The last three are the sharp ones. Two point the wrong way about pressure, and `temp.status` is a
derived judgment computed from a value nothing measured.

## Where

`renderers/system/dashboard_compute.py`. `_nan_pct` and `_last_measured` both return `0.0` when
every tick in the window is absent, and the rollup fields are plain floats, so there is nowhere for
"not measured" to go.

## Why it was not fixed alongside the per-tick work

The change that introduced those two helpers narrowed this defect rather than creating it: before,
ANY window containing a single blind tick folded that tick in as a measured zero, both on the chart
and in the percentile. Now only a window where NOTHING was measured falls back, and the charts draw
gaps in every partial case.

Closing the remaining case means the rollup fields become optional and the tiles gain an n/a path,
which is a card-surface change rather than a compute one. `format_gb_pair(None, ...)` already
returns `("n/a", "")`, so the pattern exists; it needs wiring through the payload and applying per
tile.

## Suggested fix

Let `_nan_pct` and `_last_measured` return `None` for an all-absent window, make the affected
rollup fields `Optional[float]`, and give the tiles the same n/a treatment the GPU tiles already
get when no device reported. `temp.status` should abstain rather than say "OK" when there is no
temperature.

Related: the same absence-versus-zero question, one level down, is what the per-tick and per-device
work addressed.

Pendu:
Closes #437.

The same argument as the last change, one level up: **an absence must not render as a
measurement.** That one was about a tick or a device inside a window. This one is about a window
in which nothing was measured at all, which still collapsed to `0.0` and rendered as a real reading
of an idle machine.

## The issue named the wrong fields, and this corrects it

#437 called out `gpu_mem.headroom`, `ram.headroom` and `temp.status` as the sharp cases. Measured:
**no consumer reads any of the three.** `headroom` has never appeared in `system_section.py` in
that file's entire history; it entered the typed payload in #422 as a faithful port of the
pre-existing dict field. `temp.status` is carried and never read.

The fields that DO reach a reader were not named. They are these, measured on the base at
`2fb9523`:

| surface | condition | rendered | reads as |
|---|---|---|---|
| dashboard RAM tile | host `ram_used_bytes` all NULL | `0.0 / 200 GB` | no memory in use |
| dashboard CPU label | host `cpu_percent` all NULL | `0%` | an idle host |
| dashboard GPU MEM tile | devices live, `mem_used_bytes` all NULL | `0.0 GB` | no memory in use, denominator gone |
| dashboard GPU TEMP tile | devices live, `temperature_c` all NULL | `0 °C` | a broken sensor |
| dashboard GPU UTIL tile | any | `n/a` | correct already |

## The device guard is not a metric guard

The third and fourth rows are the ones worth pausing on, because the card already has an
absence guard and it does not fire there. `gpus_unreported` asks whether the DEVICE said
anything, and these devices do: they carry a memory total and a power limit. Only utilisation was
covered, because #432 gave utilisation a count of its readings and gave the other metrics nothing.

So a host can have four live cards whose memory and temperature were never read, and the card
prints `0.0 GB` and `0 °C` beside a `n/a` utilisation tile.

## The terminal card had it worse

Both System surfaces read the same rows, and fixing one of two paths that answer the same question
is what leaves the next reader believing the wrong one. Rendered on a host whose readings were all
NULL:

```
before                                   after
CPU 0.0%      RAM 0.0%                   CPU N/A      RAM N/A
GPU UTIL N/A  GPU MEM 0.0%               GPU UTIL N/A GPU MEM N/A
GPU TMP 0.0°C GPU HDRM 15.0 GB           GPU TMP N/A  GPU HDRM N/A
```

Five of the six cells were fabricated. `GPU UTIL` is the only one that was already right, and it
is the only one that counts its readings, which is the whole point.

`GPU HDRM 15.0 GB` is the one that points the wrong way: free memory computed as capacity minus a
level nothing measured, so an unread card reports as entirely free. That is the failure the issue predicted for
the dashboard, where it turned out to be invisible; on the terminal card it is rendered.

## And the path with no data at all

Found in review, and the one that needs no failure at all to reach: before the
first sample lands, the terminal computer returns a hardcoded snapshot, and it hardcoded
`cpu=0.0, ram_used=0.0, ram_total=0.0`. The renderer is registered unconditionally, so **every run
opened by reporting a host at 0% CPU**, and the card fell back to the same payload after a read
failure outlived its stale window.

Only CPU showed through: a zero `ram_total` is falsy, so RAM already read `N/A` by accident. The
dashboard's equivalent was already coherent, because its empty payload carries empty rollups and
every field arrives as `None`.

That is this PR's own argument surviving on the one path I had not swept. I checked the paths that
aggregate readings and never asked what the path with no readings emits.

## One more twin, in the multi-node view

`SystemCLIClusterBuilder` already counted readings per metric and abstained correctly for memory,
temperature, headroom and the GPU mean. Host CPU was the single exception, and `cpu_percent` was
also the one field of `_NodeSystemSample` that was not optional. A silent node was folded into the
cluster median as a real 0%, and since zero is the lowest CPU any node can report, it also won the
worst-node comparison outright.

Measured on two nodes both pinned at 80% CPU, with the second node's readings not landing:

```
before   CPU 40.0% / 80.0% n0
after    CPU 80.0% / 80.0% n0
```

Exactly half, which is the arithmetic signature of counting an absence as a zero, and the same one
#429 measured on a four-GPU host that read 75%.

That module had no tests at all. This adds two, covering exactly the line that changed.

## The tests were checked against the old code

Seven of the thirteen added tests fail on `2fb9523` and pass here. The other five pass on both trees on
purpose, and they are the ones that make the change safe rather than the ones that prove it:

- one freezes the card's existing absence rendering, which this change relies on and does not alter
- two assert that a partly measured window or host still reports its readings
- three assert the invariant in the other direction, that a device genuinely sitting at 0% drawing
  no power, and a node genuinely at 0% CPU, still report `0.0` and never abstain

That last group matters most. Confusing absence with zero the other way round would hide real
measurements, which is worse than the defect being fixed, and nothing pinned it before.

## What the change is

The three window helpers return `None` when nothing was measured, `temp.status` abstains rather
than saying `"OK"` about a temperature nobody took, and the terminal computer counts each metric's
readings separately, because a device can report a memory total and no temperature.

The surfaces are almost untouched: `format_gb_pair(None, ...)` already returns `("n/a", "")`,
`_num(None)` already returns `"n/a"`, the CPU label already tested `is not None`, and
`fmt_percent(None)` already returns `"N/A"`. The card already knew what to do with an absence. Only
one render line needed a guard, the terminal RAM percentage, which multiplied a level by 100 before
dividing.

## Deliberately not done

`ram.headroom`, `gpu_mem.headroom` and `temp.status` on the dashboard now abstain because they sit
on the changed code path, but they still have no consumer. Deleting them or rendering them is a
separate decision, and `SystemRollups.status` is the same shape and already #426's third part.

The cluster builder is the one path of the three with no reported-ness guard, so a legacy
all-zero placeholder row still averages into its node mean. That is #429's question rather than
this one, and its only reachable input is the older trace format, so it is filed as #440.

`gpu_power_usage` and `gpu_power_limit` are in the terminal payload and nothing renders them, so
their abstention is invisible too. Same shape as the dashboard's headroom fields above.

`SystemCLIClusterBuilder` keeps its own local equivalent of the shared reading helper. It is
already correct, so there is no defect to fix there, only duplication; folding it in means proving
equivalence across a broader `except`, which #436 showed is not free.

## Scope

```
STRUCTURE: an all-absent window's abstention -> renderers/system/dashboard_compute.py and
cli_compute.py (owners of what a metric means) plus dashboard_models.py and common.py (the two
payload contracts); the optional-reading helper -> renderers/system/common.py, mirrored from
gpu_reported, which moved there in #429 for the same reason; boundaries crossed: none
DATA PATH: system_section <- SYSTEM_LAYOUT <- SystemRenderer <- SystemDashboardComputer
           <- SystemRepository <- system_samples + system_gpu_samples
SCOPE: declared renderers/system, both surfaces, plus their tests -> diff touches
dashboard_compute.py, dashboard_models.py, cli_compute.py, cli_cluster.py, common.py, renderer.py
and four test files -> within
TWINS: searched the CONCEPT, not the symbol - every place an all-absent set becomes a number.
Three window helpers on the dashboard (fixed), a derived verdict with no reading (fixed), five
coercions and a zero-initialised max on the terminal card (fixed), one coercion in the cluster
view (fixed), and a hardcoded no-data snapshot (fixed). Checked and left: renderers/shared/run_series and the reporting loader already use
absence-aware aggregation; reporting/sections/system/loader.py:111-121 still lacks a
reported-ness test, which is #429's noted twin and a different question
```

## Verification

```
GATE: pytest tests/ -> 1662 passed, 2 skipped against the base at 2fb9523 measured at 1649 passed, 2 skipped; the delta is the thirteen tests added here; black, isort and ruff clean at 79 on every file in the diff
TRANSITIONS: none; no state machine is touched. The recent-to-retained switch, the cached-payload TTL and the whole-run mode are unchanged and their tests are untouched
RANKS: the multi-node cluster view is in this diff and gained the two tests above; the System dashboard remains single-node by construction and its node selection is unchanged
TRIPWIRE: no all-absent set in renderers/system may reach a tile, a label or a terminal cell as a number, and no real reading may be turned into an abstention. Both computers abstain, each metric counts its own readings rather than sharing a row-level count, the tests assert the abstention on both surfaces plus a partly-measured case and a genuinely-zero case on each, and a NaN probe across eight payload shapes on both surfaces finds zero NaN in any field
SCENARIOS: all 22 fixture cells rendered on both trees and diffed field by field; 0 of 676 field values differ across the strip and the System card. That is the expected result and is NOT evidence the fix works: no fixture carries an all-absent window, so the matrix cannot reach this bug. Its job here is to show the measured paths are untouched. The evidence is the before-and-after measurements above and the thirteen tests
```


abhinavsriva:
@Pendu  The core dashboard fix looks correct and the separation between compute, payload models, and presentation is clean.

I only have two scope comments:

1. We currently support single-node execution, including multiple processes and multiple GPUs. The `cli_cluster.py` change is for multi-node behavior, so please remove it and its new tests from this PR. It can be handled later when multi-node support is in scope.

2. The current sampler reads all metrics for each GPU inside one `try`. Therefore, an individual GPU either reports all its metrics or reports all of them as `None`; it does not currently produce a row with capacity present but memory usage or temperature missing.

   For the partial-metric handling, either option is fine:
   - Remove that extra handling and its synthetic tests to keep this PR minimal; or
   - Keep it as defensive schema support, but calculate GPU memory used and total from the same set of GPUs. Currently one GPU can contribute usage while multiple GPUs contribute capacity, which can produce a misleading percentage.

Please choose whichever requires the smaller clean change. No larger refactor or additional abstraction is needed. The remaining dashboard and single-node absence handling looks good.

abhinavsriva (dismissed):
plz review the comments.

Pendu:
Both done.

1. `cli_cluster.py` is reverted and its test file removed. The multi-node fix is out of this PR.

2. Took the first option, and it gets the second one's property as well. There is now one count, over the devices that reported, so memory used and memory total can never be summed over different sets. The synthetic partial-metric tests are gone; the dashboard one was rewritten to use the reachable trigger instead, a host where every device writes the NVML all-zero row, since that is what the sampler actually produces.

One thing worth your call, not addressed here. #429 established the opposite treatment for utilisation: `util_gpu_count` exists precisely so a live device with a NULL `util` column is excluded from the mean, on the argument that reported is about the device while the mean counts readings. By the sampler argument that case is equally unreachable. So the block now counts utilisation by readings and memory by devices. Not wrong, but inconsistent, and I would rather you pick than have me change a merged decision unasked.

PR #443, opened 2026-09-03
Dev/abhinav/ownership

abhinavsriva:
## Summary

This PR moves training stdout/stderr ownership from the in-process TraceML runtime to the outer node launcher. It builds on previous reusable `ProcessOutputDrainer` and completes the ownership switch without adding another abstraction layer.

## Why this is one PR

Several launcher, runtime, display, test, and documentation files change because they participate in the same output-ownership boundary.

The launcher must begin draining stdout/stderr when runtime stream replacement, the old sampler, and the CLI output panel are disabled. Splitting these changes would create an inconsistent intermediate version where output could be swallowed, duplicated, or interfere with the Rich display.

Previous PR already isolated the complex pipe-draining primitive. This PR is the smallest complete integration of that primitive.

## What changed

- Save training output by default in separate node-scoped files:
  - `nodes/node_<node_rank>/training.stdout.log`
  - `nodes/node_<node_rank>/training.stderr.log`
- Add:
  - `--save-training-output`
  - `--no-save-training-output`
- Capture output reaching the supervised process tree, including Python writes, native FD writes, torchrun diagnostics, and inherited worker output.
- Mirror both streams live in `summary` and `dashboard` modes.
- Keep training output out of the Rich display in `cli` mode.
- After a CLI failure, print a bounded stderr excerpt, saved paths, and the final training outcome.
- Record output configuration, distributed path patterns, and confirmed local artifacts in the run manifest.
- Fall back to terminal output when persistence fails without changing the training exit code.
- Preserve user buffering choices; TraceML does not force `PYTHONUNBUFFERED`.

## Removed from active runs

- Runtime replacement of `sys.stdout` and `sys.stderr`
- Active `StdoutStderrSampler` registration
- The stdout/stderr Rich panel
- `--capture-stderr`
- `TRACEML_CAPTURE_STDERR`
- New production of `crash_stderr.log`

Legacy readers, SQLite projections, and artifact decoding remain available for historical runs.

## Output scope

Training output is intentionally per node, not per rank. Each launcher owns one local torchrun process tree, so output from ranks on the same node may be combined. Per-rank raw output would require a separate torchrun-native redirection design and is outside this PR.

## Reliability and validation

- Both streams are drained concurrently to prevent pipe deadlocks.
- Path, file-open, and write failures fail open to the terminal.
- Shutdown terminates the process tree before completing the drain.
- Training exit status remains authoritative.
- No-save mode inherits the original descriptors and creates no output files.
- Legacy stdout/stderr history remains readable.

Full test suite: `1,640 passed, 2 skipped`.

## Out of scope

- Aggregator stderr capture — RFC PR7. (this is PR6 in the series) 
- TraceML internal error-log consolidation — RFC PR8
- Per-rank output files
- Rotation, compression, upload, or log search
- Removal of historical artifact readers


Pendu:
This is the strongest change in the series: it removes an ownership path rather than adding one, and it collapses three overlapping capture mechanisms into one. The integration is careful and most of it holds. Two things I would fix before merge, both of which undercut the feature's own purpose, and both reproduced end to end.

## Verified

- Full suite `1658 passed, 2 skipped` at head against `1649 passed, 2 skipped` at the merge base, so exactly the nine added tests and no regressions.
- Capture is process-tree wide, not just Python level. A script writing through `print`, `sys.stderr.write`, and raw `os.write(1, ...)` / `os.write(2, ...)` lands correctly split across `training.stdout.log` and `training.stderr.log`.
- `--no-save-training-output` creates no node directory at all and leaves output on the terminal; the default path both mirrors live and persists the same bytes.
- Exit-code fidelity is unchanged from base. A script exiting 3 yields launcher exit 1 on both trees (torchrun flattens worker failures), and a clean run yields 0 on both. I checked because the description claims the training status stays authoritative, and it does, in the sense that the base already behaved this way.
- The removals are complete rather than partial. No reference to `capture_stderr`, `TRACEML_CAPTURE_STDERR` or `crash_stderr` survives in `src/` or `docs/`; `StdoutStderrSampler` is deregistered; no display surface is left showing a panel that can no longer populate.
- Both removals carry a `**Breaking:**` marker in the CHANGELOG. That is the right call and it is the thing most often skipped.
- Shutdown ordering is correct: the process group is terminated before `cleanup()` runs, so the drain cannot block waiting on a live child, and `cleanup()` is wrapped so a failure there cannot break shutdown.

## Fix before merge

**1. The drain timeout can kill a user process that outlived training.**

`finish()` joins for `DEFAULT_SHUTDOWN_TIMEOUT_SEC` (5s), then closes the read end of the pipe from a separate thread (`process.py:258-264`). EOF only arrives when every holder of the write end has closed it, so any process that inherited the pipe and outlives the direct child is still writing when the read end goes away. It gets EPIPE, or SIGPIPE if it is not Python.

Reproduced with a script that spawns a detached grandchild in a new session and exits immediately, while the grandchild prints for 20 seconds:

```
under traceml : 13 of ~40 lines captured, grandchild DIED
standalone    : 40 of 40 lines,           grandchild lived
```

The grandchild is in its own session, so the process-group termination cannot reach it. The pipe close is what killed it. That makes this more than lost output: TraceML terminates a user process it did not start and was never asked to manage. Given the rule that instrumentation must never interfere with the user's processes, I think this one has to be handled before this lands, even if the handling is only to leave the read end open and accept a leaked thread, or to make the timeout generous and say what happens past it.

**2. Dropping `PYTHONUNBUFFERED` loses exactly the output this feature exists to save.**

`env["PYTHONUNBUFFERED"] = "1"` is removed in this diff. Because stdout is now a pipe rather than a terminal, CPython switches the child from line buffering to block buffering, so unflushed output dies with the process. The description frames this as preserving the user's buffering choice, which is reasonable in isolation, but the interaction with the new pipe is what makes it costly.

Measured with a child writing 200 lines without flushing and then exiting abruptly, stdout on a pipe in both cases:

```
with PYTHONUNBUFFERED=1 (base behaviour) : 200 of 200 lines survive
without it (this PR)                     :   0 of 200
```

Not a partial tail, everything. So on an OOM kill, a scheduler kill, or a second Ctrl-C, `training.stdout.log` is empty or truncated precisely when someone needs it, and at base that same output reached the terminal. If preserving the user's buffering is the priority, the tradeoff is worth stating in the CHANGELOG next to "saved by default", because a reader will assume saved means saved.

## Worth deciding

**Default-on capture is unbounded, and it replaces something that was bounded.** The sink is opened with no size cap (`process.py:175`), while the artifact it supersedes was limited to a 64 KiB tail. The CHANGELOG says the new capture "supersedes the optional bounded `crash_stderr.log` tail" without noting that bounded became unbounded and opt-in became default-on. A long run with per-step logging now duplicates its entire output onto the logs volume on every node, and the only escape is `--no-save-training-output`, which gives up capture completely. A bounded middle setting would cover the Slurm-scratch case.

**A partial or timed-out file is reported nowhere.** A path is recorded only when the stream reached EOF with no sink failure (`process.py:226-228`), and `finish()` further excludes timed-out streams. The launcher passes only those paths into the manifest (`commands.py:921-934`), so after a mid-run write failure the bytes are on disk at a known location while the manifest has no entry and nothing prints the path. Reporting the attempted path and letting the existing `path.exists()` filter decide would close it, and would also cover case 1 above.

## Smaller

- `finish()` is now reachable both from the normal path (`commands.py:864`) and from the signal handler via `cleanup` (`:713-721`). Both are the main thread, so this is re-entrancy rather than the concurrent case, and the consequences today are benign because `close()` is idempotent and the handler exits. Worth noting that a Ctrl-C inside that window turns a completed `rc=0` run into exit 130.
- In cli mode the mirror is disabled to protect the Rich display, but the fallback is still the terminal in every mode (`commands.py:174-183`), so a persistence failure in cli mode streams raw training bytes into the display the policy was protecting. Defensible, since preserving output should win, but the two rules collide exactly in the mode where mirroring was removed.
- `StdoutStderrRenderer` is now unreachable. It was wired at `cli.py:93` on the parent commit and is now referenced only by its own definition. The description says legacy readers stay available for historical runs, which is true of the tables and the decoder but not of the renderer, so either it wants a path to reach it or it is dead code for a later PR to delete.

abhinavsriva:
Documented both trade-offs. We intentionally preserve user buffering semantics and do not force `PYTHONUNBUFFERED`. Detached processes are outside the supervised process-tree lifecycle and should redirect their streams or use `--no-save-training-output`.

PR #444, opened 2026-09-04
stderr_added

abhinavsriva:
## Summary

This PR persists raw stderr from the launcher-owned TraceML aggregator:

```text
logs/<run-name>/aggregator/process.stderr.log
```

It builds on the launcher-owned process-output infrastructure introduced in previous PRs.

## What changed

- Pipe and persist only aggregator stderr; stdout remains inherited.
- Mirror aggregator stderr in summary and dashboard modes.
- Keep stderr out of the Rich display in CLI mode and report its saved path when telemetry fails.
- Drain stderr during readiness failure, early exit, normal shutdown, and signal-driven teardown.
- Record the confirmed artifact path in `manifest.json`.
- Treat persistence failures as warnings without changing the training result.
- Restrict aggregator stderr ownership to node 0 in multi-node runs.
- Document the artifact, display behavior, flag boundary, and distributed ownership.

## Reliability

Tests cover:

- Exact binary stderr emitted before aggregator readiness.
- Aggregator sink failure and terminal fallback.
- Summary/dashboard mirroring and CLI suppression.
- Nonzero aggregator exit without stopping training.
- Normal shutdown ordering.
- Strict startup failure reporting.
- Manifest artifact recording.
- Node-0-only ownership.
- Independence from `--no-save-training-output`.

## Behavior preserved

- The training process remains authoritative after it starts.
- Aggregator stdout and Rich presentation output are not persisted.
- Standalone `traceml serve` remains unchanged.
- Existing training stdout/stderr behavior from PR 6 remains unchanged.
- Capture failures cannot fail training.

## Out of scope

- Internal TraceML error-log consolidation, which belongs to the next PR
- Per-rank training output.
- Output rotation, compression, upload, or remote log storage.
- A generic process-logging framework.

## Validation

- Focused launcher/output tests: `113 passed`
- Full suite: `1,664 passed, 2 skipped`
- `git diff --check` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Launcher-owned aggregator stderr is now saved separately at
+  `logs/<run-name>/aggregator/process.stderr.log`. Summary and dashboard modes
+  mirror it live; CLI mode preserves the Rich display and reports the saved
+  path if telemetry fails.
 - Training stdout and stderr are now saved by default in separate node-scoped
   files under `logs/<run-name>/nodes/node_<node-rank>/`. Summary and dashboard
   modes mirror them live; CLI mode keeps the Rich display clean and prints a
--- a/docs/user_guide/distributed-training.md
+++ b/docs/user_guide/distributed-training.md
@@ -57,6 +57,11 @@ launcher, which owns the aggregator process and can observe its exit and
 finalization. Non-owner launchers do not infer final aggregator health from a
 startup TCP readiness check.
 
+Each node launcher writes its own `nodes/node_<node_rank>/training.*.log`
+files when training-output saving is enabled. Only node 0 writes
+`aggregator/process.stderr.log`, because it is the only launcher that owns the
+aggregator process.
+
 `--session-id` remains accepted as a backward-compatible alias for
 `--run-name`.
 
--- a/docs/user_guide/public-api.md
+++ b/docs/user_guide/public-api.md
@@ -134,6 +134,13 @@ prints a bounded stderr excerpt, the saved paths, and the final outcome. Use
 output files. Captured streams are pipes, so `isatty()` returns false. TraceML
 does not replace Python's `sys.stdout` or `sys.stderr` objects.
 
+`--no-save-training-output` controls only the supervised training streams. The
+owner launcher always saves raw aggregator diagnostics to
+`logs/<run-name>/aggregator/process.stderr.log`; only node 0 creates this file
+in a multi-node run. Summary and dashboard modes mirror it live, while CLI mode
+reports its path if telemetry fails. Aggregator stdout remains attached to the
+terminal and is not persisted.
+
 ## Direct Launch with `traceml serve`
 
 `traceml run` starts the aggregator and your script together. For direct
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -199,6 +199,19 @@ outlive training should redirect its own stdout and stderr, or the launch
 should use `--no-save-training-output`; otherwise the launcher eventually
 closes its output pipes during shutdown.
 
+The owner launcher also saves the aggregator's raw stderr separately:
+
+```text
+logs/<run-name>/aggregator/process.stderr.log
+```
+
+This file is a fallback for startup failures, native diagnostics, and errors
+emitted before structured logging is available. Summary and dashboard modes
+mirror it live. CLI mode does not mirror it over the Rich display; if telemetry
+fails, TraceML reports the saved path. This artifact is distinct from
+`aggregator/traceml_errors.log`, which contains structured TraceML log records.
+Aggregator stdout remains attached to the terminal and is not persisted.
+
 Phases that were never measured in the current window are omitted from the
 live step-time table rather than shown as `0.0 ms`, and the diagnosis block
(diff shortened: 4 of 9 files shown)


Pendu:
This is the cleanest one in the series. The two hazards from the previous PR do not carry over, and not by luck: the ordering and the choice of stream both avoid them by construction. One stated claim does not hold on one of the four paths, one terminal line disappears in cli mode, and one inherited edge is now at its third call site.

## Verified

- Full suite `1664 passed, 2 skipped` at head against `1658 passed, 2 skipped` at the merge base, so exactly the six added tests and no regressions. Your focused count reproduces too.
- A real run writes both artifacts and finalises normally: `final_summary.json` at 23203 bytes alongside `aggregator/process.stderr.log`, with the manifest recording `training_stderr_log` and `aggregator_stderr_log`.
- The pipe-close hazard does not reach the aggregator, on any of the four paths. Readiness failure terminates at `commands.py:826` before draining at `:829`; normal shutdown terminates at `:940` with the finalize timeout before draining at `:949`; early exit only drains at `:1029` after `poll()` is already non-None; signal teardown terminates both processes in the handler before `cleanup()` runs. Since `terminate_process_group` returns only after the child is reaped, the aggregator is never writing when the read end closes. The final summary is safe for a further reason: it is produced inside the aggregator's own SIGTERM handling, before the launcher drains at all.
- Piping stderr is safe where piping stdout was not. CPython keeps stderr line-buffered regardless of whether it is a tty, so an abrupt exit loses nothing: 200 of 200 lines survive, against 0 of 200 for the stdout case in the previous PR. Worth stating because the two look like the same change and are not.
- Capture failure cannot fail training. Making only the log path unopenable, leaving the rest of the directory writable, gives exit 0, training runs, the warning surfaces, and the summary is still written. My first attempt at this test used a read-only `aggregator/` directory and showed exit 1, but that also breaks the telemetry database, so it was measuring the aggregator failing to start rather than this feature.
- Ownership is right by construction rather than by a second predicate: the drainer is only built inside `if owns_aggregator:`, so a non-owner creates no file and records no path. That reuses the same decision as the manifest writer instead of adding another one.
- The docs match the code on mode mirroring, node-0-only, the `--no-save-training-output` boundary, and aggregator stdout staying attached.
- Calling `finish` through a memoised wrapper (`commands.py:750`) is a real improvement over the previous integration, since the four paths can each reach it.

## The claim that does not hold: signal teardown records nothing

The description lists signal-driven teardown as one of the four drain paths and separately says the confirmed path is recorded in `manifest.json`. Both are true individually and false together. The handler writes `update_run_manifest(manifest_path, status="interrupted")` with no artifacts, then terminates, then runs cleanup; the drain happens after the only manifest write, and nothing writes again.

Measured with a Ctrl-C mid-run:

```
launcher exit=130
process.stderr.log on disk: PRESENT (131B)
manifest: status=interrupted, artifacts=['code_manifest']
```

The file is written, it just is not referenced. The training logs are missing from the same manifest for the same reason, so this is one gap covering both artifacts rather than something new to this PR alone.

This matters more than it looks because of what else is true on that path. Signal teardown gives the aggregator `DEFAULT_SHUTDOWN_TIMEOUT_SEC` to finalise, against `finalize_timeout + 5` on the normal path, so a Ctrl-C typically kills finalisation partway and there is no summary. That is pre-existing and not this PR's doing, but it means the interrupted run is exactly the case where someone goes looking for the aggregator's stderr to find out what happened, and it is exactly the case where the manifest does not point at it.

## One line disappears from cli mode

Aggregator stderr was previously inherited, so the aggregator's own messages reached the terminal in every mode. Head pipes it and sets no mirror in cli mode, so these three now go only to the file:

```
[TraceML] Aggregator ready on 127.0.0.1:29765 ... Press Ctrl+C to stop.
[TraceML] Logs saved under: /path/to/logs/<run-name>
[TraceML] Aggregator stopped.
```

Keeping the Rich display clean is the point, and the launcher prints its own `Aggregator ready.` at `commands.py:871`, so the first and third are covered. The middle one is not: in cli mode a user no longer sees where their logs went. Either the launcher should print the run directory itself in that mode, or this is a deliberate trim worth saying so.

## Third call site for the same edge

A partial or timed-out drain leaves the file on disk while the result reports `stderr_path=None`, so `aggregator_output_artifacts()` returns None and the manifest omits a file that exists. That is the same edge as in the two previous PRs, and this is now its third consumer.

It is small each time and it keeps reappearing because each caller inherits it rather than causing it. Returning the attempted path plus a completeness flag from the drainer, and letting the existing `path.exists()` filter decide, would close it once for all three call sites instead of a fourth caller inheriting it next.

## Coverage

The new tests are real regression tests and fail against the parent commit, since `_start_aggregator_output` did not exist and the aggregator was not spawned with a piped stderr. What they do not reach is the interaction: every launcher test mocks the shutdown handler and both drainer constructors, so nothing exercises two live drainers, the signal path calling finish on both, or the manifest gap above. That is also the shortest path to a test for it.

abhinavsriva:
Thanks.

The signal teardown manifest gap is fixed: after both drainers finish, their confirmed training and aggregator paths are merged into `manifest.json`. Regression coverage was added.

The other two observations are intentionally unchanged:

- CLI suppression is the documented display policy; the launcher reports relevant paths without replaying aggregator status lines over Rich.
- Timed-out or partial drains remain unreported because the current contract exposes only confirmed, completed artifacts. Adding partial-artifact state would require a broader manifest contract and is outside PR 7.

PR #448, opened 2026-09-04
Dev/abhinav/log consolidate

abhinavsriva:
## Summary

This PR consolidates TraceML-owned errors into one structured log per process owner.

## What changed

- Added explicit logger roles with collision-free paths:
  - `rank_<global_rank>/traceml_errors.log`
  - `aggregator/traceml_errors.log`
  - `nodes/node_<node_rank>/launcher_errors.log`
- Routed runtime, aggregator, and launcher implementation failures to their respective owners.
- Removed production of:
  - `torchrun_error.log`
  - `runtime_error.log`
  - `aggregator_error.log`
- Kept user exceptions in native Python/torchrun stderr rather than TraceML internal logs.
- Made logger setup idempotent and prevented propagation into user root loggers.
- Ensured logger initialization, write, and rotation failures remain fail-open without polluting workload stderr.
- Ensured aggregator startup and fatal finalization failures are logged exactly once.
- Preserved aggregator raw stderr separately at `aggregator/process.stderr.log`.

## Scope

This PR owns structured TraceML implementation diagnostics only. 

## Out of scope

- JSON logging or a new logging framework
- Cloud upload, log collection, or search
- Rotation-policy changes
- Per-rank raw workload-output capture
- Removal of legacy readers for historical artifacts

## Validation

- `1680 passed, 2 skipped`
- `mkdocs build --strict`
- `git diff --check`
- Verified that combined user and TraceML shutdown failures do not copy the user exception into internal logs
- Verified that logger write failures do not affect execution or terminal output
- Verified exact-once aggregator failure logging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- **Breaking:** TraceML implementation errors now have one structured owner:
+  `rank_<global_rank>/traceml_errors.log`,
+  `aggregator/traceml_errors.log`, or
+  `nodes/node_<node_rank>/launcher_errors.log`. New runs no longer create the
+  duplicate `torchrun_error.log`, `runtime_error.log`, or
+  `aggregator_error.log` files.
 - Launcher-owned aggregator stderr is now saved separately at
   `logs/<run-name>/aggregator/process.stderr.log`. Summary and dashboard modes
   mirror it live; CLI mode preserves the Rich display and reports the saved
--- a/docs/developer_guide/extending.md
+++ b/docs/developer_guide/extending.md
@@ -236,6 +236,12 @@ TraceML should not break user training because optional telemetry, rendering,
 or reporting failed. Existing code logs advisory failures through
 `traceml_ai.loggers.error_log.get_error_logger`.
 
+Process entrypoints configure that logger once with the `rank`, `aggregator`,
+or `launcher` role. Components should request a named child logger rather than
+adding handlers or choosing paths themselves. Internal logs accept ERROR-level
+TraceML implementation failures only; do not send user exceptions or workload
+stderr to them.
+
 Use that pattern for non-critical paths:
 
 ```python
--- a/docs/developer_guide/extending.md
+++ b/docs/developer_guide/extending.md
@@ -240,16 +240,23 @@ Process entrypoints configure that logger once with the `rank`, `aggregator`,
 or `launcher` role. Components should request a named child logger rather than
 adding handlers or choosing paths themselves. Internal logs accept ERROR-level
 TraceML implementation failures only; do not send user exceptions or workload
-stderr to them.
+stderr to them. Use `log_internal_exception` when recording a caught exception;
+it preserves explicit TraceML causes without copying implicit context from a
+user exception that is already unwinding.
 
 Use that pattern for non-critical paths:
 
 ```python
+from traceml_ai.loggers.error_log import (
+    get_error_logger,
+    log_internal_exception,
+)
+
 logger = get_error_logger("MyComponent")
 try:
     ...
 except Exception as exc:
-    logger.exception("[TraceML] MyComponent failed: %s", exc)
+    log_internal_exception(logger, "[TraceML] MyComponent failed", exc)
 ```
 
 Prefer returning an empty payload, `NO DATA` diagnosis, or fallback text over
--- a/docs/user_guide/integrations/ray.md
+++ b/docs/user_guide/integrations/ray.md
@@ -238,3 +238,7 @@ then stops the actor in a ``finally`` block. Each worker also stops its local
 TraceML runtime in a ``finally`` block. Normal exceptions and keyboard
 interrupts should therefore release TraceML resources. A hard ``SIGKILL`` cannot
 run Python cleanup code in any framework.
+
+If aggregator finalization fails, the actor records the failure in
+``aggregator/traceml_errors.log``. Cleanup remains best effort at the Ray driver
+boundary, so a TraceML shutdown failure does not replace the Ray Train result.
--- a/docs/user_guide/public-api.md
+++ b/docs/user_guide/public-api.md
@@ -141,6 +141,12 @@ in a multi-node run. Summary and dashboard modes mirror it live, while CLI mode
 reports its path if telemetry fails. Aggregator stdout remains attached to the
 terminal and is not persisted.
 
+TraceML implementation errors are stored separately in
+`rank_<global_rank>/traceml_errors.log`,
+`aggregator/traceml_errors.log`, and
+`nodes/node_<node_rank>/launcher_errors.log`. User exceptions remain in the
+native training stderr artifacts above.
+
 ## Direct Launch with `traceml serve`
 
 `traceml run` starts the aggregator and your script together. For direct
--- a/docs/user_guide/reading-output.md
+++ b/docs/user_guide/reading-output.md
@@ -212,6 +212,21 @@ fails, TraceML reports the saved path. This artifact is distinct from
 `aggregator/traceml_errors.log`, which contains structured TraceML log records.
 Aggregator stdout remains attached to the terminal and is not persisted.
 
+### TraceML internal error logs
+
+TraceML implementation failures are kept separate from workload output. Each
+owning process writes ERROR-level records to one rotating file:
+
+```text
+logs/<run-name>/rank_<global_rank>/traceml_errors.log
+logs/<run-name>/aggregator/traceml_errors.log
+logs/<run-name>/nodes/node_<node_rank>/launcher_errors.log
+```
+
+User exceptions remain in native training stderr and are not copied into these
+files. The internal logger does not add a terminal handler; terminal output
+continues to follow the launcher and display policies described above.
+
 Phases that were never measured in the current window are omitted from the
 live step-time table rather than shown as `0.0 ms`, and the diagnosis block
 shows `INCOMPLETE DATA` with the missing signal names when no reliable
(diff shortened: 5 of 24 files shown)


Pendu:
Reviewed at `d6be512` against `version_0.3.7` (merge-base `463b141`), head and base checked out side by side. No prior human review on this PR, and CodeRabbit skipped it because the base is not the default branch, so there is no earlier feedback to reconcile. The PR is 3 commits behind its base with no overlapping files.

I verified your validation block first and it all reproduces: `1680 passed, 2 skipped`, `mkdocs build --strict` clean, `git diff --check` clean, and the full pre-commit set green at the pinned versions.

The consolidation is right, and it fixes three real defects in the old code that are worth naming because they were invisible before:

1. The old structured log never recorded the traceback. `logger.exception(...)` was called from the `finally` block, outside the `except` clause, so `sys.exc_info()` was empty. Forcing an aggregator startup failure on base writes literally `[TraceML] Aggregator exiting due to error` followed by `NoneType: None`. On this branch the same failure records the full traceback, because you switched to an explicit `exc_info=` tuple.
2. A diagnostic write failure used to dump a `--- Logging error ---` block with a traceback onto the workload's stderr. `_SilentRotatingFileHandler` ends that. I confirmed it for both plain write failures and rollover failures.
3. On base, `setup_error_logger` raised `TypeError` when `TRACEML_LOGS_DIR` or `TRACEML_SESSION_ID` was unset, because `Path(None)` is not caught anywhere. This branch falls back to `./logs/default/...` instead.

Also worth recording: a user who attached their own handler to the `traceml_ai` logger used to silently disable TraceML's own error file entirely, because `if logger.handlers: return logger` fired before the file handler was ever attached. On this branch the file is written. That is a strict improvement.

Collision-free paths check out on a real two-rank run (`--nproc-per-node 2`): four distinct owners, `rank_0/`, `rank_1/`, `aggregator/`, `nodes/node_0/`. The consumer sweep for all nine removed symbols is clean, the only remaining mentions being the negative assertions in tests. I also mutation-tested the exactly-once work rather than reading it. Replacing `if handle is not None` with `if True` at `src/traceml_ai/aggregator/aggregator_main.py:223` fails `test_serve_configures_logging_without_preset_env`, so that half is pinned.

Two things I would fix before merge, both small.

**`propagate = False` is set before the handler is attached, so TraceML records can reach the workload's stderr during setup.** `src/traceml_ai/loggers/error_log.py:72` sets `logger.propagate = False` at the top of the function, but the handler is not attached until line 116 or 127, after path resolution, `mkdir`, and the file open. In that window the logger has zero handlers and no propagation, which is exactly the condition where Python falls back to `logging.lastResort` and writes the bare record to stderr. Base did not have this window because it set `propagate` after `addHandler`.

Measured with a watcher thread sampling the logger while `setup_error_logger(role="rank")` runs, counting records that reached the process's real stderr:

```
base : 0 zero-handler windows,     0 records on stderr
head : 68298 zero-handler windows, 68298 records on stderr
```

The window is wide because it spans filesystem work, not because of a tight race. I could not construct a first-party flow that logs from another thread during it, since every entry point calls setup before any TraceML thread starts, so this is a latent hole rather than a live break. But it contradicts the "without polluting workload stderr" goal directly and the fix is to attach the handler before flipping `propagate`, which is what base did.

**`chain=False` also drops `__cause__`, not just the user's in-flight exception.** `src/traceml_ai/runtime/executor.py:61-68`. Suppressing the chain is right for `__context__`, which is how you keep a user exception out of the log while `stop_runtime` runs in the `finally`, and `test_shutdown_log_excludes_active_user_exception` pins that well. But it also drops explicit `raise X from Y` chains inside TraceML's own code, so the root cause of a TraceML failure is no longer in any TraceML log.

Same probe on both trees, where `start_runtime_handle` raises `RuntimeError("OUTER") from ValueError("ROOT_CAUSE")`:

```
base : runtime_error.log contains "ROOT_CAUSE" (2 hits) with
       "The above exception was the direct cause of the following exception"
head : rank_0/traceml_errors.log contains "OUTER" only, "ROOT_CAUSE" 0 hits
```

Since this branch makes these files the canonical diagnostic surface, losing the cause chain costs more here than it did before. Keeping `__cause__` and clearing only `__context__` gets both properties.

Two places where the tests do not hold the line, both worth closing while the code is fresh.

**The other half of the exactly-once fix is not pinned by anything.** Restoring the eager `self._logger.exception("[TraceML] generate_summary raised")` inside the `except` at `src/traceml_ai/aggregator/trace_aggregator.py:275`, which is the literal reversal of the change you made, recreates the duplicate record at the fatal finalization boundary and the entire suite still reports `1680 passed, 2 skipped`. `test_fatal_summary_failure_is_logged_once_at_process_boundary` passes too, because it asserts `logger.error.assert_called_once()` at `tests/aggregator/test_finalization.py:516` while the duplicate arrives through `logger.exception`, a different attribute on the same Mock. Asserting on the record count in the real log file, or on the Mock's total call count across both methods, would close it.

**The end-to-end test for the user-exception boundary cannot fail.** `tests/runtime/test_executor.py:306-340` sets `TRACEML_DISABLED: "1"` at line 321, so no rank logger is ever created and line 340's `assert not list(run_root.rglob("traceml_errors.log"))` passes regardless of policy:

```
TRACEML_DISABLED=1 -> 0 traceml_errors.log files
TRACEML_DISABLED=0 -> 1 traceml_errors.log file
```

Lines 338 and 339 still pin the removal properly. For 340, running the same subprocess with TraceML enabled makes it a real test of the claim: the rank log gets created, and I confirmed it holds only TraceML's own failures with `subprocess boom` appearing 0 times.

Two questions where you may already have decided.

**`--no-save-training-output` now leaves no durable record of a user crash.** Running a crashing script through `traceml run --mode summary` on each tree:

```
base, --no-save-training-output : crash text in torchrun_error.log
head, --no-save-training-output : crash text in no file under the session directory
head, default                   : crash text in nodes/node_0/training.stderr.log
```

The default path is fine and the removal is clearly correct there. The opt-out path is the one where `torchrun_error.log` was not a duplicate but the only copy, so the CHANGELOG line calling it "duplicate" is accurate only with capture enabled. Is losing the record the intended consequence of the flag, and should the Breaking note say so?

**"Logged exactly once" is zero for a failure in `start_aggregator`'s preamble.** `src/traceml_ai/runtime/lifecycle.py:193-203` runs `session_id` resolution, `replace`, `_apply_settings_env`, and logger setup before its own `try` at line 204. Anything raised there propagates to `run_aggregator`, which leaves `handle` as `None` and skips its log at `src/traceml_ai/aggregator/aggregator_main.py:223`. Forcing `_apply_settings_env` to raise:

```
head : aggregator/traceml_errors.log is 0 bytes, nothing recorded anywhere on disk
base : 116-byte structured entry, plus the failure in aggregator_error.log
```

Narrow, and I could not reach it from real input either. If you want the invariant to hold unconditionally, a flag set once `start_aggregator` has logged reads more directly than inferring it from `handle`.

Smaller things, none blocking:

- `src/traceml_ai/loggers/error_log.py:50` builds `nodes/node_<n>` from a literal, but `node_artifact_dir` at `src/traceml_ai/launcher/manifest.py:30-32` already owns that layout. Both produce the same path today, and `manifest.py` imports only from `utils/`, so calling the owner introduces no cycle. Worth doing so the launcher log cannot strand if that layout moves again. Note the sibling role does use its owner, `rank_dir_name`.
- The failure banner at `src/traceml_ai/aggregator/aggregator_main.py:232-239` prints an absolute path to `aggregator/process.stderr.log`, which does not exist under `traceml serve` since nothing drains that process. The sentence hedges with "Launcher-owned runs", but base named a file that always existed. Printing it only when the file is present would avoid sending someone to a missing path.
- `setup_error_logger` removes and closes the existing handler at lines 94 to 98 before it knows a replacement can be built, so a later call whose path resolution fails downgrades a working file handler to a `NullHandler` and silences the rest of the process. I reproduced the mechanism but found no input that reaches it: identity resolution degrades bad `RANK` values to 0 rather than raising. Keeping the existing handler when the new path is `None` closes it.
- `clean_error_logger` at `tests/loggers/test_error_log.py:24-37` restores handlers and `propagate` but not `level`. Now that setup sets the level unconditionally, that leaks across tests in one pytest process.
- Every healthy run now leaves a 0-byte `nodes/node_<n>/launcher_errors.log`, since the handler opens its file at construction. Base created no launcher file at all. `delay=True` would keep the tree clean, though the aggregator file already behaves this way on base so this is consistent rather than new.
- The new error logs are not in the manifest's `artifacts` block, which lists `training_stderr_log` and `aggregator_stderr_log` but nothing from this PR. That gap predates the branch, since the three removed files were not listed either, but this is the change that makes these the canonical error surface, so it is the natural moment to register them. Same pattern I flagged on the previous PR.

For completeness, things I checked that came back clean: the `is_missing_torch` concern does not apply, because a torch that raises anything other than `ModuleNotFoundError` already kills the executor at `src/traceml_ai/samplers/process_sampler.py:18` on both trees before any logging is involved. Rotation policy is byte-identical to base at 50 MB and 3 backups. The launcher failures in `inspect`, `compare`, and `view` do reach `logging.lastResort` rather than a file, but that is unchanged from base and each already prints a user-facing error. And mutating `role="rank"` to `role="aggregator"` at `src/traceml_ai/runtime/executor.py:60` does fail a test, so that wiring is pinned too.

abhinavsriva:
Addressed the four items:
- Preserve explicit exception causes without including unrelated user exceptions.
- Log Ray aggregator finalization failures.
- Keep user exceptions out of publisher/exporter cleanup logs.
- Strengthen regression tests for these paths and duplicate logging.

Updated the documentation. Local validation: 1,685 tests passed and the strict docs build passed.

The remaining edge cases are follow-up improvements.

PR #450, opened 2026-09-07
Add a reproducible LeRobot training regression case study

abhinavsriva:
## Summary

Adds an advanced example for comparing LeRobot image-dataset loading before and after the upstream fix in [PR #2408](https://github.com/huggingface/lerobot/pull/2408), related to [issue #2282] https://github.com/huggingface/lerobot/issues/2282). 

This uses the fix's short ACT workload, not the original issue's full SmolVLA run. The goal is a runnable case study showing how TraceML separates input stalls from training compute. The fix belongs to LeRobot; TraceML measures its effect.

## Changes

- A runner with pinned LeRobot revisions, dataset revision and historical dependencies; supports one A/B pair or two pairs with reversed execution order.
- The same small instrumentation patch for both revisions, including explicit forward and optimizer timing in the Accelerate-based training loop.
- A local analysis notebook for reading SQLite traces and plotting per-step input wait and step time; analysis requires no GPU.
- Setup and usage instructions, plus an entry in `examples/README.md`.

No TraceML API changes. LeRobot's model, optimizer and DataLoader configuration remain unchanged.

## Case-study results

Measured on one NVIDIA L4 using ACT, 200 steps per run, PyTorch 2.7.1/CUDA 12.8 and Accelerate 1.11.0. Two pairs were run in opposite orders.

| TraceML comparison metric | Pair 1: broken → fixed | Pair 2: broken → fixed |
| --- | --- | --- |
| GPU Step Time (comparison clock) | 1496.9 → 116.8 ms | 1542.7 → 117.2 ms |
| DataLoader fetch (CPU) | 1375.7 → 1.5 ms | 1427.5 → 1.3 ms |
| Compute | 115.9 → 110.5 ms | 110.4 → 111.1 ms |

Both comparisons changed from **INPUT-BOUND** to **COMPUTE-BOUND**, with reported step time decreasing by 92.2% and 92.4%. The improvement was predominantly in input wait, not model compute. These are measurements of this pinned workload, not a general LeRobot speedup or a model-quality claim.

Source experiment: `20260907T110120Z_pPXAdO`, comparison reports `compare/pair_1_broken_vs_fixed.txt` and `compare/pair_2_broken_vs_fixed.txt`. Raw local results are not included in this PR.

## Validation

- `bash -n` and the runner's `--help` pass.
- The GPU experiment produced both A/B comparison reports.
- `git diff --check` passes.

To reproduce the two-pair experiment from the repository root:

```bash
bash examples/advanced/lerobot_v3_image_regression/run_reproduction.sh --pairs 2
```

## Summary by CodeRabbit

- **New Examples**
  - Added an advanced example reproducing a LeRobot v3 image-dataset input regression, with broken and fixed training comparisons on CUDA.
  - Added an automated workflow for preparing environments, running configurable reproduction pairs, capturing trace data, and comparing results.
  - Added guidance for reviewing results locally through a Jupyter notebook.
- **Documentation**
  - Documented pinned source, dataset, environment, and training settings for consistent reproduction.
  - Added setup requirements for notebook-based analysis and CUDA-enabled execution.


Pendu:
Reviewed at `63020c9` against `main` (merge-base `76c623a`), head and base checked out side by side. The PR is not behind its base and has no overlapping-file drift. The only prior reviewer is CodeRabbit; I disposed of its merge-risk flag at the end rather than leaving it open.

Everything below is CPU-only static and structural checking. The published numbers are the one thing I could not confirm, since the run needs an L4 and the raw results are not in the PR. More on that at the end.

First, the parts of the design I checked and found sound, because these are what would quietly invalidate the case study:

- The one instrumentation patch really does apply to both arms. `git apply --check --unidiff-zero` succeeds against `lerobot_train.py` at both pinned SHAs, and the two revisions of that file are byte-identical at 448 lines, so the arms are instrumented the same way by construction.
- `pyproject.toml` is byte-identical across the two revisions, which is what makes the `--no-deps` switch in `select_lerobot` safe. No dependency is skipped on the fixed arm.
- The five pins are not overridden by the LeRobot install. LeRobot declares `torch>=2.2.1,<2.8.0`, `datasets>=4.0.0,<4.2.0`, `accelerate>=1.10.0,<2.0.0`, `torchcodec>=0.2.1,<0.6.0`, `torchvision>=0.21.0,<0.23.0`, and all five pinned versions sit inside those ranges.
- The upstream fix is one file, `src/lerobot/datasets/lerobot_dataset.py`, +20/-5, which matches the claim that the training loop is untouched.
- `--pairs 2` does reverse execution order. I ran the script's own pairing loop with `run_revision` stubbed: pair 1 runs broken then fixed, pair 2 runs fixed then broken, and the comparison arguments stay broken-versus-fixed in both.
- The README's reason for wrapping `forward` is correct. `update_policy` calls `policy.forward(batch)` directly inside `accelerator.autocast()`, so a `__call__` hook would not fire. There is no double counting either: the patch enables only `patch_dataloader`, `patch_backward` and `patch_h2d`, and `wrap_forward` raises if automatic forward instrumentation is active.
- No `torch.cuda.synchronize()` is needed or injected. Timing uses `torch.cuda.Event` and reads `elapsed_time` only behind a non-blocking `gpu_end.query()`, so the harness does not stall the workload it measures.
- The dataset revision pin resolves (HTTP 200), and `torchcodec-0.5+cu128` for cp310 exists on the cu128 index.

**Nothing verifies that the intended LeRobot revision is actually the one being imported.** This is the one I would fix first, because its failure mode is silent and total. `select_lerobot` is the whole switch between arms:

```bash
select_lerobot() {
  "$venv_root/bin/python" -m pip install --quiet --no-deps -e "$1"
}
```

`run_revision` calls it and then trusts it. There is no assertion anywhere in the script that `import lerobot` resolves under the intended worktree; grepping for `import lerobot` or `lerobot.__file__` returns zero hits. If that reinstall ever no-ops, from a pip resolver change, a stale `__editable__` finder left by the earlier full install, or a future build-backend change, both arms import the same tree, both runs succeed, both summaries are written, and `traceml compare` produces a clean report comparing a revision against itself. Nothing in the output would look wrong. `--quiet` also suppresses the one line that would show it.

One line after the switch closes it:

```bash
"$venv_root/bin/python" -c "import lerobot,sys; sys.exit(0 if lerobot.__file__.startswith('$worktree') else 1)" || die "wrong LeRobot tree active"
```

**The notebook picks a different clock than TraceML designates for that event, then sums across clocks.** In `analyze_results.ipynb` cell 4:

```python
"input_wait_ms":  ("_traceml_internal:dataloader_next", "cpu_ms"),
"traced_step_ms": ("_traceml_internal:step_time",       "gpu_ms"),
...
sample["total_step_ms"] = sample["input_wait_ms"] + sample["traced_step_ms"]
```

The contract in `diagnostics/step_time/policy.py` is `step_time_ms = input_wait_ms + traced_step_time_ms`, so summing the two is right in principle. The problem is which clock each term uses. `samplers/step_time_sampler.py` puts exactly these two events in `_CPU_DURATION_EVENT_NAMES` and says they "keep CPU-wall `duration_ms` for compatibility while still carrying nullable `gpu_ms` evidence for later analysis." The notebook takes the nullable evidence clock for one term and the CPU clock for the other.

That word "nullable" is why the second half matters: `event_duration` does `value = sample.get(preferred_clock)` and then `if value is None: value = sample.get("duration_ms")`, recording nothing about which clock produced the number. `utils/timing.py` only sets the GPU time when `gpu_end.query()` is true, so a step whose end event has not resolved at read time legitimately has no `gpu_ms` and silently falls back to the CPU clock. A reader cannot tell which points in the plot came from which clock.

For this workload the magnitude is probably small, since input wait happens while the GPU is idle. The fix is small too: use `duration_ms` for both terms to match the sampler's stated convention, or keep `gpu_ms` and report how many samples took the fallback.

**Both plots assert the conclusion regardless of the data.** Cell 7 hardcodes:

```python
axes[0].set_title("DataLoader starvation appears as repeated input-wait spikes")
axes[1].set_title("The same stalls dominate end-to-end step time")
```

Someone whose reproduction shows no starvation still gets a figure captioned as though it does. For an artifact whose purpose is letting a reader check a claim, the figure should describe the axes and let cell 8's markdown carry the interpretation.

**The notebook crashes on a case the comparison is designed to emit.** Cell 5 formats with `f"{metric['lhs']:.1f} ms"` and no None guard, but `reporting/compare/sections/step_time.py` deliberately returns unavailable phase metrics when the two runs have no common clock, with the message "Selected-clock phase metrics are unavailable because their clocks differ or are missing." Point the notebook at such a comparison and it raises `TypeError` instead of showing that message.

**Five dependencies are pinned and eighteen are not.** `requirements.txt` pins torch, torchcodec, torchvision, accelerate and datasets. LeRobot's own list adds, unpinned: diffusers, huggingface-hub, cmake, einops, opencv-python-headless, av, jsonlines, packaging, pynput, pyserial, wandb, draccus, gymnasium, rerun-sdk, deepdiff, imageio, termcolor, setuptools. Two matter specifically here: `huggingface-hub` is the dataset download and cache path, which is the machinery this regression is about, and `wandb` is unpinned while the runner turns it on. `README.md` says "Exact dependency versions live in `requirements.txt`", which overstates what is actually pinned.

The pins hold today, as noted above. The gap is that nothing enforces them: a requirements file is not a constraints file, so the guarantee comes from LeRobot's ranges happening to be wide, not from the harness. Passing `-c "$requirements_file"` to both editable installs makes it structural.

`record_environment` then writes only those same five versions, so a reader who gets a different result cannot find out what differed. A `pip freeze > "$experiment_root/pip-freeze.txt"` next to the existing capture costs one line. It also records `git rev-parse HEAD` for TraceML without any dirty check, so a locally modified working tree reports a clean SHA.

Related, `prepare_environment` runs `pip install --upgrade pip setuptools wheel` on every invocation; the `if` above it guards only venv creation. The resolver itself therefore changes between runs, which sits awkwardly against the README's "Later runs reuse both."

**The README and the PR description say opposite things.** `README.md` says "This is a reproduction, not a published benchmark. It intentionally contains no performance claim." The PR description publishes a table with a 92.2% and 92.4% improvement. My concern is not the wording but what happens next: numbers travel and disclaimers do not, so if that table gets quoted the "no performance claim" sentence stays behind in the repository. Worth picking one, and if the numbers are meant to be citable, stating the hardware and the tolerance.

**For an input-bound result, the host CPU is unrecorded and the worker count is unpinned.** `record_environment` captures the GPU with `nvidia-smi --query-gpu=name,memory.total,driver_version` and nothing at all about the CPU: no core count, no model, no `nproc`. The README also says the run uses "LeRobot defaults for batch size, workers, and seed", so dataloader parallelism is whatever the host and LeRobot negotiate.

That matters more here than it would for a compute-bound study. The headline is that the regression was input wait, which is CPU-side dataloading work, so the number of cores and workers feeding the GPU is arguably the most load-bearing environmental variable in the whole experiment, and it is the one variable not captured. Two readers on the same L4 with different vCPU counts can both reproduce faithfully and disagree, with nothing in `environment.txt` to explain why. One `nproc` and the resolved worker count in `record_environment` would fix it.

Smaller things:

- No cleanup anywhere: no `trap`, no `git worktree remove`. Each run adds two worktrees to the shared clone, 9.0 MB each, and leaves two registrations in `git worktree list` that outlive the deleted experiment directory. Every `die` site and every Ctrl-C leaks the same way. A `git worktree prune` at startup would bound it.
- If the first `git clone` is interrupted, the directory exists without `.git`, so later runs skip the clone and die permanently at the `is not a Git checkout` check without telling the user to delete it.
- The cheap GPU check cannot catch a bad `CUDA_VISIBLE_DEVICES`, because `nvidia-smi` ignores that variable and reports all GPUs. The real check is `torch.cuda.device_count() == 1`, which only runs after the clone and the whole venv build. A torch-free check earlier would fail in seconds instead of minutes.
- `need python3.10` confirms a binary of that name exists, not that it is Python 3.10.
- `select_lerobot`'s pip failure exits bare with its output suppressed by `--quiet`; it deserves a `|| die`.
- The notebook defaults to `PAIR = 1`, but the published evidence is two pairs, and the README never mentions rerunning with `PAIR = 2`.
- `analysis-requirements.txt` uses ranges rather than pins. Low risk given the notebook only uses stable matplotlib calls, but inconsistent with the pinning discipline elsewhere.
- `--policy.repo_id=$DATASET_ID` points the policy repo id at the dataset repo. Inert while `push_to_hub=false`, but confusing.
- The runner enables W&B in offline mode. It applies to both arms so it cannot bias the comparison, but it does sit inside the absolute millisecond numbers being published.
- The patch's last hunk header reads `@@ -329,12 +341,13 @@` where the running offset from the three preceding insertions gives 340. `git apply` locates hunks by the old side so this changes nothing, and I confirmed the patch applies to both SHAs, but regenerating it from a real `git diff` would remove the discrepancy.

On CodeRabbit's merge-risk flag, which called the pinned torch a major security issue: I do not think it blocks anything. `GHSA-rrmf-rvhw-rf47` and `GHSA-qfhq-4f3w-5fph` are both rated LOW and cover `torch.jit.script` and `torch.lstm_cell`. Neither appears anywhere in the ACT policy, the training script, or the dataset module at the pinned revision. Pinning a historical torch is the point of a regression reproduction, and bumping it would break the thing being reproduced. A sentence in the README saying the pin is deliberate and historical would answer this for the next reader.

Last, the gap I could not close. The numbers are the product of this PR and right now can only be checked by renting an L4, because the raw results are not included and the notebook cannot run without them. Shipping one sanitized reference bundle, the two `final_summary.json` files, the compare `.json`, `environment.txt` and a pip freeze, would let anyone re-derive the table without a GPU and would give the notebook something to run against. Note `logs/` is gitignored, so the bundle needs a different home. Alternatively I can run the reproduction on our own GPU instance and confirm the table independently; that would also exercise the pinning gaps above, since a fresh install today is exactly the unpinned case.

Pendu:
I ran the reproduction on a clean box to check the table independently. Short version: the claim holds, the percentages are host-specific, and one supporting statement does not survive a cold first run.

Setup: `g6.4xlarge` in `eu-central-1`, NVIDIA L4 23034 MiB, driver 595.91.07, 16 vCPU AMD EPYC 7R13, stock Ubuntu 22.04 Deep Learning AMI, TraceML at `63020c9`, `--pairs 2`, nothing else on the box.

**The headline reproduces.** Both pairs moved INPUT-BOUND to COMPUTE-BOUND, and input wait collapsed by 99.9% in both:

| | published | measured here |
| --- | --- | --- |
| pair 1 step time | 1496.9 to 116.8 ms | 1341.6 to 148.4 ms |
| pair 2 step time | 1542.7 to 117.2 ms | 1184.3 to 148.1 ms |
| pair 1 input wait | 1375.7 to 1.5 ms | 1093.6 to 1.34 ms |
| pair 2 input wait | 1427.5 to 1.3 ms | 1034.4 to 1.46 ms |
| improvement | 92.2% and 92.4% | 88.9% and 87.5% |

So the mechanism and the direction are solid. The percentage is not portable: about four to five points lower on my host, from the same GPU model at the same pinned revisions.

**The "compute was unchanged" claim does not hold on a cold box, for pair 1 only.** Taking compute as step time minus input wait:

| | published | measured here |
| --- | --- | --- |
| pair 1 compute | 115.9 to 110.5 ms | 247.9 to 147.1 ms |
| pair 2 compute | 110.4 to 111.1 ms | 149.9 to 146.7 ms |

My pair 1 shows forward dropping 137.9 to 68.1 ms and backward 87.8 to 56.7 ms, which reads as a large compute improvement from a dataset fix, and it is not one. It is warm-up: CUDA context, cuDNN autotune and page cache, all paid by whichever arm runs first. Pair 2, with everything warm, is flat at forward 71.8 to 67.5 and backward 55.9 to 57.0, which matches your story exactly.

Your published numbers are flat in both pairs, which suggests your box was already warm when pair 1 started. On a cold machine the reversed-order control catches this, but only if the reader looks at pair 2. Worth either discarding a few warm-up steps before measuring, reporting pair 2 as the headline, or saying plainly that pair 1's compute delta is warm-up rather than the fix.

Also worth noting: my compute sits around 147 ms against your 115 ms on the same L4 model, a 30% gap that has nothing to do with the change under test.

**Two setup problems a clean box hits and a dev box does not.**

First, `python3.10 -m venv` fails on a stock Ubuntu 22.04 image because the `python3.10-venv` package is not installed. The harness dies with `could not create the Python 3.10 environment`. The README's prerequisites list Linux x86-64, Python 3.10, a GPU and network, so this is invisible until someone runs it somewhere clean. Adding `python3.10-venv` to the prerequisites, or having `check_host` probe `python3.10 -m venv` rather than just the binary, would fix it. Note that `python3.10 -c "import venv"` succeeds on that same box, so an import check is not a valid probe here.

Second, that failure leaves `venv/bin/python` behind, and `prepare_environment` guards venv creation on exactly `[[ ! -x "$venv_root/bin/python" ]]`. So the retry skips creation and runs on the broken partial venv, turning a clear error into a confusing downstream one. Same shape as the interrupted-clone case: the first failure poisons every later run silently.

**On the pinning.** The pinned set resolves cleanly today, so nothing is broken. But the run installed 142 packages and 5 of them are pinned. For the record, here is what the unpinned ones resolved to on this box, which would make a reasonable starting lockfile:

```
av==15.1.0                   diffusers==0.35.2        draccus==0.10.0
deepdiff==8.6.2              einops==0.8.2            gymnasium==1.3.0
huggingface-hub==0.35.3      ImageIO==2.37.4          imageio-ffmpeg==0.6.0
jsonlines==4.0.0             numpy==2.2.6             opencv-python-headless==4.12.0.88
packaging==25.0              pillow==12.3.0           pynput==1.8.2
pyserial==3.5                rerun-sdk==0.26.2        termcolor==3.3.0
wandb==0.21.4
```

Everything else in the harness behaved as documented: both worktrees built, the patch applied to both revisions, all four runs produced summaries, and both comparison reports were written.

abhinavsriva:
Thanks for the detailed reviews and the independent clean-box reproduction.

The main result reproduced: both pairs changed from **INPUT-BOUND** to **COMPUTE-BOUND**, with input wait falling by approximately 99.9%.

### Addressed

- Verify that each run imports the intended LeRobot worktree.
- Use explicit `cpu_ms` for DataLoader and step wall time, and `gpu_ms` for compute phases. The notebook no longer reads legacy `duration_ms`.
- Enforce the five regression-sensitive pins and record the complete resolved package list, CPU information, and workload settings.
- Detect missing `python3.10-venv` and incomplete reusable environments early.
- Clarify that exact percentages are host-specific and that the first pair may contain cold-start effects.

### Intentionally unchanged

- No complete 142-package lockfile: the clean-box run reproduced the diagnosis with freshly resolved dependencies, and all resolved versions are now recorded.
- No automatic warm-up trimming: that would introduce a new measurement policy. The reversed-order pair exposes order effects directly.
- PyTorch 2.7.1 remains pinned because it is part of the historical reproduction stack; changing it would change the case being reproduced.
- `duration_ms` remains in TraceML’s storage schema only for backward compatibility. Removing it is a separate schema migration outside this case-study PR.

The implementation remains scoped to one reproducible example with no TraceML API changes.

Issue #451, opened 2026-09-07
Make the HF integration callback-only and fix training-input measurement


abhinavsriva:
## Goal

Provide one clear, reliable Hugging Face integration:

- Initialize TraceML with `traceml_ai.integrations.huggingface.init()`.
- Use standard `transformers.Trainer` with `TraceMLTrainerCallback`.
- Preserve HF’s training behavior while measuring input loading, H2D, forward, backward, and optimizer work correctly.

Modern HF Trainer already uses Accelerate internally. Users should not need to create an `Accelerator` themselves. Standalone Accelerate loops remain a separate documented recipe using `trace_step`, without requiring an HF callback.

## Why this is needed

`TraceMLTrainer` currently only installs `TraceMLTrainerCallback`, so maintaining both public APIs adds no measurement capability.

There are also correctness gaps:

- HF fetches accumulation batches before `on_step_begin`; Accelerate performs H2D during that fetch, outside TraceML’s current H2D window.
- The global DataLoader patch can attribute evaluation fetches to the next training step.
- Reported step time omits the pre-step transfer phase.
- Exceptions can leave tracing active, and later cleanup can count an aborted step as completed.
- Observing one H2D event does not establish complete batch-transfer coverage.

### 1. Remove the redundant HF Trainer wrapper

- Remove `TraceMLTrainer` and its exports.
- Keep `init()` and `TraceMLTrainerCallback` as the public HF integration.
- Migrate all maintained HF examples, tests, notebooks, and documentation that use or recommend the wrapper.
- Replace wrapper-specific tests with callback tests where they protect useful behavior.
- Include a breaking-change migration note: use standard `Trainer`, initialize TraceML, and register the callback. Replace `traceml_enabled=False` with conditional callback registration.
- Remove the outdated claim that the wrapper provides stronger exception handling.

Known cleanup locations include:

- `src/traceml_ai/integrations/huggingface.py`
- `examples/integrations/huggingface_trainer_minimal.py`
- `examples/advanced/huggingface_vision_vit.py`
- `tests/integrations/test_hf_trainer.py`
- `tests/integrations/test_telemetry_conformance.py`
- HF integration, public API, and architecture documentation

Review other HF references and update only those affected. Keep examples/notebooks already using the correct callback API unless their measurement claims need correction.

### 2. Capture training-input work at the correct boundary

- Capture real H2D performed by the prepared training loader before the callback’s compute window.
[...]

PR #452, opened 2026-09-08
Dev/abhinav/issue451-a

abhinavsriva:
## Goal

Make the Hugging Face integration callback-only:

1. Call `traceml_ai.integrations.huggingface.init()`.
2. Use the standard `transformers.Trainer`.
3. Register `TraceMLTrainerCallback`.

`TraceMLTrainer` only installed the callback, so maintaining both APIs added unnecessary complexity.

Part of #451.

## Changes

- Removed `TraceMLTrainer` and its export.
- Migrated maintained examples and tests to standard `Trainer`.
- Added migration guidance for `TraceMLTrainer` and `traceml_enabled`.
- Clarified that one TraceML step represents one HF accumulation/update boundary.
- Added focused tests for accumulation, partial final groups, skipped optimizer updates and optional callback registration.
- Updated affected documentation and the changelog.
- Updated tests for the StepCapture APIs in `version_0.3.7`.

## Migration

Replace:

```python
trainer = TraceMLTrainer(...)
```

with:

```python
traceml_hf.init()

trainer = Trainer(
    ...,
    callbacks=[traceml_hf.TraceMLTrainerCallback()],
)
```

## Within scope

- Callback-only HF API.
- Wrapper migration.
- Current HF step-boundary documentation.
- Affected examples and tests.

## Outside scope

- Pre-callback input and H2D measurement.
- HF failure and retry cleanup.
- Evaluation and prediction input isolation.
- Timing and reporting changes.
- StepCapture, sampler or queue redesign.
- Changes to standalone Accelerate, Lightning or Ray.

These remaining HF improvements will follow separately under #451.

## Validation

- All GitHub CI checks passed.
- `1,344 passed, 1 skipped` in the integration suite.
- Strict documentation build passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- **Removed:** `traceml_ai.integrations.huggingface.TraceMLTrainer` and its
+  `traceml_enabled` argument. The wrapper only installed the callback. Call
+  `traceml_ai.integrations.huggingface.init()` and register
+  `TraceMLTrainerCallback()` with standard `transformers.Trainer` instead.
+  For optional tracing, register the callback conditionally. See the
+  [HF migration instructions](docs/user_guide/integrations/huggingface.md#migration)
+  for the replacement setup.
 - PyTorch Lightning: `TraceMLCallback` now opens the traced step when
   Lightning moves the batch to the device, so the H2D transfer is inside
   Traced Step Time and Step Time on GPU runs; fetches of validation,
--- a/docs/developer_guide/architecture.md
+++ b/docs/developer_guide/architecture.md
@@ -42,7 +42,7 @@ Samplers maintain an incremental append counter per rank per table. The sender s
 The `src/traceml/` package is a deprecated compatibility alias for older import
 paths. New implementation work should go under `src/traceml_ai/`.
 
-For the user-facing API surface (`trace_step`, `TraceMLTrainer`, `TraceMLCallback`, CLI usage), see the [Public API](../user_guide/public-api.md). The source tree above is the canonical reference for internals — start from the entry points and follow the imports.
+For the user-facing API surface (`trace_step`, `TraceMLTrainerCallback`, `TraceMLCallback`, CLI usage), see the [Public API](../user_guide/public-api.md). The source tree above is the canonical reference for internals — start from the entry points and follow the imports.
 
 Contributors changing Step Time should begin with the
 [Step Time pipeline contract](step-time-pipeline-contract.md), which maps its
--- a/docs/developer_guide/step-time-pipeline-contract.md
+++ b/docs/developer_guide/step-time-pipeline-contract.md
@@ -237,6 +237,68 @@ SQLite normalization. `_traceml_internal:step_time` is intentionally stable
 raw telemetry and storage vocabulary below that normalization boundary; it is
 not a public metric name or presentation label.
 
+## Hugging Face steps
+
+For normally completed HF training, one TraceML step contains the microbatches
+used for one optimizer update attempt. The callback opens `trace_step` at
+`on_step_begin` and closes it at `on_step_end`. Accumulating microbatches emit
+`on_substep_end`, which does not advance TraceML's counter.
+
+For example, ten microbatches with `gradient_accumulation_steps=4` produce
+three steps containing four, four, and two microbatches. TraceML IDs are local
+to the process; their increments match HF's for recorded, completed groups,
+but their absolute values can differ after checkpoint resume.
+
+### Timing and memory
+
+All timing events for a group are flushed together as one `StepTimeBatch`.
+`StepTimeSampler` sums repeated forward and backward events within that batch,
+keeping CPU and GPU durations separate. It does not merge separate batches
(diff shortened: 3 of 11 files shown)


Pendu:
Reviewed at `7333307` against `version_0.3.7` (merge-base `6e67e39`, which already carries #456, #457 and #458), head and base checked out side by side. No prior human review, and the PR is not behind its base. Reviewing this one as release cargo since you said it is the last before the merge to main.

The removal itself is done properly, and I checked the parts that usually go wrong. Grepping the whole tree for the removed name turns up exactly three hits and all three are deliberate: the CHANGELOG entry, the migration section in the HF doc, and the module docstring. Nothing dangles in src, tests, examples, notebooks, mkdocs nav or the README. The CHANGELOG marks it `Removed`, names the `traceml_enabled` argument alongside the class, gives the replacement and links the migration anchor, which is the part a user actually meets. `traceml_enabled` is gone everywhere except those two notes. The suite is green at 1777 passed and 1 skipped, `mkdocs build --strict` is clean, and every linter passes at the pinned versions.

Two things worth doing before this ships, both small.

**The dedup guard went away with the wrapper and nothing replaced it.** The PR description says the wrapper only installed the callback. It did one more thing:

```python
existing = getattr(self.callback_handler, "callbacks", [])
if any(isinstance(cb, TraceMLTrainerCallback) for cb in existing):
    return
self.add_callback(TraceMLTrainerCallback())
```

The callback has no equivalent check, and `test_hf_trainer_wrapper_dedups_user_supplied_callback` was removed along with the wrapper, so the invariant is now neither enforced nor tested. Driving the hook order with one and then two registered callbacks:

```
1 callback  -> 1 batch published, step counter advanced by 1
2 callbacks -> 1 batch published, step counter advanced by 2
```

So a double registration does not produce double telemetry, it produces half. One batch per real step, while the counter runs ahead at twice the rate, which makes published step numbers go sparse against a counter that keeps climbing. Nothing warns. Granted the old guard existed because the wrapper made accidental doubling easy and that path is gone, so the exposure is smaller now, but the cost of a check in `on_train_begin` is about four lines.

**The upgrade error says nothing about what replaced it.** A user on v0.3.x whose script still has `from traceml_ai.integrations.huggingface import TraceMLTrainer` gets a bare `ImportError`, and the attribute form gets a bare `AttributeError`. The CHANGELOG and the migration doc are good, but neither is in front of the person whose training script just died. There is already a module-level `__getattr__` in `src/traceml_ai/__init__.py:50` used for lazy API resolution, so the pattern is in-house: recognising the old name and raising with the one-line replacement would catch both the import and the attribute path and cost very little.

Three things I want to credit specifically, because they are the kind of work that is easy to miss in a migration PR.

The `runtime/state.py` docstring change is a real correction, not a reword, and it is more precise than it looks. At base that docstring said TraceML treats gradient-accumulation micro-batches as distinct traceable steps, while the callback's own docstring in the same repo said the opposite. The code agrees with the callback for HF: the hooks are `on_step_begin` and `on_step_end` with no `on_substep_end`, so one traced step has always spanned a whole accumulation group.

Worth knowing when that drifted. The phrase came in with `2dd0b28b` (V0.2.13), when the wrapper really did override `training_step` and bracket each micro-batch. `956a53fb`, the callback-based HF integration from early June, deleted that override, and the sentence has been false for HF ever since without anything catching it. It stayed true for Lightning the whole time, since the Lightning callback opens its step region in `on_train_batch_start`, one per training batch. So the new wording is not just a fix, it is the first version that is correct for both integrations at once.

The test rewrite does not cost coverage, which is the main risk in a change of that size. Tests go from 8 to 7 but assertions go from 19 to 23, and all three removed tests were wrapper-specific. The one with a name that does not look wrapper-specific, `test_hf_trainer_integration`, turns out to be the wrapper's end-to-end twin of `test_hf_trainer_callback_integration`, which exists at both base and head, so the surviving behaviour stays covered.

And the remaining integration test got stronger in a way the diff does not advertise. At base it never called `init()`, so none of the dataloader, H2D, forward or backward patches were installed and every phase timing was dark while the test still passed on step memory alone. In other words it was not exercising the path the docs tell users to follow. Head adds the `init()` call and tightens the step assertion from `>= max_steps` to `== max_steps`.

One note on the new tests rather than a complaint about them. All four pass against base as well, because the callback logic itself is unchanged here, so they are regression pins rather than gates on new behaviour. That is the right shape for a removal PR. The weakest of them is the optional-callback test: its untraced arm is not a true negative control, since pytest runs the file in one process and the process-wide patches installed by an earlier test's `init()` are still active, so it demonstrates callback absence rather than TraceML absence.

On scope, I noted that HF failure and retry cleanup is on your out-of-scope list and tracked under #451, so I have not re-raised the failed-step publishing behaviour here.

abhinavsriva:
Thanks.

- The duplicate-callback issue is valid and is addressed in #459 through per-run callback ownership, with a regression test checking step counts and timing/memory publication. That PR will merge before release.
- We are keeping the intentional API removal straightforward, with migration guidance in the changelog and HF guide, without adding a legacy-name lookup handler.
- Agreed on the optional-callback test: it validates conditional callback registration and unchanged training results, not a fresh process with  all instrumentation absent.

Keeping this PR focused on removal and migration; lifecycle handling staysin #459.

PR #454, opened 2026-09-08
Centralize timing event types and queue ownership

abhinavsriva:
## Goal

Make the training-to-sampler boundary explicit as part 1 of the step-event
cleanup RFC targeting version_0.3.7.

Previously, utils/timing.py owned measurement, pending events, event types,
and the queue. The sampler accessed that queue through a getter and a
generic drain helper.

This change gives the event types and queue one owner while preserving
the existing measurement and sampling behavior.

## Current flow

Timed regions and optimizer hooks
  -> utils/timing.py: measure and buffer events
  -> existing step boundary: assign the step number and flush
  -> instrumentation/step_events.py: publish the timing batch
  -> StepTimeSampler: drain, resolve CUDA events, aggregate, and persist

## In scope

- Move TimeScope, TimeEvent, StepTimeBatch, and the timing queue into
  instrumentation/step_events.py.
- Replace direct queue access with publish_step_time_batch() and
  drain_step_time_batches().
- Update affected producers, sampler imports, tests, and development scripts.
- Remove the unused append-to-deque queue helper.
- Document the ownership boundary and add a changelog entry.
- Add three focused regression tests for batch delivery, CUDA FIFO
  processing, and queue overflow.

## Behavior preserved

- Measurement boundaries, event names, clocks, and step numbering.
- Grouping performed by the existing caller's flush boundary.
- Repeated-event aggregation and persisted output.
- Queue capacity of 2,048 batches and dropping the incoming batch when full.
- FIFO processing: an unresolved earlier CUDA batch holds back later batches.
- Nonblocking CUDA readiness checks and existing event-pool reuse.
- Draining already-published batches after recording stops.
- Batch transfer by reference, without copying event objects.
- Existing lazy imports for operation without PyTorch.

After publication, producers must not modify batch membership or step
identity. The sampler resolves the batch's CUDA events.

These are internal interfaces; the documented user APIs do not change.

## Out of scope

- Moving memory events or removing model_id: part 2.
- Introducing StepCapture and completion/abort handling: part 3.
- Fixing existing failed-step attribution or session-buffer lifecycle.
- Changing HF/Accelerate input-fetch or H2D measurement boundaries.
- Changing integration-specific gradient-accumulation or step semantics.
- Synchronizing timing and memory samplers through a combined queue or barrier.
- Changing overflow, shutdown, or CUDA error-handling policies.
- Implementing persistence for global timing events.

This PR preserves existing caller-defined step boundaries. It does not
establish one universal optimizer-step definition across integrations.

## Validation

- Fresh-checkout focused suite: 191 passed, 2 skipped, 1 watch smoke test excluded.
- Existing SDK, integration, sampler, recording-state, and torch-free
  import checks included.
- Additional probes covered empty flushes, exception propagation, import
  order, and concurrent publication/draining of 500 batches.
- Executable-code comparison confirmed unchanged measurement, CUDA
  resolution, publication, and sampler calculation logic.
- Strict documentation build and formatting passed.
- CPU overhead probe showed approximately 0.4% difference, within observed
  run variation.

CUDA FIFO behavior was verified with controlled event doubles. Real-GPU
timing accuracy and distributed training were not validated in this review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Timing event types and queue access now live in
+  `instrumentation/step_events.py`. Producers and the timing sampler share an
+  explicit handoff; measurement boundaries, CUDA FIFO processing, and stored
+  output are unchanged.
 - **Breaking:** TraceML implementation errors now have one structured owner:
   `rank_<global_rank>/traceml_errors.log`,
   `aggregator/traceml_errors.log`, or
--- a/docs/developer_guide/step-time-pipeline-contract.md
+++ b/docs/developer_guide/step-time-pipeline-contract.md
@@ -10,7 +10,35 @@ For the public final-summary shape, see
 For user-facing timing definitions, see the
 [Step Time glossary](../user_guide/reading-output.md#step-time-glossary).
 
-## Current flow
+## Training-to-sampler handoff
+
+`instrumentation/step_events.py` owns `TimeEvent`, `StepTimeBatch`, and the
+private timing queue. `utils/timing.py` measures regions and buffers events
+until the existing step boundary assigns their step number and publishes one
+batch through `publish_step_time_batch()`.
+
+```text
+timed regions / optimizer hooks
+  -> utils/timing.py: pending events and step flush
+  -> instrumentation/step_events.py: timing batch queue
+  -> StepTimeSampler: pending FIFO, CUDA resolution, aggregation, storage
+```
+
+The sampler calls `drain_step_time_batches()` to take available batches without
+blocking. Publication and reading transfer references, not copies of events.
+After publication, the producer must not change a batch's membership or step
+identity. Only the sampler resolves its CUDA events.
+
+The queue holds up to 2,048 batches and retains the existing drop-on-full
+behavior. The sampler keeps its existing pending FIFO: an unresolved earlier
+CUDA batch holds back later batches. Reading continues after recording stops
+so already-published measurements can drain. No GPU synchronization is added.
+
+This first cleanup changes the timing handoff only. Memory still uses its
+existing queue, and step completion, failure handling, and input-fetch
+boundaries are unchanged. Global timing is currently not persisted.
+
+## Analysis and presentation flow
 
 ```mermaid
 flowchart LR
--- a/src/dev/repro/hf_accelerate_h2d_window.py
+++ b/src/dev/repro/hf_accelerate_h2d_window.py
@@ -56,8 +56,8 @@ def _positive_control_ms(num_bytes_mb: int = 64) -> float | None:
     """Move a tensor INSIDE trace_step; TraceML should capture this one."""
     import torch
 
+    from traceml_ai.instrumentation.step_events import drain_step_time_batches
     from traceml_ai.sdk.instrumentation import trace_step
-    from traceml_ai.utils.timing import get_step_time_queue
 
     model = torch.nn.Linear(4, 4).cuda()
     host = torch.empty(num_bytes_mb * 1024 * 1024 // 4, dtype=torch.float32)
@@ -66,9 +66,7 @@ def _positive_control_ms(num_bytes_mb: int = 64) -> float | None:
         torch.cuda.synchronize()
 
     captured = None
-    queue = get_step_time_queue()
-    while not queue.empty():
-        batch = queue.get_nowait()
+    for batch in drain_step_time_batches():
         for evt in getattr(batch, "events", []):
             if evt.name == "_traceml_internal:h2d_time":
                 captured = float(getattr(evt, "gpu_ms", 0.0) or 0.0)
@@ -87,8 +85,8 @@ def _traceml_reported_h2d(grad_accum: int = 2) -> float | None:
         TrainingArguments,
     )
 
+    from traceml_ai.instrumentation.step_events import drain_step_time_batches
     from traceml_ai.integrations.huggingface import TraceMLTrainerCallback
-    from traceml_ai.utils.timing import get_step_time_queue
 
     class _DS(torch.utils.data.Dataset):
         def __len__(self):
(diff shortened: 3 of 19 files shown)


Pendu:
Reviewed at `5defbb0` against `version_0.3.7` (merge-base `57dc66f`), head and base checked out side by side. No prior human review, and the PR is not behind its base.

You said this is a pure move with no logic change. I tested that rather than taking it, and it holds. Two independent checks:

I compared every moved symbol as an AST with docstrings stripped and `Optional[X]` normalised against `X | None`, since the new module adds `from __future__ import annotations`. `TimeScope`, `TimeEvent` and `StepTimeBatch` come out logic-identical, and `_enqueue_step_batch` to `publish_step_time_batch` is an identical body under a new name, down to the stderr string.

Then the stronger check. I retargeted your three new tests onto the base API (`timing._STEP_TIME_QUEUE`, `timing._enqueue_step_batch`, `drain_queue_nowait(get_step_time_queue())`, `timing.return_cuda_event`) and ran them against `57dc66f`: **8 passed**. They assert the same behaviour before and after the move, which is exactly the right shape for guarding a refactor and is the best evidence the claim is true.

Two things in the diff are not copy-paste, and are worth reviewing as new work rather than waved through with the rest: `drain_step_time_batches` is new code, not a moved function. It replaces the old get-the-queue-then-drain-it pattern and is behaviourally equivalent to `drain_queue_nowait`, but it is new. And the 116 test lines plus the 30-line contract section are additions. Both are good additions; they just are not what "copy paste" describes.

Also verified clean, since these are the ways a move like this usually goes wrong: there is exactly one `_STEP_TIME_QUEUE` object shared by producer and consumer, so no split-queue where events are enqueued in one and drained from another. No import cycle at runtime. Zero remaining references to `get_step_time_queue`, `append_queue_nowait_to_deque` or `_enqueue_step_batch` anywhere in src, tests, docs or examples. `drain_queue_nowait` is correctly retained for the memory sampler rather than over-deleted. All linters pass at the pinned versions, and every behavioural claim in the new contract section checks out against base code.

**The first commit does not import.** `732656c` repoints nine files at `traceml_ai.instrumentation.step_events` and deletes the definitions from `utils/timing.py`, but never adds `step_events.py`. `git show 732656c:src/traceml_ai/instrumentation/step_events.py` returns nothing while `timing.py` at that commit already imports from it twice, so anything importing `utils.timing` raises `ModuleNotFoundError`. `cbb572b` ("step time missng") is the missing file being added. The head state is fine, so this only matters for history: a squash-merge makes it moot, a merge that preserves the commits leaves a broken revision that breaks `git bisect` and per-commit CI across that range.

**The move creates a new dependency direction, and the package docstring no longer matches.** `utils/timing.py` now imports from `instrumentation/step_events.py`. At base no module under `utils/` imported `instrumentation/` at all, while `instrumentation/` already imported `utils/` in five modules, now six. So the two packages point at each other. It is acyclic today and I confirmed that by importing both, but `instrumentation/__init__.py` still says the package "owns the hook installers and monkey patches used by the SDK", and `step_events.py` is a data contract plus a queue, which is neither. In fairness `h2d.py` already stretched that sentence before this PR. Placement is your call and the contract doc argues it reasonably; the minimum is updating that docstring so the ownership map matches the code, since there is no CI guard on this axis.

Some dead code this PR is well placed to sweep, since it is the ownership-cleanup PR. `_enqueue_global` has `pass` as its body with the real `put_nowait` commented out behind a TODO, so `_GLOBAL_TIME_QUEUE` has neither a producer nor a consumer, `get_global_time_queue()` has no callers, and the `except Full:` around a bare `pass` is unreachable. All of this is identical at base, so it is not something you introduced. Worth noting that your docstring change here is a real correction: base said "GLOBAL events are enqueued immediately", which was never true, and "Global timing is currently not persisted" is the first accurate statement of what that code does. I misread this myself at first from the call site before reading the body.

Smaller things:

- `drain_step_time_batches` ends its loop on any non-`Empty` exception with no stderr line, under a `noqa: BLE001`. The convention elsewhere is to report to stderr with a `[TraceML]` prefix. If `get_nowait` ever raises persistently the consumer goes quiet while the producer keeps filling to 2048 and then starts printing drop messages, so the two halves would disagree about what is wrong. It is inherited from `drain_queue_nowait`, but the pattern now lives in two places. One print before the `break` closes it.
- `utils/timing.py` imports `TimeEvent`, `StepTimeBatch` and `TimeScope` at module level, so `from traceml_ai.utils.timing import TimeEvent` still resolves, and to the same object rather than a copy. That is a working compatibility surface for anyone importing the old path, but it exists by accident of the import rather than by intent. Worth a line in the module docstring if you want to keep it, or an `__all__` if you do not.

One thing that is not a finding but you should know about. `tests/runtime/test_executor.py::test_executor_user_failure_stays_native_and_out_of_internal_logs[False]` fails for me on this branch. It fails identically on the base, and CI is green on `57dc66f`, so it is neither yours nor a red branch. The cause is local timing: the parametrised case runs the executor with TraceML enabled, which takes about 38 seconds on my machine against the test's 30 second subprocess timeout. Worth raising the timeout, because a 30 second budget on a path that legitimately takes tens of seconds will eventually flake in CI too rather than only on slower laptops.

abhinavsriva:
Thanks for the detailed review.

Addressed in this PR:

- Updated the `instrumentation` package documentation to include event contracts and queue ownership.
- Kept the moved annotations as `Optional[X]`, consistent with the original implementation and repository style.
- Confirmed there is no runtime circular import.
- The PR will be squash-merged so the incomplete intermediate commit is not retained.

The remaining points are intentionally outside this PR:

- The unused GLOBAL timing placeholder is pre-existing and belongs to separate global-timing cleanup.
- Unexpected queue-drain error reporting should be changed consistently for both timing and memory drains, rather than only here.
- The old timing-module imports are internal implementation details and are not being established as a compatibility API.
- The executor timeout reproduces on the base branch and should be handled as a separate test-flakiness issue.

This keeps this PR limited to the behavior-preserving timing ownership move.

Issue #453, opened 2026-09-08
Clean up step-event ownership and prevent unfinished steps from being published


abhinavsriva:
Target branch: `version_0.3.7`

Implement this as one PR with three sequential parts. HF-specific lifecycle and input-measurement changes remain follow-up work.

## Goal

Create a small, shared step-recording path for timing and memory measurements while preserving the existing sampling pipeline.

The implementation should:

- Keep successful-run measurements, step boundaries, and sampler calculations unchanged.
- Give pending measurements one clear owner.
- Remove unnecessary model identity from memory telemetry.
- Publish a completed step once and discard unfinished work.
- Preserve CUDA FIFO processing and approximately the same collection overhead.

This is a focused cleanup and a correction to failed-step handling, not a redesign of the samplers or transport.

## What works today

Timing events already travel together as a `StepTimeBatch`. Repeated forward/backward calls within that batch are aggregated by the timing sampler.

Memory snapshots carry the same step number through a separate queue.

```text
Timing producers → pending timing buffer ─┐
                                         ├─ flush step N
Memory tracker → pending memory buffer ──┘
                                              │
                         ┌────────────────────┴─────────────────────┐
                         ▼                                          ▼
                Timing batch [N]                           Memory event [N]
                         ▼                                          ▼
                   Timing queue                               Memory queue
                         ▼                                          ▼
              FIFO CUDA resolution                         Memory sampler
                         ▼                                          ▼
               Timing aggregation                         Existing storage
                         ▼
                  Existing storage
```

Timing and memory do not need to be consumed simultaneously. Memory may reach step 12 while timing is still resolving step 10. Their step identifiers remain intact.

The timing sampler must continue waiting for an unresolved earlier batch before processing later batches. No cross-sampler synchronization barrier is needed.

## What needs fixing

Event definitions, pending buffers, queue ownership, and flushing are spread across `timing.py`, `step_memory.py`, and `flush_buffers.py`. The samplers also access their queues differently.

Memory includes `model_id`, although its measurements describe the process’s PyTorch allocator on a device. The SQLite projection already ignores this field.

There is also a concrete finalization bug:

1. The current step counter is 10.
2. Training raises inside the next `trace_step`.
3. The counter remains 10.
4. The `finally` block still flushes unfinished timing and memory under step 10.

The correction belongs on the producer side. Queue ordering cannot repair an incorrectly labelled event.

## Intended structure

Introduce one small shared module, such as:

[...]

PR #455, opened 2026-09-08
Simplify memory ownership and move its handoff into step_events

abhinavsriva:
## Goal

Implement part 2 of #453, building on the timing handoff from part 1.
Target: version_0.3.7.

Move memory event definitions and queue ownership into the shared step-events
module. Remove unnecessary model identity while preserving measurement,
sampling, and storage behavior for the current one-device-per-process path.

## What changes

Previously, step_memory.py owned the event definition, memory queue, and
a pending dictionary keyed by model_id.

Now:

- instrumentation/step_events.py owns StepMemoryEvent and the private memory
  queue, with typed publication and drain functions.
- step_memory.py measures memory and keeps one pending snapshot per process.
- record() replaces the pending snapshot.
- flush(step) clears the pending slot, assigns the step number, and publishes
  the event.
- The queue retains previously published snapshots until the sampler drains
  them. Replacing the pending snapshot does not overwrite queued steps.
- device remains snapshot metadata. There is no model or device grouping key.

## Data flow

StepMemoryTracker.record()
  -> one process-local pending snapshot
  -> existing flush boundary: detach, assign step, publish
  -> private memory queue in step_events.py
  -> StepMemorySampler: conversion and persistence

Timing continues through its separate queue and existing CUDA FIFO.

## In scope

- Move memory event and queue ownership to the shared handoff.
- Remove model_id from memory events and wire records.
- Replace the model-keyed pending dictionary with one optional event.
- Remove the unused model argument from internal flush functions and update
  SDK and Lightning callers.
- Share queue-draining code between timing and memory.
- Remove the obsolete sampler queue helper and model-filtering test helpers.
- Update existing tests, developer documentation, and the changelog.

No new files, capture classes, or transport layers are introduced.

## Sampler and measurement behavior

StepMemorySampler.sample() is unchanged. Sampler edits are limited to imports,
the queue reader, and removing model_id from event-to-sample conversion.

The following remain unchanged:

- Device selection and device labels.
- Peak-counter reset/read boundaries.
- Measurement timestamps, byte units, and non-CUDA None values.
- Existing step definitions and gradient-accumulation behavior.
- Queue capacity of 2,048 events and drop-on-full behavior.
- Draining published events after recording stops.
- Timing aggregation, CUDA FIFO processing, and event-pool handling.
- SQLite schema and downstream memory calculations.

Removing model_id is an intentional telemetry metadata change.
The SQLite projection already omitted it, so no database migration is needed.

## Out of scope

- StepCapture and shared completion/abort handling: part 3.
- Fixing existing failed-step attribution or session-buffer lifecycle.
- HF/Accelerate input-fetch and H2D measurement changes.
- Concurrent independent training loops or multiple pending device snapshots
  within one process.
- Changing integration-specific step boundaries.
- Combining timing and memory queues or synchronizing their samplers.
- New aggregation, backpressure, shutdown, or CUDA error-handling policies.

The single pending event can move into StepCapture later without changing
the published event or sampler interface.

## Validation

- Focused suite: 235 passed, 2 skipped, 1 watch smoke test excluded.
- Only two added memory regression tests:
  - Consecutive steps queued before draining, final-snapshot replacement,
    empty flushes, timestamps, and draining after recording stops.
  - Queue capacity, overflow, and retained-event ordering.
- Existing timestamp, wire-schema, CPU, SDK, integration, timing, and
  downstream memory tests retained and adapted.
- Code comparison confirmed unchanged device selection, memory reset logic,
  and sampler sample() behavior.
- Strict documentation build and formatting passed; no new lint findings.
- CPU collection overhead remained comparable to version_0.3.7.

Real CUDA hardware was unavailable; CUDA behavior was checked using
controlled test doubles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Memory events and wire records no longer include `model_id`. Pending
+  memory is one snapshot per process, published through the shared step-event
+  handoff. Device metadata, peak-counter boundaries, timestamps, byte units,
+  and the SQLite schema are unchanged.
 - Timing event types and queue access now live in
   `instrumentation/step_events.py`. Producers and the timing sampler share an
   explicit handoff; measurement boundaries, CUDA FIFO processing, and stored
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
-- Memory events and wire records no longer include `model_id`. Pending
-  memory is one snapshot per process, published through the shared step-event
-  handoff. Device metadata, peak-counter boundaries, timestamps, byte units,
-  and the SQLite schema are unchanged.
+- Memory events and wire records no longer include `model_id`. The previous
+  per-model pending keying is replaced by one process-local memory snapshot,
+  published through the shared step-event handoff. Device metadata,
+  peak-counter boundaries, timestamps, byte units, and the SQLite schema are
+  unchanged.
 - Timing event types and queue access now live in
   `instrumentation/step_events.py`. Producers and the timing sampler share an
   explicit handoff; measurement boundaries, CUDA FIFO processing, and stored
--- a/docs/developer_guide/step-time-pipeline-contract.md
+++ b/docs/developer_guide/step-time-pipeline-contract.md
@@ -12,16 +12,21 @@ For user-facing timing definitions, see the
 
 ## Training-to-sampler handoff
 
-`instrumentation/step_events.py` owns `TimeEvent`, `StepTimeBatch`, and the
-private timing queue. `utils/timing.py` measures regions and buffers events
-until the existing step boundary assigns their step number and publishes one
-batch through `publish_step_time_batch()`.
+`instrumentation/step_events.py` owns `TimeEvent`, `StepTimeBatch`,
+`StepMemoryEvent`, and two private queues. `utils/timing.py` measures regions
+and buffers events until the existing step boundary assigns their step number
+and publishes one batch through `publish_step_time_batch()`.
 
 ```text
 timed regions / optimizer hooks
   -> utils/timing.py: pending events and step flush
   -> instrumentation/step_events.py: timing batch queue
   -> StepTimeSampler: pending FIFO, CUDA resolution, aggregation, storage
+
+memory tracker
+  -> utils/step_memory.py: one pending snapshot and step flush
+  -> instrumentation/step_events.py: memory event queue
+  -> StepMemorySampler: conversion and storage
 ```
 
 The sampler calls `drain_step_time_batches()` to take available batches without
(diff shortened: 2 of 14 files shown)


Pendu:
Reviewed at `1114d16` against `version_0.3.7` (merge-base `7fba930`, which is #454), head and base checked out side by side. No prior human review, and the PR is not behind its base.

The plumbing move is faithful, same as its sibling. `StepMemoryEvent`, the 2048-slot queue, the drop-on-full policy, the drain loop, and the `record()` construction all carry over unchanged apart from the one dropped field. Memory and timing get separate queues rather than sharing one, so neither can starve the other, and the flush order is still memory then timing with one flush per boundary.

Two things I checked because they are the usual ways this goes wrong, both clean. The deleted block in `tests/integrations/test_hf_trainer.py` is a replacement rather than a removal: the helper now calls `drain_step_memory_events()`, every call site survives, and the assertion count is 19 before and 19 after. And `drain_queue_nowait` has zero remaining callers, so removing it leaves `samplers/utils.py` with just `ensure_session_dir`.

Where I would push back is on "minor cleanup", because three separate contract changes are sitting under it.

**The wire key removal is handled well, and I want to say so with evidence rather than just wave it through.** `model_id` disappears from the dataclass, from `to_wire()` and from `from_wire()`. I traced it: it was `id(model)`, a CPython object address, so it was never meaningful outside one process; it had zero consumers anywhere in src, tests, docs or examples; there is no SQLite column for it, so the CHANGELOG line "the SQLite schema are unchanged" is accurate; and it is declared in both the CHANGELOG and the contract doc. I also ran the actual compatibility matrix, feeding both payload shapes to `from_wire` on both trees:

```
BASE reader, OLD payload (has model_id): OK    BASE reader, NEW payload: OK
HEAD reader, OLD payload (has model_id): OK    HEAD reader, NEW payload: OK
```

All four combinations preserve every meaningful field. So the wire change is safe in both skew directions.

Two smaller notes on it. The field order changes, from `step, model_id, device, timestamp, ...` to `step, device, timestamp, ...`, so positional construction silently means something different; every construction in the repo is keyword-only, so this only reaches an external caller. And `StepMemorySample` is `@dataclass(frozen=True)`, so external code passing `model_id=` gets a `TypeError` rather than the tolerant behaviour the wire dict has. Neither is likely to bite, but the second is a harder break than the wire format itself.

**The one real behaviour change is the pending buffer, and it deserves its own line in the CHANGELOG.** Base kept `_temp_step_memory_buffer` keyed by `id(model)`; head keeps a single `_PENDING_MEMORY` slot per process. With two trackers on two devices in one process, that is measurable:

```
base: two record() calls, one flush -> 2 events  [(cuda:0, 5), (cuda:1, 5)]
head: two record() calls, one flush -> 1 event   [(cuda:1, 5)]
```

The dropped values are not duplicates. `record()` reads `torch.cuda.max_memory_allocated(self.device)` and resets with `reset_peak_memory_stats(self.device)`, so the two events held different numbers.

Having said that, I think the change is defensible and I want to give the argument for it rather than just the measurement. The renderer builds a single `global_rank -> device` map, described in its own comment as "best-effort latest known", so two devices inside one rank were never representable downstream in the first place. Head aligns the producer with the storage model the read side already assumes. And dropping per-model attribution is right on its own terms: two models on the same device share one allocator counter, so per-model attribution was never real, which is the argument your contract doc makes.

What I would still change is the disclosure. The CHANGELOG says "one snapshot per process" but does not say that per-model keying was removed, and that is the sentence a reader needs. Worth also deciding explicitly whether single-process multi-GPU is a shape TraceML intends to support: if it is, key the pending slot by `str(self.device)` rather than by nothing, which keeps your argument intact while not losing the second device; if it is not, the tracker could say so instead of silently keeping the last snapshot.

**The pending slot has no lock.** `event, _PENDING_MEMORY = _PENDING_MEMORY, None` is an unsynchronised read-then-write. Every shipped path pairs `record()` and flush adjacently in the same `finally`, so single-threaded training is fine, but two producer threads each inside `trace_step` can interleave so that one thread's flush publishes the other thread's snapshot under its own step number, which is a misattribution rather than a dropped event. `runtime/state.py` already guards analogous process-local state with an `RLock`, so the pattern is there if you want it. Otherwise a line in the module docstring stating the single-threaded-producer assumption would do.

**Two signature changes are not in the CHANGELOG.** `flush_step_events(model, step)` and `flush_step_memory_buffer(model, step)` both lost the `model` argument. Neither name is underscore-prefixed and both were importable before, so an external caller gets a `TypeError`. Every in-repo caller is updated, so this is a disclosure gap rather than a break.

For completeness, things I checked that came back clean: the suite is green at 1711 passed, 2 skipped; all linters pass at the pinned versions; the structure gate reports no read-side violations and correctly flags the wire schema as owing judgement by hand, which is exactly the `model_id` removal. `from traceml_ai.utils.step_memory import StepMemoryEvent` still resolves at head, and to the same object, so the old import path keeps working by accident of the import. And I checked whether the new single-slot test pins new behaviour: running the same sequence on both trees, two `record()` calls followed by two flushes of the same step produce one event with step 10 on base and on head, so last-record-wins and no-duplicate-on-empty-flush are existing behaviour that the test now characterises rather than anything this PR introduces.

abhinavsriva:
Thanks.

### Done

- Clarified that the previous per-model pending keying is replaced by one process-local memory snapshot.
- Documented the current assumption of one sequential training-step producer and one tracked device per process.

### Outside scope

- Device grouping is unnecessary for the current architecture because each rank/process tracks one device.
- Concurrent training loops in one process are unsupported, so adding a lock would not provide correct ownership and would add unnecessary complexity.
- Compatibility for internal flush signatures and direct schema construction is outside scope. The meaningful wire fields remain compatible, and the next PR removes the legacy flush path.

PR #456, opened 2026-09-09
Lightning callback: traced step includes H2D, non-training fetches excluded, no fabricated optimizer events, exception path

Pendu:
## What

`TraceMLCallback` for PyTorch Lightning measured the wrong span for the traced step, counted validation fetches as training input wait, fabricated a zero-length optimizer event on every accumulating micro-batch, and had no exception path. This fixes all four without changing the public interface (`traceml_ai.integrations.lightning.init()` + `TraceMLCallback()`), adds real `Trainer.fit()` tests for both the `lightning` and the legacy `pytorch_lightning` namespace, registers the Lightning conformance harness, and turns the CI integration job's `lightning` extra on so those tests run.

Base: `version_0.3.7`. No wire schema, SQLite table, payload key, artifact or CLI surface changes.

## Why, with the evidence

1. H2D sat outside the traced step. Lightning moves the batch to the device before `on_train_batch_start` fires (`lightning/pytorch/loops/training_epoch_loop.py:322` vs `:334` in 2.6.5). The callback opened `_traceml_internal:step_time` in `on_train_batch_start`, so on a GPU run the transfer was measured but excluded from Traced Step Time, and Step Time (= Input Wait + Traced Step Time) omitted it. On one T4 running `examples/integrations/lightning_minimal.py` for 200 steps, the GPU-clock envelope was 0.15 ms against a 5.42 ms transfer; the card printed "H2D 5.4ms 33%" as a share of a Step Time that did not contain it. The envelope now opens at the top of the `strategy.batch_to_device` wrapper (training batches only) and closes at `on_train_batch_end`; `on_train_batch_start` opens it only when nothing did (the `_DataLoaderIterDataFetcher` path).
2. Validation fetches landed in the next training step's Input Wait. The DataLoader patch is process-wide and gated only on arming (`src/traceml_ai/instrumentation/patches/dataloader_patch.py`). With `val_check_interval=4` and two validation batches, every fourth step carried `dataloader_next` with `n_calls=3`. A `suppress_dataloader_timing()` context manager, checked on every fetch and refcounted per thread, is entered at `on_{sanity_check,validation,test,predict}_start` and released at the matching `_end`. The count matters twice: the sanity check runs the validation loop inside it, and Lightning calls every callback's start hooks in registration order and the end hooks in the same order, so two holders exit first-in first-out rather than nesting.
3. Fabricated optimizer zeros. On accumulating micro-batches the callback recorded a `TimeEvent` with `cpu_start = cpu_end = 0.0`, `gpu_time_ms = 0.0`, `n_calls = 1`. The step-time analyzer already treats an absent occurrence-driven metric as "did not occur" for the average (`src/traceml_ai/step_time/analysis.py`, `OCCURRENCE_METRICS`), so removing the event leaves the run-level `optimizer_ms` unchanged (on the accumulation probe the summary reports 0.236 ms, the mean over all 40 steps with the absent ones as zero) while the raw rows and the per-step series stop carrying a measured zero.
4. Optimizer bracket closed in a hook that fires before backward. In Lightning 2.x `on_before_zero_grad` runs inside the closure before `backward` (`loops/optimization/automatic.py:130-140`), so the close there was a no-op and the region actually ended at `on_train_batch_end`. The dead override is removed; the region now ends at the next backward, the next forward, or the batch end, so a manual-optimization module with several optimizer steps per batch gets one event per step instead of one region swallowing the next iteration's work.
5. No exception path. Lightning calls `on_exception`, not `Callback.teardown`, when training raises (`trainer/call.py:68-73`). The callback left its timing regions open and its pending events buffered, and the next flush published them under the previous step number. `on_exception` now closes the regions, discards the pending timing events and memory snapshot, restores the wrappers, and lets the exception propagate. `teardown` discards too, so a second `fit()` in the same process starts clean whatever the first one left pending. A batch cut by `TRACEML_DISABLED=1` mid-flight is abandoned the same way instead of merging into the next one.
6. Silent dark streams. `on_train_start` now warns, never raises, when the init config leaves the DataLoader-fetch or H2D patch off, mirroring the HF callback, and when the transfer wrapper could not be installed.

The abort path uses the `StepCapture` lifecycle from #457 (this branch is rebased on #455 and #457): `_open_step_region` calls `begin_step_capture()` when it opens the envelope, which adopts the active capture so the H2D and envelope events recorded in the `batch_to_device` wrapper stay with the batch; `on_train_batch_end` calls `complete_step_capture(capture, step)`; `_abandon_pending` calls `abort_step_capture(capture)` for `teardown`, `on_exception`, and a batch cut by the kill switch. An earlier version of this branch carried its own discard helpers in `utils/timing.py`, `utils/step_memory.py` and `utils/flush_buffers.py`; they are gone, #457's capture is the primitive.

## What changes for users

- On GPU, Lightning runs report a larger Step Time because it now includes the transfer. Same T4 example: Step Time 16.3 ms to 21.2 ms, Traced Step Time 1.99 ms to 7.58 ms, H2D 5.4 ms unchanged, verdict INPUT-BOUND both times (Input Wait share 88% to 64%).
- One TraceML step is still one Lightning training batch. Under `accumulate_grad_batches=N` the optimizer phase is present only on the batches where the optimizer stepped and absent on the others; `optimizer_ms` in the summary reads as before.
- Validation, sanity-check, test and predict loaders no longer contribute to Input Wait or H2D.
- Docs (`docs/user_guide/integrations/lightning.md`) now state the launch contract: always `traceml run`, `--nproc-per-node=N` paired with `Trainer(devices=N)`; Lightning's own subprocess launcher is not used under `torchrun` (`Trainer(devices=2)` under a 1-process `traceml run` stops on Lightning's world-size check), and a bare `python train.py` trains untraced.
- `docs/data/integration_support.json`: Lightning CPU and single-process cells move to "CI tested", citing `tests/integrations/test_lightning_trainer.py`; GPU stays "Documented recipe".

## Testing

- Full suite: base `version_0.3.7` after #457: 1713 passed / 2 skipped (its own CI figure); this branch 1755 passed / 1 skipped (the disappearing skip is the Lightning conformance parametrization, which now has a harness). The `lightning` 2.6 distribution depends on `pytorch-lightning`, so the CI integration job runs both halves of the real-Trainer matrix (1325 passed, 1 skipped on this PR's first run).
- New tests: 8 for the suppress context manager (including two holders exiting in registration order and a double exit), 10 mocked callback tests, 11 real `Trainer.fit()` CPU tests parametrized over both namespaces (accumulation counts `[0,1,0,1]` with no zero-length events; validation fetch `n_calls` all 1 and the envelope and forward present on every step with `num_sanity_val_steps=2`; envelope already open inside `transfer_batch_to_device`; injected failure at batch 3 publishes steps 1 and 2 only and a fresh fit continues at 3; a full-epoch fit followed by a second fit inherits nothing; identical losses and `state_dict` with and without the callback; manual optimization with two steps per batch yields two optimizer events; warning under `init(mode="manual")`; standalone `validate()` publishes nothing; iterator-mode fallback publishes one step per batch), and a conformance negative control (no `init()` leaves exactly the fetch stream dark).
- CPU characterization, same two probes before and after: minimal example identical; accumulate-2-with-validation probe goes from fetch `n_calls` {1,3} to {1} and from 40 optimizer rows (20 zero-length) to 20 (none).
- GPU, 1x T4, torch 2.11.0+cu130, lightning 2.6.5, GPU clock over 200 steps: per-step envelope 0.15 ms before and 5.62 ms after against H2D 5.40 + compute 0.13 (the before run used `main` at 107432d, whose callback differs from this branch's base only by the #454 import move; the per-step figures come from the telemetry rows, the summary averages quoted above went through that tree's reporting); accumulate-2-with-validation run publishes 80 steps, fetch `n_calls` only 1, 40 optimizer rows, no zero-length events, H2D on all 80.
- pre-commit (isort, ruff, black, codespell) clean; `tools/integration_support_matrix.py --check` clean.
- External adversarial review of the source diff (an external review model, three passes with different lenses): 28 findings, 7 distinct issues confirmed and fixed in follow-up commits (a batch cut by `TRACEML_DISABLED=1` mid-flight merged into the next one; nothing drained the pending buffer at teardown; the optimizer region could absorb the next backward in manual optimization; no warning when the transfer wrapper failed to install; a dead flag; a silent `except` around the capability check; a save/restore suppression flag that two callback instances exiting in registration order would leave stuck on, darkening Input Wait for the rest of the process), the duplicates and refutations checked against the code or measured (event resolution happens in the sampler after publication, so the discard cannot return a CUDA event twice; `timed_region` guards its own setup; `on_exception` is only ever fired from the fit/validate/test/predict entry wrapper, which tears the trainer down in the same branch, and every new entry re-runs `setup`; a sized loader is never exhausted by Lightning's fetcher, so no epoch-end fetch event exists to leak), and the pre-existing behaviors listed below as follow-ups.

## Follow-ups, not in this PR

- The DataLoader patch records the exhausting `next()` that raises `StopIteration` as one extra fetch event: a plain `for` loop over a 4-batch loader records 5. Lightning's fetcher stops on the loader length and never exhausts a sized loader (two-epoch probe: `n_calls` 1 on every step), and with an unsized loader the exhausting call lands inside the last step of the epoch; plain loops and the other integrations carry it at every epoch end.
- `traceml.init(mode="auto")` together with `TraceMLCallback` double-counts optimizer time (global optimizer hooks plus the callback's region); the documented `traceml_lightning.init()` is selective and does not.
- `h2d_auto_timer.__exit__` clears its flag unconditionally instead of restoring the previous value; nothing nests it today.
- The same evaluation-fetch and pre-step-H2D gaps exist on the HF callback (#451); the suppress helper added here is framework-agnostic.
- Iterator mode (`training_step(self, dataloader_iter)`): Lightning never calls `batch_to_device`, the envelope opens at `on_train_batch_start`, and the user's fetch inside `training_step` lands inside Traced Step Time as well as in Input Wait. Same as before this change; pinned by a test, not changed.
- Two `TraceMLCallback` instances on one Trainer would each wrap `batch_to_device` and double the envelope; `setup` does not refuse a second instance today. The fetch suppression now survives that case; the forward, backward and H2D timers keep their own single-holder flags.
- `on_train_batch_end` advances the step counter before it flushes, as `trace_step` does; a flush failure would skip a number. Left for the step-capture lifecycle work (#453).

GATE: ran=true. `tests/integrations/test_telemetry_conformance.py` now executes the Lightning harness (`_run_lightning`) for the declared step_time, forward, backward, optimizer and dataloader_fetch streams, plus a negative control proving the harness reports a dark stream; full suite 1755 passed, 1 skipped on top of #455 and #457.
TRANSITIONS: training -> validation -> training (fetch suppression on/off, real Trainer with sanity check and mid-epoch validation); enabled -> TRACEML_DISABLED=1 mid-batch -> enabled (batch abandoned, next envelope fresh); fit -> exception at batch 3 -> fresh fit (steps 1,2 published, counter stays, second fit continues at 3); full-epoch fit -> teardown -> second fit (nothing inherited); optimizer region open -> closed by the next forward / next backward / batch end.
RANKS: not exercised beyond one rank. All new state is per process (thread-local flag, producer-side buffers, callback fields); DDP behavior is unchanged and the T4 runs were single-GPU.
TRIPWIRE: capability warning fired under `init(mode="manual")` naming dataloader_fetch and h2d (real Trainer test); missing-transfer-wrapper warning fired with a strategy-less trainer (mocked test); conformance negative control fired (fetch stream reported dark without `init()`).
TWINS: searched cpu_start=0.0|gpu_time_ms=0.0|cpu_end=0.0 (fabricated zero-length events), h2d_auto_timer()|batch_to_device (a transfer timed outside the step envelope), suppress_dataloader|on_evaluate|on_predict (non-training fetches in Input Wait), threading.local()|_prev (save/restore flags a second holder can interleave), flush_step_events callers (a stale step flushed after a failure) in src - found: no other fabricated zeros; one other transfer timer, inside `trace_step`, already within the envelope; the HF callback's evaluation fetches, tracked in #451; three sibling single-holder flags (forward, backward and H2D timers), pre-existing and listed under follow-ups; one other flush caller (`trace_step`, #453).
PLATFORM: n/a (this diff adds no platform-conditional code; it is callback, buffer and test logic in pure Python, exercised by the Linux integration job).


abhinavsriva:
Updated the Lightning integration with two focused correctness fixes:

- Validation/test DataLoader prefetches are excluded from training Input Wait, including unknown-length loaders.
- Optimizer events are omitted on gradient-accumulation-only batches, including strategy-owned accumulation such as DeepSpeed.

No broader callback redesign was added. Tests and documentation were updated accordingly. All focused tests pass.

PR #457, opened 2026-09-09
Centralize step completion and abort handling with StepCapture

abhinavsriva:
## Goal

Give pending timing and memory measurements one clear owner while preserving the existing queues, samplers, step definitions, and successful-run measurements.

Previously, timing and memory used separate pending buffers and a shared flush helper. A failed `trace_step` could still flush unfinished measurements using the previous completed step number.

This change introduces one process-local `StepCapture` for the existing sequential training-step lifecycle.

## What changed

- Added `StepCapture` as the owner of pending timing events and the final memory snapshot.
- Timing and memory producers now submit measurements to the active capture.
- Successful completion:
  - detaches the active capture;
  - assigns one step number;
  - publishes timing and memory through their existing queues.
- Exceptions propagated through `trace_step` abort the capture without publishing measurements or advancing the step counter.
- Repeated completion or abort attempts are harmless.
- Input timing collected before `trace_step` remains attached to the next step.
- Updated the direct SDK and Lightning integration to use the shared lifecycle.
- Removed the old timing and memory pending buffers and the redundant `flush_buffers.py`.
- Updated existing tests, developer documentation, docstrings, comments, and the changelog.

## Within scope

- One pending-step owner per process.
- Sequential step completion and abort handling.
- Repeated timing events within a step.
- Final process-local memory snapshot for the step.
- Correct recovery after a failed step.
- Existing queue capacity and drop-on-full behavior.
- Existing asynchronous CUDA resolution and FIFO processing.
- Existing step definitions and gradient-accumulation behavior.
- Draining measurements that were published before recording stopped.

## Outside scope

- HF-specific exception callbacks and input-attribution changes.
- Changes to framework-specific step definitions.
- Concurrent independent training loops within one process.
- Single-process multi-device grouping.
- Combining or synchronizing the timing and memory queues.
- Sampler, storage, transport, aggregation, or backpressure redesign.
- Global timing persistence.
- New synchronization or CUDA error-handling policies.

## Validation

- Focused SDK, integration, timing, memory, H2D, and sampler tests pass.
- Failed-step coverage verifies:
  - the previous completed step is not reused;
  - the failed step publishes nothing;
  - the counter does not advance;
  - the following successful step contains only its own measurements.
- Existing Hugging Face Trainer and gradient-accumulation tests pass.
- CUDA FIFO behavior remains covered without introducing synchronization.
- Formatting, linting, import checks, `git diff --check`, and the strict documentation build pass.
- No sampler implementation was changed.

Real CUDA hardware was unavailable; CUDA ordering and resolution were validated using the existing controlled test doubles.


Pendu:
Reviewed at `b5c8eff` against `version_0.3.7` (merge-base `25c4eeb`, which is #455), head and base checked out side by side. No prior human review, and the PR is not behind its base.

The headline claim holds, and I checked it by running it rather than reading it. Driving `trace_step` to failure three ways:

```
RuntimeError     -> propagated unchanged, 0 batches published
KeyboardInterrupt -> propagated unchanged, 0 batches published
SystemExit       -> propagated unchanged, 0 batches published
next step after an abort carries only its own event
after three aborts the next successful step is still step 1
```

So no partial telemetry, nothing leaks into the following step, aborted steps do not consume step numbers, and the user's exception comes out untouched. The ordering guards are careful too: `_detach_step_capture` installs a fresh capture before `_complete` publishes, so anything recorded during publication lands in the new capture rather than the one being closed, and the identity plus `_closed` checks make complete-twice, abort-after-complete and complete-after-abort all no-ops. The `flush_buffers` removal is clean, with zero remaining references and the memory-before-timing order preserved.

**One thing regressed, and I would fix it before merge.** The deleted `flush_step_events` rechecked `_traceml_disabled()` and `should_record_trace_events()` at flush time, so events already buffered were dropped if the kill switch flipped mid-step. `StepCapture._complete` and `complete_step_capture` check nothing, and the record-side gates only stop new events. Recording one event, then setting `TRACEML_DISABLED=1`, then completing the step:

```
base: 0 batches published while TRACEML_DISABLED=1
head: complete returned True, 1 batch published while TRACEML_DISABLED=1
```

So the kill switch no longer suppresses a step that was already in progress. Rechecking the gate inside `_complete` restores it.

Worth being precise about the test situation here, because it looks covered and is not. `test_disabled_env_dynamically_silences_low_level_utilities` sets the flag first and then records, so every event is stopped at the record side and the assertions correctly show nothing published. The ordering that regressed is the opposite one, record first and flip second, and no test exercises it.

**The HuggingFace path still publishes a failed step as a complete one.** This is not something you introduced, and it is identical on base, so it is not a merge blocker. I am raising it because it is the exact gap this machinery exists to close, and the PR description says step capture is going in now for later use in HF. `_close_step_cm_safely` always exits the suspended context with `(None, None, None)`, which resumes the generator normally and runs the success branch, so the `except BaseException` abort in `trace_step` is never reached from HF. Reproducing the callback shape, open the step, record an event, then close with `(None, None, None)` as the callback does after a failed step:

```
base: 1 batch published, carries the partial event
head: 1 batch published, carries the partial event
```

The comment at `on_step_begin` already names the situation, that a previous step raised and HF never fired `on_step_end`, and the recovery closes it normally. Closing those self-heal paths with `cm.gen.close()` instead would raise `GeneratorExit`, which your `except BaseException` already catches, so the abort path you built would start working for HF without any new machinery. The distinction that matters is that `_close_step_cm_safely` is called both for normal completion at `on_step_end` and for self-healing a step whose outcome is unknown, and only the second kind should abort.

**Nested captures merge silently.** `begin_step_capture` returns the active capture rather than a new one, which is deliberate and documented for input timing recorded before a framework opens its step callback. The consequence when two captures are actually nested is that only the first finalizer wins:

```
sequential: 2 batches, steps [1, 2], correct
nested:     1 batch at step 3 carrying BOTH the outer and the inner event
```

The outer completion returns `False` and publishes nothing. That return value is discarded at both call sites, `sdk/instrumentation.py:203` and `integrations/lightning.py:436`, so a lost completion produces no diagnostic anywhere. Nesting `trace_step` is unusual, but it is reachable for a user on the Lightning or HF integration who also wraps a section manually, and the symptom looks like data rather than an error. Logging when `complete_step_capture` returns `False` would surface it, and the information is already there.

On the Lightning coupling you mentioned, here is what I found so you have the concrete shape. A trial merge reports both branches touching four files: `CHANGELOG.md`, `integrations/lightning.py`, `utils/step_memory.py` and `utils/timing.py`. On `lightning.py` this PR is +37 and #456 is +204/-64. Both of us independently hardened the same path: yours closes the three contexts in `teardown` then aborts the capture, mine routes through an `_abandon_pending` helper that closes the contexts, exits suppression, discards the step events and drops the memory tracker. The difference worth keeping in whichever lands second is that #456 also adds an `on_exception` hook and this PR does not, so the merged result should end up with both entry points rather than only `teardown`. I could not check which hook Lightning fires first, since Lightning is not installed in the environment I reviewed in, so treat that ordering as unverified. Happy to rebase #456 onto this once it lands if that is easier than the other direction.

For completeness, checked and clean: the suite is green at 1713 passed and 2 skipped; every linter passes at the pinned versions; the structure gate reports no read-side violations and flags the SDK surface as owing judgement by hand, which is this PR's own new public surface. Line lengths over 79 exist in the touched files but the pinned black and ruff both pass, so nothing is enforced there and several of them predate this PR.

abhinavsriva:
Thanks!

### Addressed

The completion path now rechecks the recording gate. If recording is disabled after measurements were collected, the capture is detached and discarded instead of being published or retained for a later step.

The existing disabled-runtime test now covers the exact ordering: record timing and memory, enable `TRACEML_DISABLED`, complete the capture, and verify both queues remain empty.

Validation: 137 passed, 2 skipped. Ruff, Black, imports, `git diff --check`, and the strict documentation build pass.

### Outside scope

HF failed-step recovery remains part of #456. Nested `trace_step` support is also outside the documented sequential producer lifecycle, so this PR does not add capture stacks, locks, or additional ownership machinery.

PR #459, opened 2026-09-12
Dev/abhinav/issue451 lifecycle

abhinavsriva:
## Harden Hugging Face Trainer step lifecycle

This PR needs https://github.com/traceopt-ai/traceml/pull/452 to be merge to version_0.3.7 and then this is changed on top of that. 

### Goal

Make TraceML’s Hugging Face callback handle complete, failed, retried, and interrupted Trainer steps without publishing partial measurements.

### What changed

- Added lifecycle handling around each Hugging Face Trainer attempt.
- Discards incomplete step captures when training fails or retries.
- Prevents failed attempts from incrementing the TraceML step number.
- Handles automatic batch-size retries and early stopping during gradient accumulation.
- Makes callback reuse safe across multiple training runs.
- Ensures only one TraceML callback owns a run when duplicates are registered.
- Updated the Hugging Face guide, pipeline contract, and changelog.

### Scope

This PR covers Hugging Face `Trainer` step lifecycle and callback ownership.

### Outside scope

- Training-input and H2D measurement
- Changes to shared step capture, queues, samplers, or reporting
- Direct Accelerate loops without Hugging Face Trainer
- Other framework integrations

### Validation

- Hugging Face integration and telemetry tests pass: `24 passed`
- Broader integration test selection passes: `1348 passed, 1 skipped`
- Strict documentation build passes
- `git diff --check` passes

### Dependency

This PR is based on the callback-only Hugging Face integration branch and should be reviewed after that change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- **Removed:** `traceml_ai.integrations.huggingface.TraceMLTrainer` and its
+  `traceml_enabled` argument. The wrapper only installed the callback. Call
+  `traceml_ai.integrations.huggingface.init()` and register
+  `TraceMLTrainerCallback()` with standard `transformers.Trainer` instead.
+  For optional tracing, register the callback conditionally. See the
+  [HF migration instructions](docs/user_guide/integrations/huggingface.md#migration)
+  for the replacement setup.
 - PyTorch Lightning: `TraceMLCallback` now opens the traced step when
   Lightning moves the batch to the device, so the H2D transfer is inside
   Traced Step Time and Step Time on GPU runs; fetches of validation,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ which carry the full historical notes for versions predating this file.
 
 ## [Unreleased]
 
+- Hugging Face Trainer now discards unfinished step captures when training
+  raises, including before automatic batch-size retries. Reused and duplicate
+  TraceML callbacks cannot publish a failed group or count a step twice.
 - **Removed:** `traceml_ai.integrations.huggingface.TraceMLTrainer` and its
   `traceml_enabled` argument. The wrapper only installed the callback. Call
   `traceml_ai.integrations.huggingface.init()` and register
--- a/docs/developer_guide/architecture.md
+++ b/docs/developer_guide/architecture.md
@@ -42,7 +42,7 @@ Samplers maintain an incremental append counter per rank per table. The sender s
 The `src/traceml/` package is a deprecated compatibility alias for older import
 paths. New implementation work should go under `src/traceml_ai/`.
 
-For the user-facing API surface (`trace_step`, `TraceMLTrainer`, `TraceMLCallback`, CLI usage), see the [Public API](../user_guide/public-api.md). The source tree above is the canonical reference for internals — start from the entry points and follow the imports.
+For the user-facing API surface (`trace_step`, `TraceMLTrainerCallback`, `TraceMLCallback`, CLI usage), see the [Public API](../user_guide/public-api.md). The source tree above is the canonical reference for internals — start from the entry points and follow the imports.
 
 Contributors changing Step Time should begin with the
 [Step Time pipeline contract](step-time-pipeline-contract.md), which maps its
--- a/docs/developer_guide/step-time-pipeline-contract.md
+++ b/docs/developer_guide/step-time-pipeline-contract.md
@@ -287,10 +287,10 @@ training step. The existing raw input event names are
 These gaps can omit transfers from Step Time or assign loader work to the
 wrong training step.
 
-If training is interrupted, callback cleanup can record an unfinished group
-as completed. This requires the separate lifecycle fix. See the
-[HF integration guide](../user_guide/integrations/huggingface.md#limitations)
-for user-facing limitations.
+If training is interrupted, the Trainer lifecycle guard aborts the open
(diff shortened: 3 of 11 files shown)


Pendu:
Reviewed at `214d6d08` against `version_0.3.7` (merge-base `6e67e39`), head and base checked out side by side. No prior human review on this PR, and it is not behind its base.

First the structural note, since it changes what to read. This branch contains #452's two commits plus one new one, so `733330756` is an ancestor of this head and the actual new surface is `733330756...214d6d08f`: +312/-39 across five files, with `integrations/huggingface.py` +152 and the HF test file +181. Your PR body already says the plan is to land #452 first and rebase this on top, which is the right order. Worth closing one of the two once that happens, since both are currently open against the same base and both contain those two commits. My review of #452 applies to this branch too.

**The two things I raised earlier are closed, and I measured both rather than reading them.**

The failed-step case first. The old path closed a suspended step with `cm.__exit__(None, None, None)`, which resumes the generator normally and runs the success branch, so a step that failed got published as a complete one. `_abort_step_cm_safely` now exits with `cm.__exit__(type(abort), abort, None)`, which throws into the generator and lands in `trace_step`'s `except BaseException`. Driving both paths on this head with a partial step open:

```
old close (success branch) -> 1 batch published, counter +1
new abort path             -> 0 batches published, counter +0
```

So nothing partial is published and the counter is untouched, which is exactly the two claims in your description.

Then duplicate registration. The guard elects `callbacks[0]` as owner and marks the rest non-owners, and the three hooks gate on `_owns_run`:

```
1 callback,  with election    -> 1 batch, counter +1
2 callbacks, without election -> 1 batch, counter +2
2 callbacks, with election    -> 1 batch, counter +1
```

The middle row is the behaviour I measured on #452 after the wrapper's dedup guard was removed. The third row is this PR fixing it, and it does fix it on the path the docs describe.

**But the fix only holds where the guard is installed, and the changelog claims it unconditionally.** The election lives inside `guarded_inner_training_loop`, which only exists once `init()` has patched `Trainer._inner_training_loop`. The callback itself carries no duplicate check. I checked whether that gap is reachable, and it is, because the callback publishes perfectly well on its own:

```
no init(), 1 callback  -> 1 batch published, counter +1
no init(), 2 callbacks -> 1 batch published, counter +2
```

So a user who registers the callback without calling `init()` still gets telemetry, and still gets the old double-counting. Two other ways to land there: the patch install failing, which is caught and reduced to a single stderr line, and a `Trainer` subclass that defines its own `_inner_training_loop` and shadows the class attribute. Worth either scoping the changelog sentence to the `init()` path or moving the ownership check somewhere the callback always reaches, such as electing lazily in `on_step_begin`.

Related, and the reason I would not leave this purely to the docs: `_owns_run` is sticky. Once an instance is demoted it stays demoted until a guarded attempt re-elects it, so reusing a demoted callback on a run that bypasses the guard gives a completely silent dark run:

```
after demotion, _owns_run = False
reused on an unguarded run -> 0 batches, counter +0, nothing on stderr
```

Zero telemetry and zero diagnostic is a worse failure than double counting, because nobody notices it.

One item from that review is still open, and I mention it only so it does not get lost: the upgrade error for the removed `TraceMLTrainer` still carries no migration hint. That is untouched here, which is expected since this commit predates the review.

The rest of the design holds up where I poked at it. Putting the cleanup in a `finally` around `Trainer._inner_training_loop` is the right seam, because that is inside Accelerate's OOM retry boundary rather than outside it, so a failed attempt cannot leak into the retry. The `_traceml_lifecycle_guard` attribute keeps a second `init()` from stacking wrappers.

I did go looking for one specific problem and did not find it, so I will say so rather than leave it implied. Patching `Trainer._inner_training_loop` means depending on a private method of a third-party library, and `original = Trainer._inner_training_loop` is a bare attribute access that would raise on a transformers version that renames it. But `init()` wraps the whole installation in `try/except` and routes the failure to `_log_hf_error`, so a future rename degrades to one stderr line rather than breaking the user's setup. That is the right call. The residual is only that when it does degrade, the user silently gets the pre-PR behaviour back with nothing but a startup message to say so, so if that line is ever seen in the wild it will be worth treating as load-bearing rather than noise.

Two small things in the new code.

The `except _TraceStepAbort: pass` branch is unreachable. `@contextmanager.__exit__` catches the re-raised injected exception and returns `False` rather than propagating it, which I confirmed by calling `cm.__exit__(type(abort), abort, None)` on a live `trace_step`: it returns `False` and does not raise. The comment above it says it is defensive for context managers that propagate the injected signal instead, which describes a `trace_step` that does not exist in this codebase. Dead code is minor; dead code whose comment implies it fires is the part worth removing, because the next reader will assume that path is live.

The guard's early return skips the cleanup. When `_traceml_callbacks(trainer)` comes back empty the wrapper calls the original loop directly, with no `try/finally` and so no pending-capture abort. A user who calls `init()` but registers no callback still has the DataLoader patches installed, so fetch events accumulate in the active capture with nothing on that path to clear them. It is a pre-existing pattern rather than something this PR introduces, and the guard's preamble now cleans it up on the next run that does have a callback, but the no-callback path itself is still uncovered.

On coverage: the five new tests each pin one of the defects the old code had, which is the right shape. The gap I would name is `KeyboardInterrupt`. It travels the same `except BaseException` route as the injected abort so it should behave identically, and I verified that route directly on the earlier step-capture PR, but nothing here pins it.

For completeness: the suite is green at 1782 passed and 1 skipped, `mkdocs build --strict` is clean, every linter passes at the pinned versions, and the structure gate reports no read-side violations while flagging the runtime lifecycle contract as owing judgement by hand, which is this PR's own subject.

PR #460, opened 2026-09-16
Dev/abhinav/retention window correction

abhinavsriva:
# Merge cleanup: history labels, Lightning accumulation, and Ray migration docs

## Summary

This is focused cleanup before merging `version_0.3.7` into `main`. It addresses three specific issues found during the pre-merge review without redesigning the surrounding systems.

### 1. Correct retained-history labels

- Changes System chart wording from **Whole Run** to **Recent History**.
- Updates tooltips to describe the history currently available after retention.
- Does not change chart data, retention behavior, point budgets, or time axes.

### 2. Correct Lightning gradient-accumulation boundaries

- Treats all microbatches in one automatic accumulation group as one TraceML step.
- Publishes one timing batch and one memory event per optimizer-update boundary.
- Uses the existing sampler behavior to sum repeated timing events.
- Measures CUDA peak memory across the complete accumulation group.
- Handles shorter final groups and discards incomplete groups after failures.
- Keeps manual optimization batch-scoped.

This does not change the sampler, storage schema, Hugging Face integration, or plain PyTorch instrumentation.

### 3. Document the Ray retention migration

- Documents the replacement of `summary_window_rows` with `history_retention`.
- Clarifies that row counts cannot be automatically converted to durations.
- Removes the unsupported claim that TraceML retains five additional raw minutes.
- Does not add a compatibility alias or change Ray runtime behavior.

## Out of scope

- Retention or pruning algorithm changes.
- Telemetry-health and data-loss hardening.
- New aggregation or storage behavior.
- Broader integration refactoring.
- Compatibility support for removed configuration arguments.

## Verification

- Full suite: **1,788 passed, 1 skipped**.
- Both `lightning` and `pytorch_lightning` namespaces tested.
- Full and partial accumulation groups tested.
- Mid-group failure and memory-reset failure paths tested.
- Ray integration: **9 tests passed**.
- Ruff, diff checks, and integration-support matrix passed.


Pendu:
Reviewed at `3507406` against `version_0.3.7` (merge-base `99234a2`). Commit 2 is not a cleanup: it moves the Lightning traced-step boundary from one step per training batch to one step per optimizer-update group. Measured on real Lightning runs:

| run | base | head |
| --- | --- | --- |
| 8 batches, `accumulate_grad_batches=1` | 8 steps | 8 steps |
| 8 batches, `accumulate_grad_batches=4` | 8 steps | 2 steps |
| 10 batches, `accumulate_grad_batches=4` | 10 steps | 3 steps |

Items, most important first.

1. A crash inside an accumulation group now publishes nothing where it used to publish everything up to the failure. Crashing on micro-batch 7 of an 8-wide group: base published 6 steps of pre-crash evidence, head publishes 0. `lightning.py:412-430` discards the whole open capture through `_abandon_pending`. This is the run a user is most likely to be diagnosing, so it is worth deciding deliberately rather than inheriting it from the boundary move. The same widening applies to early stop, `max_time`, a mid-group kill switch, and the step cap landing mid-group.

2. No CHANGELOG entry for the boundary change. `CHANGELOG.md` is not in the diff. Any user with `accumulate_grad_batches>1` gets fewer steps, larger per-step numbers, and a memory peak that now spans the group, with no notice. The existing `[Unreleased]` Lightning bullets still describe the old per-batch world.

3. Under DeepSpeed the short final group is silently dropped, and the contract doc claims the opposite. Lightning's `_should_accumulate` returns `not accumulation_done and (strategy.handles_gradient_accumulation or not is_final_batch)`, and `DeepSpeedStrategy.handles_gradient_accumulation` is `True`, so the final partial group still reports "accumulating", never completes, and is dropped at teardown. `step-time-pipeline-contract.md:247` says "Lightning's final-batch decision completes a shorter final group", which holds for the default strategy but not this one.

4. `lightning.md:361` says the recorded step count advances with `trainer.global_step`. Lightning's own docstring for `global_step` says it "includes multiple optimizers (if enabled)", so with two optimizers `global_step` advances twice per boundary while TraceML advances once. Manual optimization diverges too. The sentence needs the single-optimizer, automatic-optimization qualifier the rest of that section carries.

5. Validation landing inside a group can inflate that step's memory peak. The reset happens at the first micro-batch and the record at the boundary, and Lightning runs validation in `on_advance_end`, after the `on_train_batch_end` hook. With `val_check_interval` not a multiple of `accumulate_grad_batches`, the CUDA peak window spans the validation loop. Timing is unaffected because the val gates exclude it. No test combines validation with accumulation.

6. `accumulate_grad_batches` changed mid-run by a scheduler is untested. Groups of different sizes then publish with different micro-batch counts per step and are not comparable across the boundary.

7. `limit_train_batches` cutting a run mid-group is untested.

8. The new label and its own tooltip disagree: `system_section.py:535` renders "recent history" while `:539` renders "Available history". Pick one.

9. The internal `whole_run` identifier stays (10 occurrences in `system_section.py`) while the user-facing label became "recent history". No SQLite column, payload key or saved layout carries the old label, so nothing is broken. It is a vocabulary mismatch for the next reader, not a half-applied rename.

10. Not yours, but worth knowing: `tests/runtime/test_final_summary_smoke.py::test_final_summary_json_smoke` failed once in a full-suite run on this branch and then passed on two re-runs in both orderings, and passes in isolation on both trees. It is a flaky test in the release branch, not this PR.

On the change itself: aligning Lightning with the HuggingFace boundary is defensible, and the docs describe the new state accurately. Items 1 and 2 are the ones I would settle before merge, since one changes what a failed run gives you and the other is what a user reads on upgrade. Items 3 and 4 are claims in the new docs that are not true for every configuration.

Measurements above came from real Lightning runs on both trees; the DeepSpeed and `global_step` items are from Lightning's installed source. The one thing I could not measure directly is the effect on `trace_max_steps`: since a step now spans N micro-batches, a given cap covers N times more training, but my probe for that failed for a harness reason on both trees, so treat it as arithmetic from the table rather than a measurement.

abhinavsriva:
Thanks for the detailed review. 

The optimizer-update boundary is intentional: under automatic gradient accumulation, a completed TraceML step now represents one completed update group rather than one micro-batch. Therefore, if training fails before the update boundary, that incomplete group is discarded. Publishing its micro-batches as completed steps would restore the previous semantic mismatch; retaining partial forensic evidence would require a separate partial-step model and is outside this PR.

Validation allocations can contribute to the CUDA peak when validation occurs while an accumulation group is open. We consider this a conservative peak for that step window: the group closes at the optimizer boundary, the next step resets the peak counters, and validation timing remains excluded. We will keep the implementation simple.

Before merging, I will add the behavior change to the CHANGELOG, qualify the shorter-final-group wording for strategy-owned accumulation such as DeepSpeed, qualify the `trainer.global_step` statement for automatic single-optimizer training, and make the “Recent history” label and tooltip consistent. Dynamic accumulation schedules, a separate partial-step representation, validation-segmented memory tracking, and internal `whole_run` renaming will remain out of scope.

PR #462, opened 2026-09-17
rfdetr integration

abhinavsriva:
## Summary

Adds an opt-in TraceML integration for RF-DETR’s existing `model.train()` API. Users can inspect training phases and compare configurations using `traceml run` and `traceml compare`.

```python
from traceml_ai.integrations import rfdetr as traceml_rfdetr

traceml_rfdetr.init()
model.train(...)
```

## Changes

- Attach tracing without replacing RF-DETR’s default callbacks, EMA, checkpoints or training loop.
- Measure the inner detection model and attach forward instrumentation after EMA initialization.
- Exclude dataset previews and evaluation from training measurements.
- Preserve gradient-accumulation boundaries, including partial final groups.
- Log forward-instrumentation setup failures while allowing actual training errors to propagate.
- Add a runnable worker-count comparison example, integration guide and dedicated CPU CI job.

RF-DETR remains an optional dependency. Initial scope is eager object detection on CPU/CUDA with single-process execution or ordinary DDP.

## Validation

- Verified numerical parity for losses, model weights, optimizer state and EMA weights.
- Tested accumulation, evaluation isolation, failure cleanup and repeated fitting.
- Ran an offline Nano smoke test through the public training API.
- Verified two-rank CPU/Gloo collection and comparison using a lightweight RF-DETR fixture.
- Ran the documented example with pretrained Nano at 384px using zero and two loader workers.
- Checked RF-DETR 1.10.1 and development commit `0ed5be8`; lint and strict documentation checks passed.

CUDA/NCCL and physical multi-node execution remain unverified. The short example runs establish functionality, not performance improvements.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in RF-DETR integration support with training instrumentation, documentation, usage guidance, and a runnable example.
  - Added RF-DETR to the supported integrations matrix and documentation navigation.
  - Added scoped DataLoader timing to prevent idle or skipped instrumentation from recording fetches.

- **Bug Fixes**
  - Improved Lightning and RF-DETR error handling so diagnostic output failures do not interrupt training.
  - Corrected completed-step totals in step-time and step-memory summaries, including when history is limited.

- **Tests**
  - Added offline CPU validation and expanded integration and reporting tests for RF-DETR, Lightning, DataLoader timing, and completed-step totals.


Pendu:
Reviewed `77c91c6b` against `main` `bf302699` (23 files, +1858/-9, 0 behind).
CI is green including the new job.
The only prior review is CodeRabbit's: its CI-token point holds, its `accelerator` point does not, and its "do not raise into training" point is item 2.

1. **`training_total_steps` drops by one for every integration, and nothing announces it.**
`step_time/builder.py:351` and `step_memory/loader.py:265` go from `latest_step + 1` to `max(0, latest_step)`.
The new value is the right one.
A real 3-step Lightning fit publishes steps `[1, 2, 3]` on both trees, so the same run reports `training_total_steps = 4` at `bf302699` and `3` here.
This was wrong in every released version, not only 0.4.0.
At `v0.2.15`, `src/traceml/integrations/lightning.py:186` already advanced before flushing, so `latest_step` has been one-based throughout and the fix is right for archived summaries too.
Blast radius is contained: `compare/`, `renderers/` and `aggregator/` never read either field.
Two asks.
Put it in the release notes as a corrected value.
And check the schema-version precedent, since #277 bumped `schema_version` to 1.7 for a value-semantics change and `SCHEMA.md:3` still says 1.8.
One more thing worth knowing: the only end-to-end proof is `test_rfdetr_trainer.py:269-271`, which runs only when rfdetr installs.
A change affecting every integration is pinned behind the rfdetr-gated job.
Cleanest would be its own PR.

2. **`rfdetr.py` aborts the user's training run over a TraceML configuration judgement.**
The three `raise ValueError` at `rfdetr.py:76`, `:116` and `:132` sit inside the `build_trainer` wrapper, which RF-DETR calls from `model.train()`.
Every other integration raises exactly one thing, a missing-dependency `ImportError`.
The rule is written in our own tree at `_capability.py:7`: "Warn, never raise (fail-open: instrumentation must never break training)".
`lightning.py:320` follows it.
The guide documents this as deliberate (`rfdetr.md:121`), so this is a question rather than a fix request.
Is a user with `segmentation_head=True` who forgot to remove `init()` meant to lose the run?
The module is also inconsistent with itself.
An unqualified rfdetr *version* only warns (`rfdetr.py:197-204`) while an unqualified *mode* is fatal.
A `TRACEML_RFDETR_STRICT=0` escape hatch, or warn-plus-passthrough, gets the same "no wrong numbers" guarantee without costing the run.

3. **`strategy="ddp_find_unused_parameters_true"` kills the run, while the identical implicit spelling is accepted.**
`_validate_mode` allows `{"auto", "ddp"}` only (`rfdetr.py:62`).
`ddp_find_unused_parameters_true` and `_false` are registered Lightning strategies on the same subprocess launcher, and RF-DETR passes them straight through.
RF-DETR's own `strategy="ddp"` path builds `DDPStrategy(find_unused_parameters=True)`, which is exactly what that alias means.
So we accept it implicitly and reject it explicitly.
Suggest widening the allowlist to the `ddp*` aliases that are not spawn, fork or notebook.
Related, in the same function: `rfdetr.py:66` imports `DDPStrategy` from `pytorch_lightning` only.
With both Lightning namespaces installed those are two different classes, so a `lightning.pytorch` strategy instance fails the `isinstance` and gets rejected.
That path is only reachable through a direct `build_trainer` call, not through `model.train()`.
`lightning.py:66-93` already carries dual-namespace handling worth reusing.

4. **The CUDA-graphs guard cannot fire.**
`rfdetr.py:49-53` reads `getattr(config, "cuda_graphs", False)`, and `cuda_graphs` appears nowhere in rfdetr 1.10.1, so it is always `False`.
`test_rfdetr.py:187` passes only because the fixture sets the attribute on a `SimpleNamespace`.
Either drop the branch and the claim at `rfdetr.md:119`, or key it on something RF-DETR actually exposes.
The other three field names carry the same risk if RF-DETR renames one, since `getattr(..., False)` fails open and we would trace an unsupported config instead of refusing it.

5. **CodeRabbit's `accelerator` comment on `rfdetr_minimal.py:116` is wrong, do not apply it.**
RF-DETR maps `device` to the Lightning accelerator itself.
`_resolve_trainer_device_kwargs` at `detr.py:899-903` returns `"cpu"` for `cpu` and `"gpu"` for `cuda`, and `detr.py:1090-1093` passes that as `build_trainer`'s `accelerator`.
`TrainConfig` deliberately has no `device` field (`config.py:1228`).
The suggested edit is redundant, and for `--accelerator cuda` it would set a different string than the one RF-DETR derives.

6. **The new CI job installs and imports third-party training code with the checkout token still in the git config.**
`ci.yml:148-173` sets neither `permissions: contents: read` nor `persist-credentials: false`.
The same file already does both for the one other job that runs third-party code, `test-latest-deps` at `ci.yml:188-200`, with a comment explaining why.
Two lines.

7. **47 of the new job's tests already run in `test-integration`.**
That job runs `tests/integrations` wholesale (`ci.yml:137`), so `test_rfdetr.py` (34) and `test_rfdetr_example.py` (13) run there too.
Neither needs rfdetr.
Only 7 test functions require the real package, 1 in `test_rfdetr_nano.py` and 6 in `test_rfdetr_trainer.py`.
Narrowing `test-rfdetr` to those two files keeps the coverage and drops the duplicate.

8. **The two `lightning.py` changes fix real bugs for every Lightning user, not just RF-DETR.**
With a two-batch DataLoader pass before `fit()`, fetch events per published step go from `[3, 1]` at `bf302699` to `[1, 1]` here.
The new `_abandon_pending` at `lightning.py:306` stops pre-fit fetches inflating step 1's Input Wait.
The `self._step_capture is None` guard at `lightning.py:364` is the second fix.
A forward fired while `trainer.training` is True but outside a batch used to land in whatever capture was globally active, so it was charged to the next step.
Neither change drops an in-batch measurement, because `_open_step_region` assigns the capture before the region can fail.
Two follow-ups.
The comments explain both only in RF-DETR terms and the only new coverage is the fake-hook test at `test_lightning.py:742`, so a real-Trainer test in `test_lightning_trainer.py` would be worth having.
And `setup()` now aborts the process-wide capture on every stage, not only fit, so scoping it to `stage == "fit"` would keep the blast radius where you intend it.

9. **`0ed5be8` is cited without naming the repository.**
It appears in `rfdetr.md:128` and in the generated matrix row.
It is not a TraceML commit, since `git cat-file -t` fails on `main`; it resolves in `roboflow/rf-detr`.
A reader will assume it is ours.
"RF-DETR development commit `0ed5be8`" fixes it.

10. **Small ones.**
RF-DETR is missing from `docs/user_guide/public-api.md`, which carries mkdocstrings entries for huggingface, lightning and ray.
`rfdetr.py:198` is the only `warnings.warn` in `integrations/`, where everything else reports to stderr with the `[TraceML]` prefix, and a `UserWarning` disappears under `-W ignore`.
Four new lines exceed the 79-char limit: `lightning.py:345`, `rfdetr.py:201`, `rfdetr_minimal.py:37` and `:104`.
No hook catches them, since ruff's defaults have no E501 and the pre-commit scope excludes `examples/` and `tests/`.
`test_fixtures.py:144` compares an int to a bool (`steps_analyzed == bool(steps)`).
And `training_total_steps` now always equals `training_latest_step`, so one of the two fields is redundant.

Checked and found fine:

- The generated `integration_support.json` regenerates byte-identical.
- The adapter is idempotent, and `evaluate()` stays uninstrumented because it passes `include_training_callbacks=False`.
- The stub in `test_rfdetr.py` matches the real `build_trainer` signature exactly, `strategy` included, which lands in `**trainer_kwargs` rather than a named parameter.
- The example's `devices`, `num_nodes` and `strategy` kwargs are all real `TrainConfig` fields.
- The empty `"issues": []` in the matrix row is justified: `test_rfdetr_trainer.py` starts a real aggregator subprocess and `test_rfdetr_example.py` validates the new example.
- The forward-wrap move to `on_train_start` is sound and load-bearing.
  RF-DETR's EMA callback builds its averaged copy in `setup()`, and a `deepcopy` of a wrapped forward would have pointed the EMA model at the live one.
- Appending the callback after Trainer construction does place it after the checkpoint callbacks.
  Nothing in RF-DETR saves on `on_train_batch_end`, and `rfdetr.md:104` already documents the effect.
- The 50 Lightning tests pass at head.

Pendu:
Re-checked at `113562c5`.
Items 2 to 10 are addressed, and item 1 is now announced in `CHANGELOG.md` and `SCHEMA.md`.
The three new real-Trainer tests in `test_lightning_trainer.py` close the coverage gap from item 8.
Your fix for item 2 surfaced something my suggestion there missed, which is item 1 below.
Items 4 and 5 correct things I got wrong in the first review.

1. **The warn-and-skip path leaves fetch timing armed with nothing to collect it.**
`init()` calls `lightning.init()` at `rfdetr.py:182`, before the interface check at `:187-197`.
So the DataLoader patch is armed before the adapter knows it can instrument.
On a skip, no callback is installed, so no fetch-timing scope is entered and no capture is completed or aborted.
Every training fetch then records into the process-global capture and stays there.
I measured it on CPU with the real `lightning.init()` and a real `Trainer`.
With a segmentation config, a 1000-step fit published no steps and left 1000 pending `dataloader_next` events.
A supported config on the same harness published all 20 of its steps and left none pending.
Nothing reaches the summary, so the numbers stay clean.
The capture still grows by one event per step for the whole run.
On CUDA, each fetch region also takes two CUDA events (`timing.py:97-98`), and the pool creates new ones when it is empty (`cuda_event_pool.py:25-32`).
The unit tests cannot see this because the `rf_factory` fixture replaces `lightning.init` with a stub (`test_rfdetr.py:53`).
My item 2 suggestion was incomplete: passthrough also has to stop the timing that `init()` already armed.
One option is to run the interface check before `lightning.init()`.
For a per-fit skip, the factory could attach a minimal callback that holds a disabled fetch-timing scope and abandons the capture at teardown.

2. **Evaluation after training accumulates the same way, and the `init()` docstring says otherwise.**
RF-DETR's `evaluate()` builds its trainer with `include_training_callbacks=False` (RF-DETR 1.10.1 `detr.py:1272`), so it gets no callback and no scope.
On a Lightning analogue, a traced fit left nothing pending.
One 1000-batch `validate` on an uninstrumented `Trainer` then left 1000 pending events, and a second left 2000.
This predates the PR, and the general accumulation belongs in its own issue.
For this PR, `rfdetr.py:156` should drop the claim that the hook leaves standalone evaluation unchanged.

3. **Lightning's tuners still land in step 1.**
The comment at `lightning.py:305` says tuning fetches are outside the training measurement.
`LearningRateFinder` and `BatchSizeFinder` do their search in `on_fit_start`, after `setup()`, so the abort there does not reach them.
I used `LearningRateFinder(num_training_steps=10, early_stop_threshold=None)` with a 50 ms training step.
Without the early-stop setting the finder can stop early, and the count varies between runs.
Step 1 absorbed 11 fetch events in every run.
Its envelope was 613 to 630 ms over three seeded runs each on this head and on `main`.
Steps 2 and 3 were 52 to 59 ms, with one fetch each.
`main` behaves the same, so this is not a regression.
Narrowing the comment to pre-fit previews would make it accurate.

4. **The RF-DETR support-matrix row should keep #141 and #323.**
I wrote that the empty `"issues": []` was justified, which holds only for #320.
`test_rfdetr_trainer.py:329-338` runs the real launcher, which covers #320.
#141 still applies because the adapter is the Lightning callback.
#323 still applies because `test_rfdetr_example.py` tests the example with RF-DETR and torch stubbed.
The Lightning row lists all three.

5. **Three more corrections to the first review.**
- Item 6: `test-latest-deps` is not "the one other job that runs third-party code".
  Five other jobs in `ci.yml` install packages without these two settings.
  Your fix for the RF-DETR job is right, and the other five are worth a separate issue.
- Item 5: I called CodeRabbit's `accelerator=` edit redundant, but it does change one thing.
  RF-DETR 1.10.1 reads `TrainConfig.accelerator` in `module_data.py:180-186` to default `pin_memory`.
  So on a CUDA host, `--accelerator cpu` leaves memory pinning on for a CPU run.
  The device Lightning trains on is unaffected, so the edit stays optional.
- Found-fine list: moving the forward wrap to `on_train_start` is correct, but it is not what protects the RF-DETR path.
  RF-DETR's EMA copies the model in its `setup()`, and the adapter appends its callback after that one (`rfdetr.py:142`).
  The move matters for plain Lightning users who list `TraceMLCallback` before an EMA or SWA callback.

6. **CodeRabbit's new `_warn` comment holds, and it has a twin.**
`_log_lightning_error` prints to stderr outside its `try` block at `lightning.py:150`.
The same guard belongs in both places.

PR #469, opened 2026-09-18
Add reproducible RF-DETR Nano T4 training case study

abhinavsriva:
## What this adds

This PR adds a reproducible RF-DETR Nano training case study for [RF-DETR issue #1410](https://github.com/roboflow/rf-detr/issues/1410).

The study measures eager training with real COCO train2017 batches on one Tesla T4 and four T4s using single-node DDP. It records native throughput, TraceML phase timing, criterion and matcher attribution, run overhead, resolved
configuration and environment metadata.

The completed measurements found:

- One T4: 18.56 images/s at a median 215.5 ms/step.
- Four-T4 DDP: 60.53 images/s at a median 264.3 ms/step.
- Fixed per-GPU batch weak scaling: 3.26x throughput and 81.5% efficiency.
- Exposed input wait remained below 0.25 ms/step.
- Backward increased from 79.3 ms to 114.7 ms under DDP, while forward and  optimizer timing remained comparatively stable.
- The four ranks remained closely aligned; the largest within-run step-envelope  difference was 1.6 ms.
- A separate PyTorch Profiler run measured criterion at 37.0 ms CPU duration /  32.1 ms attributed CUDA time per active step, with the nested matcher scope at  20.4 ms / 15.8 ms.

These results complement the existing L4 and RTX 4060 measurements in the upstream issue. Absolute timings are not directly compared because those runs used synthetic batches, another RF-DETR revision and different precision.

## Scope

This PR covers:

- RF-DETR Nano at resolution 384
- Real COCO train2017 batches
- Eager FP16 training on Tesla T4 GPUs
- Single-GPU and four-GPU single-node DDP
- Batch 4 per GPU
- Fifty optimizer-step attempts with steps 1–10 discarded
- CUDA-synchronized wall throughput over steps 11–50
- Asynchronous CUDA-event phase timing
- A separate PyTorch Profiler run using `wait=20, warmup=5, active=10`
- Three matched native/traced pairs per topology
- Reproducibility and report-validation checks

## Out of scope

This PR does not evaluate:

- `torch.compile`
- CUDA graphs
- A100 or other GPU models
- Multi-node training
- Segmentation or keypoint training
- Training accuracy or mAP
- Any RF-DETR optimization or change to RF-DETR defaults

Any later optimization claim still requires before/after accuracy evaluation on the same validation set.

### Case-study runner

`examples/case_studies/rfdetr_nano_training/train.py`:

- Pins RF-DETR to `0ed5be8e8d6762c4978a11671cbf34cfc0595e25`.
- Uses RF-DETR’s native Nano configuration, training module, data module and  trainer factory.
- Preserves the native model, loss, optimizer, scheduler, EMA and callback  behavior.
- Supports one GPU and ordinary single-node DDP.
- Measures a CUDA-synchronized wall-clock window after warm-up.
- Records source revisions, resolved configuration, package environment, GPU and  CPU information, annotation checksums and checkpoint identity.
- Provides a separate PyTorch Profiler mode with criterion and matcher scopes.
- Keeps profiler timings separate from benchmark and TraceML-overhead runs.

### Report generation

`examples/case_studies/rfdetr_nano_training/summarize.py`:

- Validates paired native and traced configurations before comparison.
- Rejects missing ranks, incomplete step windows and reused run directories.
- Reads TraceML timing for exactly steps 11–50.
- Uses the slowest rank’s elapsed window for DDP throughput.
- Produces native throughput, per-rank phase timing, profiler attribution,  configuration differences and paired overhead.
- Preserves residual as unassigned time rather than presenting it as  criterion/matcher cost.
- Reads CPU annotation duration from the raw profiler trace when required by  PyTorch 2.9.

### Case-study documentation

`examples/case_studies/rfdetr_nano_training/README.md`:

- Presents the completed single-T4 and four-T4 results.
- Documents the environment, timing boundaries and measurement limitations.
- Provides installation, COCO setup, single-GPU, DDP, profiler and report  commands.
- Identifies the four-GPU comparison as weak scaling because the global batch  increases from 4 to 16.

`examples/case_studies/README.md` adds the completed RF-DETR study to the case-study index.

### Integration documentation

The RF-DETR integration guide and generated support matrix now:

- Present DataLoader comparison as an example rather than the integration’s  purpose.
- Separate supported modes from internal validation details.
- Link to the completed T4 case study.
- Record real single-GPU CUDA and four-GPU NCCL evidence without treating it as  CUDA CI or broad compatibility coverage.
- Use concise installation and version-warning language independent of the case  study.

No TraceML public API or RF-DETR training behavior changes in this PR.

### Tests

`tests/integrations/test_rfdetr_case_study.py` covers:

- Warm-up exclusion and exact measured-step windows
- Slowest-rank DDP wall-time selection
- Missing and incomplete rank data
- Mismatched comparison configurations
- Profiler scope validation and CPU-duration fallback
- Source identity and environment recording
- Native Nano smoke training
- Callback and EMA preservation
- Matcher fast and fallback paths
- Report generation

## Validation

- RF-DETR and case-study tests: 81 passed, 5 skipped
- Three single-T4 native/traced pairs completed
- Three four-T4 DDP native/traced pairs completed
- All four DDP ranks completed 50 optimizer steps with finite loss
- PyTorch profiler exported ten active criterion/matcher steps
- `python tools/integration_support_matrix.py --check`
- `mkdocs build --strict`
- `git diff --check`

## Summary by CodeRabbit

- **Documentation**
  - Updated RF-DETR support details with one- and four-GPU validation findings, timing coverage, tested versions, and clarified limitations.
  - Added comprehensive RF-DETR Nano training case study documentation, including throughput, scaling, timing results, setup, reproduction steps, and scope limitations.
  - Refined case-study guidance and examples.

- **New Tools**
  - Added reproducible training and reporting utilities for validating runs, analyzing profiler output, and generating Markdown reports.

- **Bug Fixes**
  - Clarified RF-DETR installation and version-validation messages.


Pendu:
Reviewed `43f0ff89` against `main` `27b3fca5` (9 files, +1799/-53, 0 behind).
CI is green.
Most of the #462 follow-ups landed before that merge: warn-and-continue instead of raising, the `ddp_find_unused_parameters_*` aliases, and the narrowed `test-rfdetr` job.
The measurement design here is careful, and I have not second-guessed the measured numbers.
The items below are about whether the code, the tests and the write-up say the same thing.

1. **The profiler's "CUDA time" column is probably the GPU-side span of each scope, not attributed kernel time.**
`train.py:151` builds `{event.key: event for event in profiler.key_averages()}`, keyed by name only.
In torch 2.9.1, `key_averages()` also groups by `device_type` and `is_user_annotation` (`torch/autograd/profiler_util.py:330-336`).
So a `record_function` scope yields two entries with the same name, the CPU annotation and the GPU-side annotation, and the dict keeps whichever comes last.
The comment at `train.py:170-171` says `cpu_time_total` is zero for these scopes.
In 2.9.1 that only happens for a non-CPU entry (`profiler_util.py:584-588`), so the entry kept is the GPU-side one.
For that entry, `device_time_total` returns `time_range.elapsed_us()`, the span from first to last GPU activity (`profiler_util.py:599-618`), idle gaps included.
The CPU entry's `device_time_total` is the one that sums kernels (`profiler_util.py:602-607`).
This matters most for the matcher row: with the SciPy fallback running on the host, its GPU span includes the time the GPU waits on SciPy.
The column is labelled "attributed CUDA kernel time" (`summarize.py:149`) and "Attributed CUDA time/step" (`README.md:47`).
You can check this on the traces you already have.
Compare `cuda_total_ms` in `profile-summary-rank-0.json` against the summed `dur` of the `gpu_user_annotation` events named `rfdetr/matcher` in `profile-rank-0.json`.
If they match, the fix is to select the entry with `device_type == DeviceType.CPU` for the kernel column, or to relabel it as GPU span.

2. **The Reproduce section fails when run top to bottom.**
The DDP block (`README.md:178-182`) sets `BATCH` to `t4-ddp-01` and exports `CUDA_VISIBLE_DEVICES=0,1,2,3`.
The single-GPU profiler and `summarize.py` block comes after it (`README.md:189-199`), so the profiler run lands in the DDP directory with four visible devices.
`train.py:512-514` records `cuda_visible_devices`, and `summarize.py:138` compares everything except `mode`, `trace_db` and `output_dir`.
So `summarize.py` stops with "Profiler run differs from the benchmark configuration".
Resetting `BATCH` alone is not enough; `CUDA_VISIBLE_DEVICES` has to be reset too.
`README.md:185` also refers to "the same `summarize.py` command" before that command appears.
Moving the profiler and report block into the single-GPU flow, and giving DDP its own self-contained block, would fix both.

3. **The TraceML revisions in the Setup table will not exist on `main`.**
`README.md:99` pins TraceML to `622399c` and `8549e91`, both commits on this branch.
PRs here are squash-merged (#462 landed as the single-parent `27b3fca5`), so neither SHA will be reachable from `main`.
I checked that the timing code is identical between those two commits and this head; only messages and docs changed.
Pinning to the merge commit after merge would keep the study reproducible from a plain clone.

4. **Every traced case-study run prints a warning that contradicts the guide.**
The pinned RF-DETR commit reports version `1.11.0.dev0` (its `pyproject.toml` at `0ed5be8`).
`rfdetr.py:211` warns for anything other than `1.10.1`, and this PR changes the text to "has not been validated with this integration. Tested versions are listed in docs/user_guide/integrations/rfdetr.md".
That page (`rfdetr.md:125-129`) lists `0ed5be8` as the revision the case study exercised.
The previous wording acknowledged the development commit; either keep that, or note in the case-study README that the warning is expected.

5. **The "do not add H2D to the total" note contradicts the phase tables.**
`README.md:41-43` and the generated report (`summarize.py:277-278`) say host-to-device work can overlap other regions and should not be added to the total.
TraceML computes residual after subtracting H2D (`src/traceml_ai/step_time/analysis.py:374-378`), and in this integration H2D is timed inside Lightning's `batch_to_device`, before forward runs (`lightning.py:403-412`).
The README's own rows add up only with H2D included.
Single T4: 44.3 + 79.3 + 56.2 + 39.7 + 0.23 + 1.57 = 221.3 ms, the envelope.
Four-T4 DDP: 46.7 + 114.7 + 56.6 + 49.7 + 0.24 + 1.63 = 269.57 ms, against 269.6 ms.
Without H2D a reader gets 219.73 ms and goes looking for 1.57 ms that is not missing.

6. **The five tests that need rfdetr run in no CI job.**
`test-integration` collects `tests/integrations` (`ci.yml:137`) but does not install rfdetr, so `test_rfdetr_case_study.py:336`, `:425` and `:480` skip.
`test-rfdetr` installs rfdetr but runs only `test_rfdetr_trainer.py` and `test_rfdetr_nano.py`.
Those five are the only tests of `criterion_scopes`, `export_profile`, `config_changes` and the Nano smoke run.
The smoke run is also the only check that a `--profile` run installs no TraceML callback and publishes no steps (`test_rfdetr_case_study.py:584-590`).
The PR body's "81 passed, 5 skipped" is exactly what `test_rfdetr.py`, `test_rfdetr_example.py` and this file give with rfdetr not installed; I got the same count here.
With rfdetr importable this file cannot produce five skips.
Did those five pass against `0ed5be8`?
The profiler test also needs `_match_many`, which 1.10.1 lacks, so it would still skip if the file were added to `test-rfdetr`.

7. **Three published quantities are not pinned by any test.**
The slowest-rank rule: the fixture makes `elapsed_s` grow with rank (`test_rfdetr_case_study.py:103`), so replacing `max(...)` at `summarize.py:107` with `ranks[-1]["elapsed_s"]` still passes all 20 runnable tests.
Images/s: no test asserts that column, so using the per-GPU batch at `summarize.py:271` would also pass.
The window boundaries: the fake clock at `test_rfdetr_case_study.py:452-453` returns 10.0 and 14.0 whenever it is read.
So starting the clock after step 11, or stopping it after step 49, still gives `elapsed_s == 4`, and I confirmed both mutations pass.
Either one would time 39 steps and divide by 40.
Making a middle rank the slowest, asserting one Images/s cell, and recording `global_step` inside the fake `perf_counter` would close all three.

8. **`summarize.py` accepts a repeat whose phase timings fell back to the CPU clock.**
The analyzer switches the whole window to the CPU clock if any rank-step row lacks a GPU value (`analysis.py:277-286`).
`summarize.py:83-100` checks only that the phase fields exist, and prints the clock without checking it (`summarize.py:303`).
The README averages repeats by hand and describes GPU event timing (`README.md:110-112`).
One unresolved CUDA event would mix host-clock and GPU-clock means into the published table.
Rejecting a CPU-clock window when the ranks ran on CUDA would prevent that.

9. **"TraceML added a median 5.6 ms/step (2.55%)" cannot be rederived.**
`summarize.py:237` reports overhead only as a per-pair percentage, and the report never prints an ms delta.
2.55% of the three native wall times in the table (214.5, 215.5, 217.6 ms) is 5.47 to 5.55 ms.
Both can hold only if they come from pair 3 with a delta between 5.55 and 5.56 ms, at the edge of display rounding.
A reader has no way to confirm that from the README or the report.
Adding the per-pair ms delta to the overhead table would make the figure reproducible.

10. **Small ones.**
`README.md:8` says "about 18.5 images/s", but the median is 18.56, which rounds to 18.6; 3.26x and 81.5% use 18.56.
The support matrix drops "CUDA memory require[s] GPU validation" (`integration_support.json:150`), but the case study does not look at memory, so that caveat is still true.
COCO is fetched over plain HTTP with no checksum (CodeRabbit's point).
`https://s3.amazonaws.com/images.cocodataset.org/...` serves the same files over valid TLS, and `run.json` already hashes the annotation files, so publishing those hashes in the README would let a reproducer confirm the data.
`train.py:331` checks `TRACEML_SESSION_ID` but `train.py:464` reads `TRACEML_LOGS_DIR` after the model and trainer are built (CodeRabbit's other point); `traceml run` sets both, so this only bites a manual launch.

Checked and found fine:

- The measurement window covers steps 11 to 50 on every rank, with CUDA synchronization at both ends and a barrier at the start (`train.py:203-238`).
- The window callback is appended after the TraceML callback (`train.py:456`), so traced runs pay TraceML's per-step work inside the window and native runs do not.
- `build_trainer` is imported after `init()` (`train.py:389-395`), so the case study goes through the same patched factory as `model.train()`.
- The matcher call-count test pins single counting on both the batched and the fallback path.
- Pairing, reuse, missing-rank and incomplete-window checks reject bad input, and warm-up exclusion is well tested.
- 3.26x and 81.5% follow from 18.56 and 60.53 images/s, and every Images/s cell follows from its wall time and batch.
- The generated support matrix regenerates byte-identical, and no links to the old anchors remain.

abhinavsriva:
@pendu Thanks. I corrected the profiler labeling and H2D explanation, made the single-GPU and DDP reproduction flows independent, added verified HTTPS dataset downloads and earlier environment validation, exposed per-pair overhead deltas, and strengthened the lightweight tests for timing boundaries, throughput, slowest-rank selection and GPU-clock enforcement.

I did not add the full COCO/GPU case study to CI. The deterministic reporting and timing tests remain in CPU CI, while the pinned RF-DETR smoke checks and GPU benchmark remain manually validated. The published throughput and scaling results are unchanged.

PR #470, opened 2026-09-19
Capture pre-step H2D timing in Hugging Face Trainer

### LINKED ISSUE #451: Make the HF integration callback-only and fix training-input measurement

abhinavsriva:
## Summary

Closes #451.

This completes the Hugging Face Trainer integration work started in #452 and #459. It captures the H2D transfers that Accelerate performs before `on_step_begin` and prevents evaluation or prediction input work from leaking into training steps.

One completed optimizer group still produces exactly one TraceML step.

## How the timing works

Hugging Face collects the microbatches for an optimizer group through `Trainer.get_batch_samples`. Accelerate moves each prepared batch to the device during this collection, before the normal `trace_step` region opens.

    Trainer.get_batch_samples
      microbatch 1:
        DataLoader fetch                    -> Input Wait
        CPU-to-CUDA transfer:
          step_time
            h2d_time
              tensor.to("cuda")

      microbatch 2:
        DataLoader fetch                    -> Input Wait
        CPU-to-CUDA transfer:
          step_time
            h2d_time
              tensor.to("cuda")

    on_step_begin
      existing trace_step:
        forward
        backward
        collective and optimizer work

    on_step_end
      assign one step number
      publish the complete optimizer group

TraceML reuses its existing DataLoader timing scope:

- Training iterator calls temporarily enable DataLoader and H2D timing.
- Evaluation, prediction, and collection bookkeeping remain outside the timing scope.
- Direct `trainer.evaluate()` and `trainer.predict()` calls leave no pending input events.
- Multiple tensors and gradient-accumulation microbatches produce repeated events that the existing sampler sums under the same optimizer-step number.

The short H2D segments do not advance the step counter or publish independently.

## Accounting

The existing formulas remain unchanged:

    Traced Step Time =
        pre-step H2D step-time segments
        + forward/backward/optimizer trace_step

    Step Time =
        Input Wait
        + Traced Step Time

H2D is included in Traced Step Time once and subtracted from residual once. No sampler, diagnosis, schema, storage, or frontend conditional was added.

## Input Wait attribution

Evaluation and prediction DataLoader fetches are no longer carried into subsequent training steps. Existing Hugging Face Input Wait values may therefore decrease for runs that evaluate during training.

## Compatibility

- Standard Hugging Face `Trainer` is supported.
- Gradient accumulation and partial final groups retain one TraceML step per optimizer group.
- CPU, D2D, dtype-only, and model-parameter `.to()` calls remain excluded.
- Transformers versions without `get_batch_samples` warn once and continue without training Input Wait or pre-step H2D.
- Trainer subclasses overriding `get_batch_samples` warn once and continue without those input signals rather than reporting partial measurements.
- Standalone Accelerate remains out of scope. Its documented workaround uses `device_placement=False` and moves the batch inside `trace_step`.

## Validation

- [x] Training collection ordering with and without gradient accumulation
- [x] One published TraceML step per optimizer group
- [x] Evaluation and prediction input isolation
- [x] Standalone `evaluate()` and `predict()` isolation
- [x] Collection bookkeeping exclusion
- [x] Repeated H2D and step-time aggregation
- [x] Residual accounting
- [x] Partial and failed-group cleanup
- [x] Duplicate callbacks and repeated initialization
- [x] Unsupported-version and Trainer-override warnings
- [x] Full CPU test suite
- [x] Strict documentation build

GPU validation is not part of this PR’s automated validation.


Pendu:
Reviewed `f4867f25` against `main` `27b3fca5` (9 files, +792/-90, 0 behind).
CI is green and the 64 tests in the three changed test files pass locally.
The mechanism is sound.
`begin_step_capture()` returns the active capture rather than a fresh one (`instrumentation/step_events.py:186-188`), so the pre-step transfer events are adopted by the later `trace_step`.
The sampler already sums repeated events by name (`samplers/step_time_sampler.py:100`).
The items below are about the edges: what happens when the collection hook is not in play, and when the transfer happens inside the fetch.

1. **When the hook does not apply, Input Wait disappears and takes the published Step Time with it.**
The lifecycle guard now holds DataLoader timing disabled for the whole inner loop (`huggingface.py:419`), and only the iterator proxy re-enables it.
So a Trainer that overrides `get_batch_samples`, or any transformers before 4.46, records no fetch events at all.
I ran a real CPU Trainer for three steps on both trees and counted published events per step:

| Trainer | `dataloader_next` at `27b3fca5` | at `f4867f25` |
|---|---|---|
| stock | 2, 1, 1 | 2, 1, 1 |
| subclass delegating to `super()` | 2, 1, 1 | 2, 1, 1 |
| subclass with its own collection loop | 2, 1, 1 | 0, 0, 0 |

That last row matters more than it looks.
`step_time = input_wait + traced_step_time` only when both are present (`step_time/analysis.py:379-383`), so with no fetch events the analyzer returns `input_wait_ms = None` and `step_time_ms = None`.
I confirmed that by running the production analyzer over both shapes.
Those users go from an Input Wait that was sometimes polluted by evaluation fetches to no Input Wait and no Step Time.
The docs say TraceML "continues without training Input Wait or pre-step H2D signals" (`huggingface.md:125-128`, `step-time-pipeline-contract.md:341-345`), which understates it.
Would falling back to the old always-on scope when the hook is absent be better than dropping the headline metric?

2. **The new bypass warning can abort `trainer.train()`.**
`_warn_if_training_batch_timing_is_bypassed(trainer)` is called at `huggingface.py:406` with no try/except, and it ends in `logger.warning` (`huggingface.py:108`).
`logging.Handler.handle` does not shield `emit`, so a handler that raises propagates out.
I reproduced it end to end with a real CPU Trainer on transformers 5.9.0.
With a subclass overriding `get_batch_samples` and a handler whose `emit` raises, `trainer.train()` exits with `RuntimeError: third-party log handler blew up` before the first step.
This is the only new instrumentation call in the file that is not wrapped, and `integrations/_capability.py:38-60` already wraps its own `logger.warning` in a try/except.
Same conclusion as the automated reviewer's comment, with a repro.

3. **A transfer performed inside the fetch is counted twice.**
The proxy arms both scopes over the same span (`huggingface.py:277-279`), so anything that runs inside `next()` is inside the `dataloader_next` region as well.
That is fine for Accelerate, which places the batch after the fetch returns.
It is not fine for a `collate_fn` or a Dataset that moves the batch to the device itself, which runs inside `next()`.
I armed the H2D predicate and ran one fetch through the proxy with a device-moving collate, on both trees:

```
base   dataloader_next  start=...803866 end=...806046
head   h2d_time         start=...162961 end=...162971
head   step_time        start=...162957 end=...162977
head   dataloader_next  start=...162839 end=...162997
```

At head the `step_time` segment is nested inside the `dataloader_next` region.
The transfer therefore lands in Input Wait and in Traced Step Time, and `step_time = input_wait + traced_step_time` adds it twice.
At base neither the H2D event nor the segment existed, so there was nothing to double count.
Arming the step-time segment only when the transfer is not already inside a timed fetch region would close it.
The PR's "included once" wording would then hold for every path.

4. **The warning is a false positive for a subclass that delegates.**
`_warn_if_training_batch_timing_is_bypassed` only checks whether the bound method carries `_traceml_training_batch_timing`, so a subclass whose `get_batch_samples` just calls `super()` trips it.
In the table above that subclass still publishes 2, 1, 1 fetch events, while the warning tells the user the signals are gone.
Wrapping and delegating is a common pattern.
Softening the message to "cannot confirm", and saying a delegating override still works, would fit.
The two docs pages state the loss as certain and would need the same change.

5. **The Accelerate workaround as written raises `TypeError`.**
`accelerate.md:96-97` says to "prepare the DataLoader with `device_placement=False`".
The guide's own example is `accelerator.prepare(model, optimizer, dataloader)` (`accelerate.md:34`).
`Accelerator.prepare` is `(self, *args, device_placement=None)` and indexes that argument per positional arg.
Against accelerate 1.13.0: `prepare(dl, device_placement=False)` gives `TypeError: object of type 'bool' has no len()`, while `device_placement=[False]` works.
`Accelerator(device_placement=False)` is the bool form.

6. **The user-visible metric change is only in the PR description.**
Existing HF runs that evaluate during training will report lower Input Wait, and Step Time now includes pre-step transfers.
Neither `huggingface.md` nor the contract doc tells a user their numbers will move, and `SCHEMA.md:3` still reads `1.8`.
Worth a sentence where users will see it when comparing a new summary against an old one.

7. **Two gaps in otherwise strong test coverage.**
The negative control at `test_hf_h2d_window.py:324` asserts that `"cannot guarantee pre-step H2D timing"` is absent.
The emitted message is "cannot guarantee training Input Wait or pre-step H2D timing", and that exact substring appears nowhere in the source.
So the assertion holds whether or not the warning fires.
Separately, replacing the call at `huggingface.py:406` with `pass` leaves 71 tests green across the three HF files.
Nothing pins that the guard warns at all.
Asserting `caplog.records == []` for the first, and adding a test for the second, closes both.

8. **Small one.**
These capability warnings go through `logging` while every other message in the module uses `print(..., file=sys.stderr)` with a `[TraceML]` prefix (`_log_hf_error`, `huggingface.py:100`).
A user who routes logging to a file or raises the level past WARNING silently loses the one message telling them Input Wait is missing.

Checked and found fine:

- For the Accelerate path there is no double counting.
  The DataLoader patch times only `next(it)` and yields the batch after leaving that region (`dataloader_patch.py:118-140`), so Accelerate's device placement happens outside it.
  Item 3 is the case where that does not hold.
- `get_batch_samples` only ever calls `next(epoch_iterator)` in both transformers 4.46.0 and 5.9.0, so the two-method proxy is enough.
  `_get_num_items_in_batch` stays outside it, as the contract says.
- The documented `transformers>=4.46` floor is right: the method is absent at v4.45.0 and present at v4.46.0.
- `n_calls` is written to SQLite but read by no analysis, diagnosis or renderer code, so more than one `step_time` event per step breaks no consumer.
- The stock Trainer path is unchanged between the two trees, measured above.
- `tests/test_h2d_timing.py` does observe the real emitted segments, including ordering, the raising-transfer path, and that they land in the same pending capture.
- Every `h2d_auto_timer` call site constructs a fresh instance, and Lightning keeps the default `include_step_time=False`, so its behaviour does not change.
- The repro script fix is real: it read `evt.gpu_ms`, which does not exist, so the positive control was always reporting 0.0.

abhinavsriva:
Thanks, I addressed the actionable items:

- Warning delivery is fail-open, including application logging-handler failures.
- The HF extra now requires `transformers>=4.46.1`; older manual installations warn and omit unsupported Input Wait/H2D measurements.
- Standalone Accelerate’s limitation and manual-device-placement workaround are documented.
- Collection-side bookkeeping limitations and H2D context state handling are documented.
- Focused CPU tests pass on Transformers 4.46.1 and 5.17.0, and the strict documentation build passes.

Not completed in this PR: real CUDA validation and CUDA-event overhead measurement. These require a GPU and will be handled separately; no GPU CI workflow is being added.

PR #471, opened 2026-09-20
Dev/abhinav/hf resume attribution

abhinavsriva:
## Summary

This PR fixes incorrect input attribution when resuming a Hugging Face `Trainer` from a checkpoint.

Closes #451.

When an iterable dataset resumes, Accelerate may fetch the already-completed batches from the checkpoint and the first new training batch in the same loader call. TraceML cannot reliably separate those two pieces of input work.

Without this change, that skipped input work could appear as input wait or H2D time for the first real resumed step.

## What changes

For standard Hugging Face `Trainer` with `transformers>=4.46.1`:

- Map-style datasets continue to record normally after resume because their completed batches are skipped by the sampler before they are fetched.
- Iterable datasets detect Accelerate's lazy skip path.
- TraceML omits the first resumed optimizer group when its input work contains both skipped and real batches.
- Training behavior does not change: forward, backward, optimizer updates, checkpoint restoration, and HF step counting continue normally.
- TraceML resumes recording from the following optimizer group.
- Failed attempts and retries clear the temporary resume state safely.

This avoids publishing a misleading partial step or treating unavailable input timing as zero.

## Why one group is omitted

The first resumed iterable group can look like this:

    Accelerate iterator call
      checkpoint-skipped batches
      first real training batch
      return first real batch to Trainer

    Trainer callback
      forward
      backward
      optimizer update

The loader does not expose a reliable boundary between the skipped input work and the first real batch. Rather than assigning all of that time to the real step, TraceML drops that one ambiguous group.

The next group is normal:

    next input fetch + H2D
    forward + backward + optimizer
    publish one TraceML step

## Version support

`skip_first_batches` is existing upstream behavior; it has been present in Transformers since the 4.27 era and Accelerate since 0.16.

This integration supports `transformers>=4.46.1` because that is when `Trainer.get_batch_samples` provides the collection boundary needed to time training input correctly.

## In scope

- Standard Hugging Face `Trainer`
- Checkpoint resume
- Iterable and map-style datasets
- Gradient accumulation
- `ignore_data_skip=True`
- Failure and retry cleanup
- CPU validation of ordering, grouping, state cleanup, and accounting

## Out of scope

- Standalone Accelerate training loops
- Custom Trainers that replace `get_batch_samples`
- TRL or other Trainer subclasses with custom collection loops
- Changing checkpoint restore behavior
- Changing diagnosis, storage, telemetry schema, or frontend logic
- CUDA timing-value validation; this remains a separate GPU validation step

## Validation

- [x] Iterable resume omits exactly one ambiguous optimizer group
- [x] Map-style resume continues to record normally
- [x] Gradient accumulation 1 and 2
- [x] `ignore_data_skip=True`
- [x] Failed resumed attempts do not leak timing or memory telemetry
- [x] Resume state resets before a later retry
- [x] Current HF suite: 105 passed
- [x] Supported floor, Transformers 4.46.1: focused suite passed
- [ ] GPU run for CUDA timing values, separate from resume-boundary correctness


Pendu:
Reviewed `0f21e23d` against `version_0.4.1` `2dc86368` (5 files, +390/-5, 0 behind).
No prior review to carry over: CodeRabbit skipped this PR because auto reviews are off for non-default base branches, and there are no formal or inline reviews.
The 43 tests in the two HF test files pass locally, and lint and the ownership check are clean.

The premise holds and I could measure it.
Running the same iterable resume on both trees, the first resumed group at `2dc86368` carries three `dataloader_next` events against one in steady state:

| tree | published rows | step 1 `dataloader_next` | step 2 |
|---|---|---|---|
| `2dc86368` | 2 | n=3, [0.063, 0.018, 0.014] ms | n=1, [0.018] ms |
| `0f21e23d` | 1 | n=1, [0.019] ms | n/a |

Three is exactly what `accelerate/data_loader.py:568-601` predicts: `DataLoaderShard.__iter__` reads one batch ahead, so the first consumer `next()` performs `skip_batches + 2` underlying fetches.
The detection is also right for the shapes a real Trainer builds.
A CPU Trainer gives `DataLoaderShard` with `skip_batches=0` for map-style.
It gives `DataLoaderDispatcher` with `skip_batches=1` for iterable.
Both match the branch at `accelerate/data_loader.py:1397-1402` and 1424-1438.

Items below.

1. **The `_abort_pending_capture_safely()` call at `huggingface.py:260` is load-bearing and nothing pins it.**
Replacing it with `pass` leaves all 43 tests green.
It is not dead, though. With that line removed, the same iterable resume publishes `optimizer_step` twice in the next step:

```
head     step=1  optimizer_step  n=1  ms=[0.029]
mutant   step=1  optimizer_step  n=2  ms=[0.047, 0.028]
```

The omitted group's optimizer event survives into the following capture, and `samplers/step_time_sampler.py:100` sums repeated events by name, so the optimizer phase would be double counted.
Asserting exactly one `optimizer_step` in the first recorded group would close it.
The other five mutations I tried were all caught, so this is the only gap of its kind.

2. **A resumed run whose only remaining group is the omitted one publishes nothing, silently.**
Checkpoint at step 1, resume with `max_steps=2`:

```
2dc86368   HF global_step=2; TraceML published 1 step row  -> [1]
0f21e23d   HF global_step=2; TraceML published 0 step rows -> []
```

HF ran a full optimizer update and TraceML reports no step at all, with no message on either stream.
This module already has the opposite convention: when the collection seam is missing it warns once and says which signals are gone (`huggingface.py:501-507`).
The resume path at `huggingface.py:391-447` has no `warn`, `logger` or `print` anywhere in it.
A one-time warning naming the omitted group would make a resumed summary readable without the docs.

3. **`tracked_skip_first_batches` is the one new runtime path with no try/except, and it pins a third-party signature.**
`huggingface.py:436-444` runs three TraceML expressions around the real call with no guard.
`_loader_skips_lazily` has an internal `try`, but the `num_batches <= 0` comparison at `huggingface.py:393` sits above it, so a non-int raises straight into HF's resume:

```
_loader_skips_lazily(obj, None) -> TypeError: '<=' not supported between instances of 'NoneType' and 'int'
```

No current Transformers version passes a non-int, so this is hardening rather than a live bug.
The installation is already guarded at `huggingface.py:89-92`, but the per-resume call path is not.
That is the same shape as the unguarded warning that could abort `train()` in #470.
The signature `(dataloader, num_batches=0)` matches `accelerate/data_loader.py:1385` exactly today; `*args, **kwargs` passthrough plus a try/except would make both concerns go away.

4. **Question: omitting the whole group also discards its real compute and memory.**
The group's forward, backward, optimizer and memory work is measurable and correct; only the input span is ambiguous.
Dropping the entire group is the conservative choice and the docs say so plainly.
It still costs a real step.
`training_total_steps` (`reporting/schema.py:114`) is defined at `SCHEMA.md:17` as the highest observed completed-step counter.
After a resume it therefore reports one fewer step than HF executed.
Was recording the group with Input Wait and H2D absent, and only those absent, considered?
That keeps the compute row and leaves the unmeasurable part unavailable rather than dropping everything.
If dropping the group is the intended answer, item 2 is the thing that makes it safe to live with.

5. **`test_resume_skip_tracking_distinguishes_accelerate_loader_shapes` does not exercise Accelerate's loader shapes.**
The test at `test_hf_h2d_window.py:285` builds raw `DataLoader` objects, which take the plain branch at `accelerate/data_loader.py:1451-1456`.
A real Trainer produces `DataLoaderShard` and `DataLoaderDispatcher`, handled by the different branches at 1424-1450.
The two end-to-end resume tests do cover the real shapes, so the behaviour is pinned; it is the characterization test that is not testing what its name and docstring claim.
Building the loaders through `Accelerator().prepare(...)` would make it match.

6. **Carried over from #470, still open in this file.**
`test_hf_h2d_window.py:434` asserts `"cannot guarantee pre-step H2D timing"` is absent.
The emitted message at `huggingface.py:503-505` is "cannot guarantee training Input Wait or pre-step H2D timing", and the asserted substring appears zero times anywhere in `src/`.
The assertion holds whether or not the warning fires.

Checked and found fine:

- Two levels of `.dataloader` descent are enough for every shape `skip_first_batches` returns.
  `accelerate/data_loader.py:1393` unwraps exactly one XLA level on the way in and 1459 rewraps exactly one on the way out.
- Nested attempts cannot lose a pending suppression.
  `huggingface.py:535-539` saves all three flags before overwriting them and 558-562 restores them in the `finally`.
- One group is the right amount to omit.
  `SkipDataLoader.__iter__` and the `DataLoaderShard` loop both consume every skipped batch inside the first consumer `next()`, and the look-ahead means the following group is already in steady state.
- `get_batch_samples` is called once per accumulation group in both 5.9.0 (`trainer.py:1700-1718`) and 4.46.1, so one flag per group is sound.
- Map-style resume still records its first group, measured on a real Trainer at both ga=1 and ga=2.
- The documented floor matches the packaging: `pyproject.toml:82` pins `transformers>=4.46.1` and every doc mention agrees.
- The failure and retry cleanup works as described; forcing `training_step` to raise leaves all three flags false and publishes nothing.

Pendu:
One more parametrization, from a second pass over the test matrix. Not a bug report.

`test_hf_h2d_window.py:549-550` covers `(1, False, False)`, `(2, False, False)` and
`(1, True, True)`, so `ignore_data_skip=True` is only exercised at
`gradient_accumulation_steps=1`. The uncovered combination is gradient accumulation
with the skip disabled, where the first resumed group holds several real microbatches
and nothing should be suppressed. That is the input shape that would catch a
suppression decision keyed off the wrong thing.

I added `(2, True, True)` to the list and ran it against `0f21e23d`:

```
2 passed, 6 deselected in 11.79s
```

So the behaviour is already right. The case is just unpinned, and one tuple is cheap
insurance for the branch that decides whether to suppress at all.

abhinavsriva:
@Pendu  Thanks. Omitting the complete first resumed iterable group is intentional: Accelerate combines checkpoint-skipped batches with the first real batch inside one iterator call, so TraceML cannot establish a reliable attribution boundary. We therefore discard that entire ambiguous group and resume normal recording with the following group, rather than publishing incorrect attribution or zero-valued measurements. This behavior is documented.

I added the optimizer-event leakage assertion, the `gradient_accumulation_steps=2` plus `ignore_data_skip=True` case, corrected the ineffective warning assertion, and updated the loader-shape test to exercise the actual `DataLoaderShard` and `DataLoaderDispatcher` shapes produced through HF’s internal Accelerate path. All 45 focused HF tests pass. The signature concern is future hardening rather than a current supported-path bug. This is ready to merge.

Issue #468, opened 2026-09-18
RFC: MONAI integration through the Ignite engine's events, delivered as four PRs


Pendu:
TraceML has no MONAI integration yet. This issue is the plan. Each part below is its own issue and its own PR against `main`.

## Why

A MONAI user can already wrap each iteration in `trace_step`, through `iteration_update=` or a pair of handlers on `ITERATION_STARTED` and `ITERATION_COMPLETED`. That gets two things wrong. A step is one iteration, so under `accumulation_steps=2` every other step has no optimizer time. And Input Wait comes from the DataLoader fetch patch, which counts `ThreadDataLoader`'s producer thread instead of the training loop (#463). With the handler pair, an iteration-level `ValidationHandler` also lands inside the step.

MONAI's [fast training guide](https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_model_training_guide.md) offers caching datasets and `ThreadDataLoader` for slow data loading. Comparing Input Wait across runs shows whether one of them helped, so Input Wait has to be right on those loaders. The report in #263 came from a MONAI user trying to install TraceML, before its NumPy pin was lifted.

## Approach

`monai.engines.Workflow` subclasses `ignite.engine.Engine` (`monai/engines/workflow.py:47`). The integration is a MONAI handler, passed in `train_handlers=[...]`, that times each phase at one of these seams.

```
GET_BATCH_STARTED -> GET_BATCH_COMPLETED     Input Wait
prepare_batch, wrapped                       step opens, H2D on CUDA
engine.inferer(...), wrapped                 forward
LOSS_COMPLETED -> BACKWARD_COMPLETED         backward
hooks on engine.optimizer                    optimizer step
MODEL_COMPLETED                              step closes
```

The integration's own `init()` calls `traceml.init(mode="selective", patch_h2d=True)`. That arms H2D timing and never installs the DataLoader fetch patch. Input Wait comes from Ignite's bracket around `next()` on the training thread, so a producer thread is never counted. On a loader with no length, the first epoch ends with a `GET_BATCH_STARTED` that has no matching `GET_BATCH_COMPLETED` (`ignite/engine/engine.py:1213-1236`). The handler drops that unmatched start. If a script calls plain `traceml.init()` instead, the patch is installed and records fetches on top of the handler's own, twice per step on a plain `DataLoader`. The handler then warns once, naming the MONAI `init()`, and publishes no steps.

[...]

Issue #466, opened 2026-09-18
MONAI integration, part 3 of 4: guide, support-matrix row and minimal example


Pendu:
Part 3 of 4 for MONAI support, after #465.

**Goal.** Document what #464 and #465 measure, and ship the smallest runnable example.

**Scope**
- [ ] `docs/user_guide/integrations/monai.md`, with its `mkdocs.yml` nav entry and an entry in `docs/user_guide/public-api.md`
- [ ] The guide covers install and `init()`, which engines are traced, the seam for each phase, and the step under `accumulation_steps`.
- [ ] It says where MONAI's own work lands. Decollate, postprocessing and the loss sit inside the step in no phase, and `ITERATION_COMPLETED` handlers sit outside it.
- [ ] It says that a step the AMP scaler skips has no optimizer event, and that under plain `traceml.init()` the handler traces nothing.
- [ ] A support-matrix row: `cpu` and `single_process` are `CI tested` by the `test-monai` job, and `gpu`, `multi_process` and `multi_node` are `Not claimed`.
- [ ] `integrations.md` regenerated with `python tools/integration_support_matrix.py --write`
- [ ] A MONAI link in the README's Frameworks list (`README.md:351`), and a row in `examples/README.md`
- [ ] `examples/integrations/monai_minimal.py` with synthetic 3D volumes, a small UNet and a `SupervisedTrainer` carrying the handler, with no download
- [ ] The `test-monai` job runs the example under `traceml run` on CPU and checks that `final_summary.json` reports Input Wait, forward, backward and optimizer time.
- [ ] A CHANGELOG entry

**Size.** About 450 changed lines. The guide stays near the length of the Accelerate guide, which is 177 lines.

**Done means.** The guide and the matrix row describe exactly what #464 and #465 measure. The example runs on CPU in CI, and its summary shows all four phases. The matrix check passes.

Issue #465, opened 2026-09-18
MONAI integration, part 2 of 4: forward, backward and optimizer time


Pendu:
Part 2 of 4 for MONAI support, after #464.

**Goal.** Add forward, backward and optimizer time to the step from #464.

| Phase | Seam |
| --- | --- |
| Forward | Around the `engine.inferer(...)` call (`monai/engines/trainer.py:253`). This keeps `network.train()` and `optimizer.zero_grad()` (`:274-276`) out of forward. |
| Loss | No stream. The loss stays in the step and in no phase, as with the Lightning callback. |
| Backward | `IterationEvents.LOSS_COMPLETED` to `BACKWARD_COMPLETED` (`:256`, `:283`, `:291`) |
| Optimizer | Step pre- and post-hooks registered on `engine.optimizer` for the run, removed at its exit |

A prototype on `main` checked the forward window on a model whose `zero_grad` costs more than its forward. After the first step, forward read 0.11 to 0.22 ms per step. A bare forward took 0.10 ms, and `zero_grad` took 3.42 ms.

The optimizer hooks fire inside `scaler.step()` under AMP (`monai/engines/trainer.py:285`) and on `optimizer.step()` otherwise (`:293`). A step that the AMP scaler skips never calls `optimizer.step()` (`torch/amp/grad_scaler.py:371-372` in torch 2.12), so it records no optimizer event.

TraceML's global optimizer hooks are not used here. They time every optimizer in the process for the rest of its life. Installing them also makes `traceml.wrap_optimizer()` refuse (`src/traceml_ai/sdk/wrappers.py:71-76`). The event bracket from `BACKWARD_COMPLETED` to `MODEL_COMPLETED` is not used either. It also holds `scaler.update()`, and on an accumulating iteration it holds no optimizer step at all (`:284-286`, `:292-293`).

**Scope**
- [ ] A forward region around `engine.inferer`, installed at run start and restored at run exit only while it is still the handler's wrapper
- [ ] A backward region from `LOSS_COMPLETED` to `BACKWARD_COMPLETED`
- [ ] Optimizer step hooks on `engine.optimizer`, removed at every run exit. An optimizer already wrapped with `traceml.wrap_optimizer()` is not hooked again.
- [ ] `REQUIRED_STEP_TIME["monai"]` and a harness in `tests/integrations/test_telemetry_conformance.py`, owing step time, forward, backward, optimizer and Input Wait
- [ ] A CHANGELOG entry

[...]

Issue #464, opened 2026-09-18
MONAI integration, part 1 of 4: handler lifecycle, step and Input Wait


Pendu:
Part 1 of 4 for MONAI support. Each part lands after the one before it.

**Why.** MONAI's training loop is an Ignite engine. `monai.engines.Workflow` subclasses `ignite.engine.Engine` (`monai/engines/workflow.py:47`). MONAI adds its own iteration events (`monai/engines/utils.py:47`). `SupervisedTrainer` fires them in every iteration (`monai/engines/trainer.py:254`, `:256`, `:283`, `:291`, `:307`). A handler is a plain object with `attach(engine)`, passed in `train_handlers=[...]` (`monai/engines/workflow.py:269`). So a handler can time each phase from MONAI's and Ignite's events, wrapping a callable where no event exists. The report in #263 came from a MONAI user trying to install TraceML, before its NumPy pin was lifted. Line numbers are for MONAI 1.6.0 and Ignite 0.5.5.

**Design.** `init()` calls `traceml.init(mode="selective", patch_h2d=True)` and leaves the DataLoader fetch patch off. Input Wait comes from the engine's own `GET_BATCH_STARTED` to `GET_BATCH_COMPLETED` bracket. That bracket times the training thread's `next()` on the loader, so `ThreadDataLoader`'s producer thread is never counted (#463). Without the check below, a prototype under plain `traceml.init()` recorded two fetch events per step on a plain `DataLoader`. A prototype of this design matched a timer around `next()` on the training thread to within 1 ms, on both loaders, on Ignite 0.5.5 and 0.4.11.

The step opens inside a wrapped `prepare_batch`, the first statement of the iteration (`monai/engines/trainer.py:223`). So the transfer to the device is inside the step. The step closes at `IterationEvents.MODEL_COMPLETED` (`:307`). MONAI registers its decollate and postprocessing on that event before any user handler (`monai/engines/workflow.py:185-194`, `:202`, `:217`). They sit inside the step, in no phase. Handlers on `ITERATION_COMPLETED` run after the step has closed. Those include metric updates and an iteration-level `ValidationHandler`. That holds whatever order the handlers are listed in.

A TraceML step is one optimizer-update group, as in the Lightning callback (`src/traceml_ai/integrations/lightning.py:203`). Under `accumulation_steps` the group follows MONAI's own rule, including a partial final group (`monai/engines/trainer.py:258-272`).

[...]

PR #472, opened 2026-09-21
MONAI integration, part 1 of 4: handler lifecycle, step and Input Wait

Pendu:
Part 1 of 4 for MONAI support (#464), planned in #468.

MONAI's `SupervisedTrainer` is an Ignite engine, so this adds a MONAI handler.
Pass `TraceMLHandler()` in `train_handlers` and each training step is timed from
the engine's own events.

```python
from traceml_ai.integrations import monai as traceml_monai

traceml_monai.init()
trainer = SupervisedTrainer(..., train_handlers=[traceml_monai.TraceMLHandler()])
```

## What lands

- `traceml_ai.integrations.monai` with `init()` and `TraceMLHandler`.
- One TraceML step is one optimizer update, as in the Lightning callback, so a
  step spans several iterations under `accumulation_steps`.
- Input Wait comes from the engine's `GET_BATCH_STARTED` to
  `GET_BATCH_COMPLETED` bracket, timed on the training thread. `init()` leaves
  the torch DataLoader fetch patch off, so a loader that fetches on a background
  thread, such as `ThreadDataLoader`, is timed by what the loop waited rather
  than by what its producer did (#463).
- Step time, step memory, and H2D on CUDA.
- Only a `SupervisedTrainer` that runs `SupervisedTrainer._iteration` unchanged
  is traced, including a subclass that inherits it. `GanTrainer`,
  `AdversarialTrainer`, evaluators, `iteration_update=` and an overridden
  `_iteration` each get one warning and nothing attached.
- A `monai` extra, and a `test-monai` CI job that also reruns the file at the
  declared Ignite floor.

Forward, backward and optimizer time are #465. The guide, the support-matrix row
and the example are #466, and the CHANGELOG entry says so.

## Evidence

GATE: ran=true. `pytest tests/integrations/test_monai.py` is 43 passed on MONAI
1.6.0 with Ignite 0.5.5. The `test-monai` job on this head builds Python 3.10
with torch 2.14.0 and runs the file twice, once at the pinned versions and once
at the declared floor Ignite 0.4.11: 43 passed both times. `pytest
tests/integrations` is 264 passed, 8 skipped. The full suite is 1954 passed, 8
skipped, against 1911 passed, 8 skipped on `version_0.4.1` at 2dc86368. The 43
tests this branch adds are the whole difference.

TRANSITIONS: tested as A to B, not as single states. Not tracing to tracing to
cleanup, once per run. Group open to group published, and group open to group
dropped at a crash or an unfinished epoch. Iteration with a step to iteration
without one, which is a `prepare_batch` replaced mid-run and now warns.
Recording to draining when the step budget is spent. Kill switch off to on
during a run. First attach to second attach refused, per handler and per
trainer. Fetch start to fetch recorded, and fetch start to fetch dropped, which
is the unmatched start Ignite leaves at the end of a length-less loader's first
epoch. Capture owned to capture detached by another producer.

RANKS: not exercised and not claimed. The handler has no per-rank branch. Step
numbers come from the existing process-local session counter, so each rank
publishes its own exactly as the other integrations do. `multi_process` and
`multi_node` stay unclaimed until the support-matrix row lands in #466.

TRIPWIRE: every guard is fired on purpose by a test. The unsupported-engine
warning across 8 engine types, the refusal when a DataLoader fetch patch is
already installed, the H2D-not-armed warning on a CUDA engine, the mid-run
`prepare_batch` replacement, the second-handler refusal, a group restarted by
another producer, and an internal error forced at run start, at step open and
at run end. A mutation pass removed each promised behaviour in turn: 25 of 25
mutations were caught by at least one test, with the unmutated control green.

TWINS: searched the concept, not a symbol. The concept is a timing region opened
in one callback and closed in another, which is where an orphaned region comes
from. Other sites are `lightning.py:239`, `:300` and `huggingface.py:180`,
`:192`, and both are bounded by hooks their framework always pairs. Input Wait
is produced only by `dataloader_patch.py:131` and `wrappers.py:95`, each inside
a single `with`. No other integration reads a framework fetch bracket, so the
unmatched-start case has no twin.

End to end under `traceml run`, 32 iterations with `accumulation_steps=2` on a
`ThreadDataLoader` that cannot keep up: the summary reports 16 steps for the
trainer's 16 optimizer updates, and Input Wait of 120.53 ms per step, 1.9285 s
over the run, against the 1.9291 s the loop measured for itself.

## Review

An external model and five internal review passes produced 32 findings. Fixed
here: a test-isolation defect where a DataLoader patch left installed by an
earlier integration test made this file fail in a full-directory run; a guard
that was per handler instead of per trainer, so two handler instances doubled
every number silently; tests that counted events without ever reading a clock;
a fetch clock that kept taking CUDA events after the step budget was spent; a
`prepare_batch` replaced mid-run that stopped tracing in silence; a fail-open
read of `accumulation_steps`; a module-level `import torch`; an unpinned CI gate
whose declared Ignite floor was never run; and a group that could lose
measurements when another producer detached the shared capture. The rest were
refuted against sibling behaviour in `lightning.py` and `huggingface.py` or
filed as follow-ups.

## Summary by CodeRabbit

* **New Features**
  * Added MONAI integration for `SupervisedTrainer` workflows.
  * Captures optimizer-update steps, step duration, input wait time, memory usage, and CUDA host-to-device transfer timing.
  * Supports accumulation-step grouping and safely handles interrupted or incomplete captures.
  * Added optional MONAI dependencies for installation.

* **Documentation**
  * Added an unreleased changelog entry describing MONAI support and captured metrics.

* **Tests**
  * Added comprehensive integration coverage for MONAI training scenarios and supported loader configurations.


abhinavsriva:
@Pendu: The overall design looks appropriate for Part 1. Ignite’s fetch events measure the training thread’s actual wait, and no shared PyTorch DataLoader patch change is needed. The dedicated CPU integration job and Ignite-floor check also make sense.

Two small changes:
1. Fix unfinished accumulation crossing epochs. With a length-less loader, four batches per epoch, accumulation_steps=3, and two epochs, TraceML reports groups of [3, 4, 1] batches instead of [3, 3, 1]. MONAI clears the unfinished first-epoch gradients, but TraceML carries their measurements into the next update. Please discard that unfinished capture at the epoch boundary and add this regression test. This concerns measurement attribution; training itself is unaffected.

2. Correct the CI comment. It says test-latest-deps provides MONAI compatibility coverage, but that job does not install or test MONAI. Correcting the comment is enough here; expanded scheduled coverage can follow separately.

The planned phase timings, documentation, support matrix, and examples can remain in their follow-up PRs. I do not see a need for a broader redesign before accepting Part 1 once these points are addressed.

abhinavsriva (changes requested):
Minor change is requested

Pendu:
Both points addressed.

1. Fixed. The rule you traced is right.
MONAI zeroes an unstepped window's gradients (`should_zero_grad`) once it learns the epoch length.
The old code never read that signal, so it kept merging the leftover iteration into whatever group stepped next.
`_zeroes_gradients` now mirrors `should_zero_grad`, the same way `_ends_update_group` mirrors `should_step`.
It runs on `GET_BATCH_STARTED`, the earliest hook for a new iteration, so the dropped window's own fetch measurement isn't lost along with it.
Groups are `[3, 3, 1]` now, with a warning fired once per run.
Added `test_an_epoch_boundary_mid_accumulation_drops_the_unstepped_group`, plus a mutation proving the guard binds.

2. Fixed. The comment now says `test-latest-deps` doesn't install MONAI, rather than implying it does.

A second review pass, with real repo access this time, found two more gaps in that fix and one in the existing handler.
The discard ran outside the file's own best-effort boundary, so an exception in it (an empty loader with `epoch_length == 0`, for one) would have reached training.
Reproduced the crash on the unwrapped code and confirmed the fix survives it, both under a real `traceml run`.
The accumulation tests asserted group counts and sizes but never which iterations MONAI actually stepped on, so a handler whose rule quietly drifted from MONAI's could still pass.
Every row now asserts the exact iteration list.
A `zero_grad` spy proves MONAI really does wipe the window at iteration 5 without stepping on it, which is the one premise the whole fix rests on.
Also added the missing negative control: a loader whose windows always step cleanly, proving the discard never fires when there is nothing stale.

While tracing the fix I found the mutation runner had two anchors that collided with new code.
Two module-level lines now appear in two functions instead of one.
Fixed those too so the mutation pass still discriminates correctly.

Also picked up CodeRabbit's import-guard suggestion: `except Exception` instead of `except ImportError`, so a broken MONAI install still sets `IS_MONAI_AVAILABLE = False` instead of crashing the import.

Re-verified: 47 passed in `test_monai.py`, 268 passed / 8 skipped in `tests/integrations`, 1958 passed / 8 skipped full suite.
27 of 27 mutations killed, control green.

abhinavsriva (approved):
Re-reviewed Part 1. Both requested changes are addressed, including regression coverage for unfinished accumulation. I found no further blocking issue within this PR’s scope. The changelog conflict is resolved locally, and 179 targeted tests pass. Once the resolution is pushed and CI is green, this looks ready to merge.

PR #473, opened 2026-09-21
MONAI integration, part 2 of 4: forward, backward and optimizer time

Pendu:
Part 2 of 4 for MONAI support (#465), planned in #468. Stacked on #472, so review this against that branch rather than against the release branch it lands on.

Part 1 published the step envelope, Input Wait, step memory and H2D. This adds the three compute phases.

| Phase | Seam |
| --- | --- |
| Forward | the `engine.inferer(...)` call MONAI makes, so `network.train()` and `zero_grad()` stay outside it |
| Backward | `IterationEvents.LOSS_COMPLETED` to `BACKWARD_COMPLETED` |
| Optimizer | step hooks on the trainer's own optimizer, registered for the run and removed at its exit |

The loss is still in no phase, because TraceML has no loss stream. It stays inside the step, as with the Lightning callback.

A later external review pass caught one defect here and it is fixed in this branch: the inferer proxy had no `__setattr__`, so a handler that stored state on `engine.inferer` during a run wrote it to the proxy, and the proxy is swapped out at run end. Reproduced end to end under `traceml run`, where the attribute was lost, against an untraced run where it survived. Writes now reach the real inferer, and a test asserts it.

TraceML's global optimizer hooks are not used.
They would time every optimizer in the process for the rest of its life, accumulate events outside any traced run, and make `traceml.wrap_optimizer()` refuse.
Per-optimizer hooks are public torch API, are scoped to this trainer, and fire inside `scaler.step()` under AMP.
An optimizer already wrapped by `traceml.wrap_optimizer()` is left alone, and an optimizer without the hook API loses only its own phase.

## Evidence

GATE: ran=true.
`pytest tests/integrations/test_monai.py` is 57 passed on MONAI 1.6.0 with Ignite 0.5.5.
`pytest tests/integrations` is 279 passed, 8 skipped.
The full suite is 1969 passed, 8 skipped, against 1954 passed, 8 skipped on the branch below: the 14 tests this part adds plus its conformance row.
The `test-monai` job now also runs the conformance file, at the pinned versions and again at the declared Ignite floor.

TRANSITIONS: forward open to close across the model call, including an inferer that raises.
Backward open at the loss to close after the backward pass, on both the AMP and the non-AMP branch.
Optimizer not-stepping to stepping, once per update group, and stepping to skipped on an accumulating iteration.
Installed to degraded, per phase: a phase that cannot be installed leaves the others tracing.
Inside an iteration to between iterations, where an optimizer step belongs to no step and is not timed.
Hooked to unhooked at every run exit, including a crash.
Reported to reported-again, since a transient phase failure warns once per run rather than once per process.

RANKS: not exercised and not claimed, unchanged from the branch below.
No phase has a per-rank branch.

TRIPWIRE: each new guard is fired on purpose by a test.
A duck-typed optimizer with no `register_step_pre_hook`, an inferer replaced mid-run, an optimizer step taken between iterations, a subclass whose `step()` calls `super().step()` and so fires the hook pair twice, the kill switch flipped mid-run, and a forced failure of each phase installer.
A mutation pass removed each promised behaviour in turn: 37 of 37 were caught, with the unmutated control green.
The set spans both parts: 23 of the 25 mutations written for the branch below still apply verbatim here and still fail, and the two that do not are the `prepare_batch` restore pair, superseded by the generalised per-attribute restore that carries its own mutation.

TWINS: searched the concept "a phase timed from a framework signal that may not arrive".
Forward now warns when its callable is replaced mid-run, exactly as `prepare_batch` already did; both go through the same restore path.
The optimizer's hooks are the third instance and are gated on being inside an iteration.
Input Wait's unmatched-start case, swept in the branch below, is unchanged.

The conformance gate was checked by breaking it: with the optimizer hook removed, `test_declared_step_time_streams_emit[monai]` fails; with it present, the file passes.
The harness runs with the process-wide DataLoader patch and the global optimizer hooks removed, and puts the patch back afterwards. An integration that runs earlier in the same file installs both, and either one would satisfy a row here on its own.

End to end under `traceml run`, 32 iterations with `accumulation_steps=2`: forward 66.55 ms per step across the group's two model calls, backward 1.16 ms, optimizer 0.05 ms, Input Wait 121.02 ms, for 16 published steps against the trainer's 16 optimizer updates.

## Known edges, for the guide in #466

- Backward is measured between this handler's own callbacks, so another handler registered between them is inside the window. No ordering of `train_handlers` avoids both edges.
- While a run is traced, `engine.inferer` is a timing proxy. Reads and writes both reach the real inferer, so state a handler stores on it survives, but the object is not MONAI's `Inferer` for an `isinstance` check, `repr` shows the proxy, and it does not deepcopy while the run is live. The original is restored at every run exit.
- Optimizer time here is `optimizer.step()` only, and the step ends at `MODEL_COMPLETED`, so a step-interval LR scheduler falls outside both. The Lightning callback includes it in optimizer time.
- The AMP branch is covered on CPU, but torch disables a CPU `GradScaler`, so the scaler's skip-on-inf path rests on torch's source rather than on a test here.


abhinavsriva:
@pendu The overall approach looks appropriate for Part 2. Forward wraps MONAI’s inferer call, backward uses MONAI’s events, and optimizer timing uses hooks on the trainer’s own optimizer. Keeping these hooks scoped to the run makes sense.

Please address these points:

1. Only record forward calls inside a training iteration. Currently, a custom handler calling engine.inferer between iterations adds that call’s forward time to the next training step. I reproduced forward-event counts of [1, 2, 1, 1] for four ordinary training steps. Please use the same inside-iteration check already used by optimizer timing and add one regression test.

2. Resolve the inherited unfinished-epoch capture issue in Part 1. For a length-less loader with four batches, accumulation of three, and two epochs, measurements are grouped as [3, 4, 1] instead of [3, 3, 1]. MONAI clears the leftover gradients, but TraceML keeps their measurements. Please fix this in Part 1 and carry it through the stack. Part 3’s guide should then be corrected; it currently describes this carryover as expected behavior.

3. Run CI on the stacked branches before merging. These PRs currently show only CodeRabbit because the CI branch filter excludes their target branches. A manual CI run is sufficient; no broader workflow redesign is needed.

Backward callback overhead and inferer-proxy limitations are already disclosed. GPU evidence and the real-workload case study can remain follow-ups.

abhinavsriva (changes requested):
Plz have a look at the comments.

Pendu:
All three addressed.

1. Fixed. Forward timing is now gated on `_inside_iteration()`, the same check the optimizer hooks already used.
A call to `engine.inferer(...)` between iterations no longer attributes its time to the next step.
Added `test_an_inferer_call_between_iterations_is_not_timed`, mirroring the existing optimizer test.

2. Fixed in #472 and carried through here.
The window's own fetch measurement is preserved too, not just its step-time region, by hooking the discard on `GET_BATCH_STARTED` rather than at `prepare_batch`.
Groups are `[3, 3, 1]`, not `[3, 4, 1]`.
The guide's "What one step is" section is corrected.
It now says the window is dropped at the point MONAI zeroes its gradients, not that it carries into the next epoch.

A second review pass on #472's fix, this time with real repo access, found a real crash path and a real test-rigor gap.
The discard ran outside the file's own best-effort boundary; it's wrapped now, same as every other entry point.
The accumulation tests never pinned which iterations MONAI actually stepped on, only the count and the group sizes.
Every row now asserts the exact iteration list, and a `zero_grad` spy proves the fix's premise directly rather than assuming it.
`_opt_depth` is also reset at run end now, so that invariant doesn't depend on the next run always starting cleanly.
Rebased this branch on top of both rounds and re-verified.

3. Done. `workflow_dispatch` on all three branches, green.

Re-verified after the rebase: 62 passed in `test_monai.py`, 284 passed / 8 skipped in `tests/integrations`, 1974 passed / 8 skipped full suite.
38 of 38 mutations killed on this branch.
Re-ran #472's own 27-mutation set against this tree too: 25 apply verbatim and still fail.
The other 2 are the known `prepare_batch` restore pair, already superseded by the generalised restore.

abhinavsriva (approved):

Issue #306, opened 2026-08-03
wrap_dataloader_fetch before init double-times fetches once the auto patch installs


Pendu:
### Summary

Calling `traceml.wrap_dataloader_fetch(loader)` before `traceml.init(mode="auto")` leads to every fetch being timed twice once the automatic `DataLoader` patch installs: once by the patched `__iter__` generator and once by the wrapper's `timed_region`. Input-wait timing for that loader reads roughly 2x.

### Mechanism

The duplicate-instrumentation guard runs only at wrap time: `_ensure_dataloader_wrapper_allowed` (`sdk/wrappers.py`, line 41) early-returns when `DataLoader._traceml_patched` is false (line 49), which is the state before `init`. After `init` installs the patch, `iter(loader)` returns the patch's timing generator, and the wrapper then times each `next()` on top of it.

Reproduced on #130's head (`ee74523`): wrapping a loader, then installing the patch, then consuming 4 batches produced 10 `_traceml_internal:dataloader_next` timing events (each real fetch timed twice, plus both layers timing the final empty fetch).

### Prior art in the codebase

`_WrappedH2D.to()` handles exactly this wrap-then-init ordering with an explicit deferral to the auto patch. #130 added the same deferral for the batch-size bytes (the wrapper detects the patched generator via its `__name__` and skips recording; in the reproduction above the bytes side correctly recorded 4 events for 4 fetches). The fetch timing predates #130 and still lacks the deferral.

### Possible fix

Reuse #130's detection in `_WrappedDataLoaderIterator.__next__`: when the wrapped iterator is the patch's own generator, skip the wrapper's `timed_region` as well, not just the byte recording. One condition, mirrors the `_WrappedH2D` behavior.

Pendu:
Deferred past 0.3.6 together with #130, which has been retargeted to land in the next release. Not scheduled for the current release branch.

PR #474, opened 2026-09-21
fix: enforce deterministic wrapper ownership

abhinavsriva:
## Summary

Closes #306.  
Related to #141.

Makes `traceml.init(...)` the required configuration boundary before public manual wrappers and centralizes ownership between automatic and manual instrumentation.

This prevents wrap-before-init ordering from producing duplicate DataLoader timing and makes conflicting configurations fail immediately with a clear error.

## What changed

- Added one internal policy for wrapper initialization and ownership.
- Applied it consistently to DataLoader, H2D, forward, backward, and optimizer wrappers.
- Defined wrapper behavior by mode:
  - `auto`: automatic instrumentation owns supported PyTorch phases.
  - `manual`: public wrappers are allowed.
  - `selective`: wrappers are allowed only for phases whose automatic patch is disabled.
- Kept `wrap_dataloader_fetch(...)` available in `auto` for custom non-PyTorch iterators such as Ray Data.
- Moved patched-iterator recognition into the DataLoader patch module.
- Updated public documentation, migration guidance, and regression coverage.

## Scope

This changes public manual-wrapper configuration and validation only.

Existing automatic PyTorch, Hugging Face, Lightning, Accelerate, DeepSpeed, Ray, and MONAI instrumentation paths remain unchanged. `trace_step(...)` and custom `trace_time(...)` are also unchanged.

## Out of scope

- Timing boundaries or aggregation behavior
- Telemetry schemas, storage, or frontend behavior
- Diagnosis calculations
- Framework callback or integration redesign
- General instrumentation lifecycle and completeness work still tracked by #141

## Validation

- Full test suite: 1,937 passed, 8 skipped
- Strict documentation build passed
- Ruff, Black, and `git diff --check` passed


Pendu:
Reviewed `f0f3bef0` against `version_0.4.1` `f1cd1f97` (18 files, +477/-130).
No prior review to carry over: CodeRabbit skipped this PR because auto reviews are off for non-default base branches.
The 157 tests in `tests/sdk`, `tests/test_h2d_timing.py` and the missing-signals test pass locally, and Ruff and Black are clean.
I did not run the full suite or the docs build.

The mode matrix matches the PR body.
I ran it through the real `init` on head: `auto` rejects `wrap_forward`, `manual` allows it, and `selective` with `patch_dataloader=True` rejects a torch `DataLoader` but allows a plain generator.
The strict-init requirement is documented as **Breaking** in the CHANGELOG, so the items below are what that entry does not cover.

1. **`TRACEML_DISABLED=1` no longer protects a wrapper call made without `init()`.**
With the kill switch set and no `init()`, `wrap_forward(model)` returns on base and raises `RuntimeError` on head.
The `None` check at `instrumentation_policy.py:52` runs before the disabled branch at line 31 can be reached.
Users who set the kill switch to turn TraceML off in production and leave wrapper calls in their code now crash instead.
Checking the kill switch before the init requirement would keep the old behaviour for them.

2. **Two more calls that base allowed now raise.**
Both return normally on base and raise `RuntimeError` on head:

```
init(mode="auto"); reset_optimizer_timing(); wrap_optimizer(opt)
init(mode="auto"); wrap_h2d(container_with_own_to_method)
```

The first is stated in the new docstring at `optimizer_hooks.py:110-113`.
The second is a batch object whose `.to()` does its own work, which the automatic `Tensor.to` patch cannot see, so nothing measured it on either tree.
The CHANGELOG entry covers wrap-before-init and auto-owned phases, and these two sit outside both.

3. **Question: the raise happens at wrapper creation, and that is often inside the training loop.**
`_capability.py:7` says "Warn, never raise (fail-open: instrumentation must never break training)", and `CONTRIBUTING.md:81` and `:88` say the same.
The PR chooses to raise deliberately, so this is your call rather than a defect.
Was warn-and-pass-through considered for the conflict case, keeping the raise only for the missing-init case?
A user who moves `init()` after a per-step `wrap_*` call will hit the error mid-run instead of at startup.

4. **A bad argument before `init()` now reports the wrong error.**
`wrap_dataloader_fetch(123)` and `wrap_h2d(123)` with no `init()` raise `TypeError` on base and `RuntimeError` on head.
The `api.py` docstrings promise `TypeError` for these inputs (`api.py:136`, `:163`).
Running the type check before the policy check restores that.

5. **Coverage gap, not a bug: no ownership test goes through the real `init`.**
The new tests set the config with `_set_wrapper_config` (`tests/sdk/test_init_and_wrappers.py:40`), and `test_instrumentation_policy.py` never calls `init`.
The behaviour is right, since the matrix above came from the real `init`.
It is pinned only by hand-built configs, so a change to how `init` stores its mode would leave every test green.
One real-`init` test per mode would close it.

Checked and found fine:

- The stored config always matches what is installed.
  `initial.py:619` stores the requested config only after `_apply_requested_patches` succeeds (`:602`), and both failure paths store a `disabled=True` no-op config (`:599`, `:617`), which the policy treats as "wrappers allowed".
- A second `init` with a different config raises (`initial.py:562`), so an `init(manual)` followed by `init(auto)` cannot leave both owners active.
- In-repo consumers survive.
  `examples/manual_custom_minimal.py:74` calls `init(mode="manual")` before wrapping, the Ray integration calls `traceml.init` in each worker at `integrations/ray.py:279`, and no code under `src/` outside the SDK calls the guarded wrappers.
- `instrumentation_policy` imports cleanly as the first import in a fresh process.
- A `DataLoader` subclass with its own `__iter__` is rejected in auto mode, and a non-`DataLoader` that delegates to one is allowed.
  Both behave the same on base, so neither comes from this PR.

